PR #1, opened 2018-04-25
Add code and examples

cgrushko:


cgrushko:
@mhlopko

cc_configure_make/def.bzl
```diff
@@ -0,0 +1,98 @@
...
+    out_lib = ctx.actions.declare_file("{}.a".format(ctx.attr.name))
+    outputs = [out_includes, out_lib]
+
+    cpp_fragment = ctx.fragments.cpp
+    compiler_options = [] # cpp_fragment.compiler_options(ctx.features)
+    c_options = compiler_options + cpp_fragment.c_options
+    cxx_options = compiler_options + cpp_fragment.cxx_options(ctx.features)
+
+    CFLAGS = "\"{}\"".format(' '.join(c_options))
+    CXXFLAGS = "\"{}\"".format(' '.join(cxx_options))
```
mhlopko: FYI, but there's a *lot* of env vars that we should support (https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html). We will add them as we go.

cc_configure_make/def.bzl
```diff
@@ -0,0 +1,98 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+def _cc_configure_make_impl(ctx):
+    out_includes = ctx.actions.declare_directory(ctx.attr.name + "-includes.h")
+    out_lib = ctx.actions.declare_file("{}.a".format(ctx.attr.name))
+    outputs = [out_includes, out_lib]
+
+    cpp_fragment = ctx.fragments.cpp
```
mhlopko: Skylark api on the CppConfiguration is being removed as we speak, see cl/194106058 for an example migration.

cc_configure_make/def.bzl
mhlopko: Hmm but we should specify `CC` and `CXX` at least, and once https://docs.google.com/document/u/1/d/1g91BWJITcYw_X-VxsDC0VgUn5E9g0kRBGoBSpoO41gA/edit is done (O(weeks)) we can specify different tools for those

cc_configure_make/def.bzl
```diff
@@ -0,0 +1,98 @@
...
+    compiler_options = [] # cpp_fragment.compiler_options(ctx.features)
```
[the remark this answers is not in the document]
cgrushko: This trips `c-ares`, so I want to add it only after I have an idea how to handle it.

cc_configure_make/def.bzl
cgrushko: How will the code look like here?

cc_configure_make/def.bzl
cgrushko: Is there an external version?

cc_configure_make/def.bzl
mhlopko: roughly like this: cc_common.configure_features(cc_toolchain).get_command_line(cc_common.c_compile_action_name, cc_common.get_compile_variables(....))

cc_configure_make/def.bzl
mhlopko: I'm not aware of any yet.

PR #7, opened 2018-07-30
use tools paths and flags from the toolchain

irengrig:


irengrig:
Happy to report, that I was able to use cmake_external library with the custom toolchain and with android_binary.

NB
the Bazel flag --experimental_enable_cc_skylark_api is nessesary

**Android** //framework_example/cmake_android:app: Please note that Android SDK and Android NDK should be installed, and environment variables should be passed as flags to the bazel build command: --action_env=ANDROID_HOME=... --action_env=ANDROID_NDK_HOME=...

Flags for **custom crosstool toolchain** //framework_example/cmake_cross:test_zlib:
--crosstool_top=//framework_example/cmake_cross/tools/arm_compiler:toolchain
--cpu=armeabi-v7a
--experimental_enable_cc_skylark_api

Also, for //framework_example/cmake_cross:test_zlib, I modified the wrapper scripts so that they could be called with relative paths by the Bazel rules (like cc_library), and with absolute paths with the cmake_external - see, for example, https://github.com/irengrig/rules_foreign_cc/blob/f4226edfffd5c0b5ba0ddd65ca5d70540b284d20/framework_example/cmake_cross/tools/arm_compiler/linaro_linux_gcc/arm-linux-gnueabihf-gcc

tools/build_defs/cc_linking_util.bzl
```diff
@@ -2,24 +2,56 @@
 """
```
mhlopko: I'd rename this file, it's not strictly speaking only dealing with linking, right? Wdyt about cc_toolchain_util?

framework_example/cmake_cross/compilers/linaro_linux_gcc_5.3.1.BUILD
```diff
@@ -0,0 +1,81 @@
+package(default_visibility = ['//visibility:public'])
```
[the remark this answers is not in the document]
irengrig: Thank you for the advice! This totally makes sense!

I plan to address this later, when I will write the proper tests. Currently, it is still prototype stage to let our first users give us some feedback.
(This concrete code comes from the Bazel tests.)

irengrig:
Thank you for the review!

I updated the code.
I removed the example with CROSSTOOL file, can not make it work now (which might be not related to this rule but to the toolchain itself), will return to that later.
Also, for now I do not pass the static Cxx linker flags, as now, when the correct linker (AR) is being passed, it breaks the CMake linking.

The flag for the Skylark C/C++ API has changed: 
--experimental_cc_skylark_api_enabled_packages=tools/build_defs

tools/build_defs/cc_linking_util.bzl
irengrig: Done.

dslomov:
Please update README for the repo with new examples to try and instructions.

irengrig:
For the README file, I created a separate pull request: https://github.com/bazelbuild/rules_foreign_cc/pull/9

PR #25, opened 2018-08-14
Add ptions to cmake_external to filter information from C/C++ toolchain

irengrig:
The user can pass a whitelist of CMake environment variables and CMake options to be created from the toolchain information (and pass alternative values with the common cmake_options attribute).
By default, all created variables and options are passed.

### DIFF
--- a/tools/build_defs/framework.bzl
+++ b/tools/build_defs/framework.bzl
@@ -180,6 +180,7 @@ def cc_external_rule_impl(ctx, attrs):
     ]
 
     script_text = "\n".join(script_lines)
+    print("script text: " + script_text)
 
     ctx.actions.run_shell(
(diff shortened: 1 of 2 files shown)


tools/build_defs/framework.bzl
```diff
@@ -180,6 +180,7 @@ def cc_external_rule_impl(ctx, attrs):
     ]
 
     script_text = "\n".join(script_lines)
+    print("script text: " + script_text)
```
mhlopko: Is this intentional print?

irengrig: Yep; I added this in another pull request already.

irengrig:
As we spoke, I changed this to rather adjust values of flags, taken from toolchain.
Let us have a separate pull request for that: #27 
Please review :)

Closing this one.

PR #68, opened 2018-08-29
implements pcl cmake_external build rule

bbarnes52:


irengrig:
Thank you for the patch! Could you please check that about the CLA? (See automatic comment)

irengrig:
Also, seems it uses locally installed Boost. Which would be anyway not so trivial to compile in an example :)
We should probably add a comment about that..

bbarnes52:
testing CLI check

irengrig:
It seems that you attached the testing commit which you did not intent to include in pull request :)

Also, please sign the CLA, I can not merge the pull request without that.

irengrig:
I created a pull request to your fork with building boost.

Also, seems that we are using different versions of CMake, can not think of other differences - but for me it works only if I specify EIGEN_INCLUDE_DIR and FLANN_INCLUDE_DIR, not DIRS variant.
Lets keep both variants, it does not affect anything.

I was able to build the target in 4500 seconds :)

irengrig:
Cool, we did it!

PR #112, opened 2018-09-21
External C/C++ libraries rules need to pass transitive info

irengrig:
- to have the same interface as cc_library to hide the "implementation details",
so that the information about the install directory and its contents of non-direct external dependencies of external library was available for external build
- this is particularly needed for configure-make rule, where we do not control the way needed libraries are found, so we want to provide the install directory and the -L, -I flags (include and library directories to be searched) consistently,
pointing to the directories under $EXT_BUILD_DEPS
- as a side effect, *.la files that contained links to absolute paths of dependencies (pointing somewhere under ...sandbox/bazel-out/...) now will point to directories under $EXT_BUILD_DEPS and so will be replaced correctly/will not point to paths under the previous sandbox

### DIFF
--- a/tools/build_defs/configure_script.bzl
+++ b/tools/build_defs/configure_script.bzl
@@ -40,11 +40,11 @@ def _define_deps_flags(deps, inputs):
     lib_dirs = []
 
     # Here go libraries built with Bazel
-    lib_dirs_set = {}
+    gen_dirs_set = {}
     for lib in inputs.libs:
         _dir = lib.dirname
-        if not lib_dirs_set.get(_dir):
-            lib_dirs_set[_dir] = 1
+        if not gen_dirs_set.get(_dir):
+            gen_dirs_set[_dir] = 1
             lib_dirs += ["-L$EXT_BUILD_ROOT/" + _dir]
 
     include_dirs_set = {}
@@ -64,13 +64,13 @@ def _define_deps_flags(deps, inputs):
     # -config or other mechanics.
     # Since we need the names of include and lib directories under
     # the $EXT_BUILD_DEPS/<lib_name>, we ask the provider.
-    lib_dirs_set = {}
+    gen_dirs_set = {}
     for dep in deps:
         external_deps = dep[ForeignCcDeps]
         if external_deps:
             for artifact in external_deps.artifacts:
-                if not lib_dirs_set.get(artifact.gen_dir):
-                    lib_dirs_set[artifact.gen_dir] = 1
+                if not gen_dirs_set.get(artifact.gen_dir):
+                    gen_dirs_set[artifact.gen_dir] = 1
 
                     dir_name = artifact.gen_dir.basename
                     include_dirs += ["-I$EXT_BUILD_DEPS/{}/{}".format(dir_name, artifact.include_dir_name)]
@@ -99,18 +99,18 @@ _CONFIGURE_TOOLS = {
 }
 
 def _get_configure_variables(tools, flags, user_env_vars):
-    dict = {}
+    vars = {}
 
     for flag in _CONFIGURE_FLAGS:
         flag_value = getattr(flags, _CONFIGURE_FLAGS[flag])
         if flag_value:
-            dict[flag] = flag_value
+            vars[flag] = flag_value
 
     # Merge flags lists
     for user_var in user_env_vars:
-        toolchain_val = dict.get(user_var)
+        toolchain_val = vars.get(user_var)
         if toolchain_val:
-            dict[user_var] = toolchain_val + [user_env_vars[user_var]]
+            vars[user_var] = toolchain_val + [user_env_vars[user_var]]
 
     tools_dict = {}
     for tool in _CONFIGURE_TOOLS:
@@ -124,14 +124,14 @@ def _get_configure_variables(tools, flags, user_env_vars):
         if toolchain_val:
             tools_dict[user_var] = [user_env_vars[user_var]]
 
-    dict.update(tools_dict)
+    vars.update(tools_dict)
 
     # Put all other environment variables, passed by the user
     for user_var in user_env_vars:
-        if not dict.get(user_var):
-            dict[user_var] = [user_env_vars[user_var]]
+        if not vars.get(user_var):
+            vars[user_var] = [user_env_vars[user_var]]
 
-    return dict
+    return vars
 
 def _absolutize(workspace_name, text):
     return absolutize_path_in_str(workspace_name, "$EXT_BUILD_ROOT/", text)
--- a/tools/build_defs/configure_script.bzl
+++ b/tools/build_defs/configure_script.bzl
@@ -42,10 +42,10 @@ def _define_deps_flags(deps, inputs):
     # Here go libraries built with Bazel
     gen_dirs_set = {}
     for lib in inputs.libs:
-        _dir = lib.dirname
-        if not gen_dirs_set.get(_dir):
-            gen_dirs_set[_dir] = 1
-            lib_dirs += ["-L$EXT_BUILD_ROOT/" + _dir]
+        dir_ = lib.dirname
+        if not gen_dirs_set.get(dir_):
+            gen_dirs_set[dir_] = 1
+            lib_dirs += ["-L$EXT_BUILD_ROOT/" + dir_]
 
     include_dirs_set = {}
     for dir_list in inputs.include_dirs:
--- a/tools/build_defs/configure_script.bzl
+++ b/tools/build_defs/configure_script.bzl
@@ -1,5 +1,5 @@
 load(":cc_toolchain_util.bzl", "absolutize_path_in_str")
-load(":framework.bzl", "collect_libs")
+load(":framework.bzl", "ExternallyBuiltTransitive", "collect_libs")
 
 def create_configure_script(
         workspace_name,
@@ -11,9 +11,10 @@ def create_configure_script(
         user_vars,
         is_debug,
         configure_command,
+        deps,
         inputs):
     vars = _get_configure_variables(tools, flags, user_vars)
-    deps_flags = _define_deps_flags(inputs)
+    deps_flags = _define_deps_flags(deps, inputs)
 
     vars["LDFLAGS"] = vars["LDFLAGS"] + deps_flags["LDFLAGS"]
     vars["CPPFLAGS"] = deps_flags["CPPFLAGS"]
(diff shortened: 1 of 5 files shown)


tools/build_defs/framework.bzl
```diff
@@ -404,23 +429,29 @@ def _define_inputs(attrs):
...
 
     # This framework function-built libraries: copy result directories under
     # $EXT_BUILD_DEPS/lib-name
     ext_build_dirs = []
+    ext_build_dirs_set = {}
 
     for dep in attrs.deps:
-        provider = dep[OutputGroupInfo]
-        ext_built = provider and hasattr(provider, "gen_dir")
+        provider = dep[ExternallyBuiltTransitive]
```
dkelmer: Could you give "provider" a more specific name. Both for readability and also so that you won't shadow the global built-in "provider"

tools/build_defs/framework.bzl
```diff
@@ -404,23 +429,29 @@ def _define_inputs(attrs):
...
 
         linking_infos_all += [dep[CcLinkingInfo]]
         compilation_infos_all += [dep[CcCompilationInfo]]
 
-        if ext_built:
-            ext_build_dirs += provider.gen_dir.to_list()
+        if provider:
+            for artifact in provider.artifacts:
+                if not ext_build_dirs_set.get(artifact.gen_dir):
+                    ext_build_dirs_set[artifact.gen_dir] = 1
```
dkelmer: how big do you expect this set to get? is it worth it for the faster look up vs just checking if the item is in the list?

tools/build_defs/framework.bzl
```diff
@@ -404,23 +429,29 @@ def _define_inputs(attrs):
...
-        provider = dep[OutputGroupInfo]
-        ext_built = provider and hasattr(provider, "gen_dir")
+        provider = dep[ExternallyBuiltTransitive]
 
         linking_infos_all += [dep[CcLinkingInfo]]
         compilation_infos_all += [dep[CcCompilationInfo]]
 
-        if ext_built:
-            ext_build_dirs += provider.gen_dir.to_list()
+        if provider:
```
dkelmer: Is this loop not flattening the depset for every library?

tools/build_defs/framework.bzl
```diff
@@ -404,23 +429,29 @@ def _define_inputs(attrs):
...
+            for artifact in provider.artifacts:
+                if not ext_build_dirs_set.get(artifact.gen_dir):
+                    ext_build_dirs_set[artifact.gen_dir] = 1
+                    ext_build_dirs += [artifact.gen_dir]
         else:
```
dkelmer: In what case would there not be a provider?

tools/build_defs/configure_script.bzl
```diff
@@ -34,27 +35,50 @@ def create_configure_script(
     )]
     return "\n".join(script)
 
-def _define_deps_flags(inputs):
-    libs = collect_libs(inputs.deps_linking_info)
-
+def _define_deps_flags(deps, inputs):
```
dkelmer: What is the type of inputs? I can't seem to trace it. And did it change in this change?

tools/build_defs/configure_script.bzl
```diff
@@ -34,27 +35,50 @@ def create_configure_script(
...
     # It is very important to keep the order for the linker => put them into list
     lib_dirs = []
 
-    # For filtering duplicates
+    # Here go libraries built with Bazel
     lib_dirs_set = {}
```
dkelmer: Same question here about the need for the set for counting purposes only

tools/build_defs/configure_script.bzl
```diff
@@ -34,27 +35,50 @@ def create_configure_script(
...
-    for lib in libs:
+    for lib in inputs.libs:
         dir = lib.dirname
         if not lib_dirs_set.get(dir):
             lib_dirs_set[dir] = 1
-            lib_dirs += [dir]
+            lib_dirs += ["-L$EXT_BUILD_ROOT/" + dir]
+
+    include_dirs_set = {}
+    for list in inputs.include_dirs:
```
dkelmer: Rename list to a more specific name and also one that doesn't shadow the starlark "list"

tools/build_defs/configure_script.bzl
```diff
@@ -34,27 +35,50 @@ def create_configure_script(
...
+def _define_deps_flags(deps, inputs):
     # It is very important to keep the order for the linker => put them into list
     lib_dirs = []
 
-    # For filtering duplicates
+    # Here go libraries built with Bazel
     lib_dirs_set = {}
-    for lib in libs:
+    for lib in inputs.libs:
         dir = lib.dirname
```
dkelmer: rename "dir" to "directory" or something more specific to not shadow the starlark "dir"

tools/build_defs/configure_script.bzl
```diff
@@ -34,27 +35,50 @@ def create_configure_script(
...
+    include_dirs = include_dirs_set.values()
+
+    # For the external libraries, we need to refer to the places where
+    # we copied the dependencies ($EXT_BUILD_DEPS/<lib_name>), because
+    # we also want configure to find that same files with pkg-config
+    # -config or other mechanics.
+    # Since we need the names of include and lib directories under
+    # the $EXT_BUILD_DEPS/<lib_name>, we ask the provider.
+    lib_dirs_set = {}
+    for dep in deps:
```
dkelmer: Same question here, is this not flattening the depset at every level?

dkelmer:
Also, can you double check that the skylint linter is on? It would've caught the using built-in names issue

tools/build_defs/framework.bzl
```diff
@@ -118,6 +118,22 @@ def create_attrs(attr_struct, configure_name, configure_script, **kwargs):
         dict[arg] = kwargs[arg]
     return struct(**dict)
 
+ExternallyBuiltTransitive = provider(
```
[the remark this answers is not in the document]
irengrig: The idea here is that we need the following information about each externally-built dependency: the gen/install directory, and relative paths to lib/include/bin directories under it.
We can not flatten it all into depsets of files/strings.
We need to have a depset of the providers ExternallyBuildProvider, and ExternallyBuildProvider contains the information I mentioned.
I also found that example;)
Will rename it yes

tools/build_defs/framework.bzl
```diff
@@ -384,7 +408,8 @@ _InputFiles = provider(
     doc = """Provider to keep different kinds of input files, directories,
 and C/C++ compilation and linking info from dependencies""",
     fields = dict(
-        headers = """Include directories built by Bazel.
+        headers = """Include files built by Bazel. Will be copied into $EXT_BUILD_DEPS/include.""",
+        include_dirs = """Include directories built by Bazel.
```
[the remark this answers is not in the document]
irengrig: Headers are files, include directories are directories. In configure-make I need to point to the directories to include. So I need to differentiate them.

tools/build_defs/framework.bzl
irengrig: provider == null if the library was built with Bazel
Yep by the way you are right, we should consider iterating the depset in the place where the script text is created.
Could we do it as a separate PR?

tools/build_defs/framework.bzl
```diff
@@ -231,13 +247,21 @@ def cc_external_rule_impl(ctx, attrs):
...
+        bin_dir_name = attrs.out_bin_dir,
+        lib_dir_name = attrs.out_lib_dir,
+        include_dir_name = attrs.out_include_dir,
+    )
     return [
         DefaultInfo(files = depset(direct = outputs.declared_outputs)),
         OutputGroupInfo(
-            gen_dir = depset([outputs.installdir]),
             bin_dir = depset([outputs.out_bin_dir]),
-            out_binary_files = depset(outputs.out_binary_files),
```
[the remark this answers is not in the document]
irengrig: Yep you are right, actually this one should stay

tools/build_defs/configure_script.bzl
irengrig: Yep, this comes from the commits I mentioned earlier: https://github.com/bazelbuild/rules_foreign_cc/commit/d4c55bb51e70c736658d3afc8365575f49471bfe
I will add the documentation.
Thank you for a catch!

tools/build_defs/framework.bzl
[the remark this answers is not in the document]
irengrig: Renamed to ForeignCcDeps; this way it is named the same way as the rule (in future, there can be more external rules, better indicate that this one is for CC and relates to this rule), also to shorten the name; I also renamed the -artifact provider similarly, ForeignCcArtifact

tools/build_defs/framework.bzl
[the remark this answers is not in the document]
irengrig: I also returned the gen_dir back, this also can be used for getting the results of this rule for some broader usage - say, building the tool which is not self-contained binary, but rather directory

tools/build_defs/framework.bzl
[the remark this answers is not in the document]
irengrig: And yep, they are flattened into a list for building this particular library, since we need to create a text script based on that information

tools/build_defs/configure_script.bzl
irengrig: The improvemnts for the naming and documentation go here: https://github.com/bazelbuild/rules_foreign_cc/pull/116
The pull request is based on this one.

tools/build_defs/configure_script.bzl
irengrig: Yes, this is flattening, needed for script text creation.

irengrig:
Please review the corrections; also, I have created a separate pull request for the configuration script callback: https://github.com/bazelbuild/rules_foreign_cc/pull/116

irengrig:
Thank you for your review efforts:) Please see the corrections.

dkelmer:
Hey,
Most of my comments are "resolved", but not actually addressed. I'll take another look when that's done :)

irengrig:
Which ones do you mean?

dkelmer:
I think it's impossible to link to comments, so here are screenshots: 
https://screenshot.googleplex.com/5Lxh0c7M3oQ
https://screenshot.googleplex.com/zM20BnpbWPZ
https://screenshot.googleplex.com/pECorFGJqEs

tools/build_defs/framework.bzl
irengrig: It is a bit unreadable, sorry.
I was referring to the explanation above: we are iterating the transitive depset of dependencies, which in general can contain duplicates (some libraries B and C in "upper fragment" of a dependency chain can use the same "leaf" library A).

There would not be a provider in case the dependency is not build with this foreign_cc_rules rule, and is just usual Bazel cc_library (or something compatible). In that case we should act according to know Bazel interface - use library, use include directory.

PR #114, opened 2018-09-21
Use env param to pass in environment, and add PATH and CC/CXX.

mbabinski-at-google:


irengrig:
I am in the process of the implementing it in similar fashion as for CMake, PR is in progress: https://github.com/bazelbuild/rules_foreign_cc/pull/112. Please have a look.

This code in def.bzl was used on the early stage of experiments.

NB. I will be busy this week and might answer irregularly.

irengrig:
This part of the repository is not developed any more (since we moved to have the common framework file for both cmake and configure-make).

Please see it [there](https://github.com/bazelbuild/rules_foreign_cc/blob/master/tools/build_defs/configure.bzl), I think most of this request is already in that implementation.

irengrig:
Closing this one, as already implemented in the configure_make rule.

PR #124, opened 2018-10-05
Add $EXT_BUILD_ROOT to the path so CROSSTOOL files work

pweaver:
When using the build in compilers or ones from external rules this works fine because _prefix prepends $EXT_BUILD_ROOT. But when using a CROSSTOOL in the existing workspace the toolchains are given relative paths.

When using a CROSS tool in the WORKSPACE the following error occurs:

```
The CMAKE_CXX_COMPILER:

    tools/compilers/arm/gcc/arm-buildroot-linux-musleabihf-gcc

  is not a full path and was not found in the PATH.

  Tell CMake where to find the compiler by setting either the environment
  variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.
```

Adding $EXT_BUILD_ROOT to the path fixes this issue.

### DIFF
--- a/tools/build_defs/framework.bzl
+++ b/tools/build_defs/framework.bzl
@@ -187,6 +187,7 @@ def cc_external_rule_impl(ctx, attrs):
         set_envs,
         "set_platform_env_vars",
         "export EXT_BUILD_ROOT=$BUILD_PWD",
+        "path $EXT_BUILD_ROOT",
         "export BUILD_TMPDIR=$(mktemp -d)",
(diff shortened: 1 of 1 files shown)


irengrig:
Sorry it took me long, I needed to figure the things for myself.

Your correction is correct, but the codebase has changed since then,

I also had to apply some additional fixes (to make my example work) - #174.

Closing this.

PR #127, opened 2018-10-09
removes directories from output declarations to enable remote execution

bbarnes52:


bbarnes52:
@irengrig would love your feedback if you have time to review!

#131 for example would be overridden by this change (we can't declare out_bin_dir and files that belong in that directory, this violates one of the constraints of remote builds)

PR #133, opened 2018-10-16
removes directories from output declarations to enable remote execution

bbarnes52:


tools/build_defs/framework.bzl
```diff
@@ -272,12 +273,11 @@ def cc_external_rule_impl(ctx, attrs):
         bin_dir_name = attrs.out_bin_dir,
         lib_dir_name = attrs.out_lib_dir,
         include_dir_name = attrs.out_include_dir,
+        declared_outputs = outputs.declared_outputs,
     )
     return [
         DefaultInfo(files = depset(direct = outputs.declared_outputs)),
         OutputGroupInfo(
-            gen_dir = depset([outputs.installdir]),
```
irengrig: Let's keep gen_dir (further we make it refer to the copy of the directory), and out_binary_files (later each binary will be in its own group)
* Please read the comment to the PR first, with more detailed explanation

tools/build_defs/framework.bzl
```diff
@@ -401,35 +399,33 @@ def _define_outputs(ctx, attrs, lib_name):
 
     _check_file_name(lib_name, "Library name")
 
-    installdir = ctx.actions.declare_directory(lib_name)
```
irengrig: Let's keep installdir as is (in my PR I am making a copy)
* Please read the comment to the PR first, with more detailed explanation

tools/build_defs/framework.bzl
```diff
@@ -401,35 +399,33 @@ def _define_outputs(ctx, attrs, lib_name):
...
-    out_include_dir = ctx.actions.declare_directory(lib_name + "/" + attrs.out_include_dir)
-    out_bin_dir = ctx.actions.declare_directory(lib_name + "/" + attrs.out_bin_dir)
-    out_lib_dir = ctx.actions.declare_directory(lib_name + "/" + attrs.out_lib_dir)
+    out_binary_files = []
```
irengrig: We can have it a one-line call, similar to libraries

tools/build_defs/framework.bzl
```diff
@@ -401,35 +399,33 @@ def _define_outputs(ctx, attrs, lib_name):
...
     libraries = LibrariesToLinkInfo(
-        static_libraries = _declare_out(ctx, lib_name, out_lib_dir, static_libraries),
-        shared_libraries = _declare_out(ctx, lib_name, out_lib_dir, attrs.shared_libraries),
-        interface_libraries = _declare_out(ctx, lib_name, out_lib_dir, attrs.interface_libraries),
+        static_libraries = _declare_out(ctx, lib_name, attrs.out_lib_dir, static_libraries),
+        shared_libraries = _declare_out(ctx, lib_name, attrs.out_lib_dir, attrs.shared_libraries),
+        interface_libraries = _declare_out(ctx, lib_name, attrs.out_lib_dir, attrs.interface_libraries),
     )
-    declared_outputs = [installdir, out_include_dir, out_bin_dir, out_lib_dir] + out_binary_files
+    declared_outputs = [out_include_dir] + out_binary_files
```
irengrig: Since we are keeping includedir as it was, we should include it back into the list

tools/build_defs/framework.bzl
```diff
@@ -136,6 +136,7 @@ Instances of ForeignCcArtifact are incapsulated in a depset ForeignCcDeps#artifa
         "bin_dir_name": "Bin directory, relative to install directory",
         "lib_dir_name": "Lib directory, relative to install directory",
         "include_dir_name": "Include directory, relative to install directory",
+        "declared_outputs": "List of files that must be generated",
```
irengrig: Why do we need it here?

ext_buld_dirs are gathered to copy directories, built with this rule, under the $EXT_BULD_DEPS
It is enough then; all the contents with all libraries and include directories etc is already there; why do we need to have files as well?

bbarnes52:
Thank you very much for taking the time to help me understand the details!

PR #136, opened 2018-10-16
fixes boost

bbarnes52:


irengrig:
Thank you for the patch!!! And sorry for breaking it...

PR #144, opened 2018-10-24
minor code cleanup

bbarnes52:
this way is a tad more readable imho, but feel free to reject if you disagree!


irengrig:
Thank you, this looks much better indeed!

PR #159, opened 2018-11-28
implements exports - not meant for submission

bbarnes52:
hi @irengrig!

I was hoping if you had a moment to lend your expertise, I cannot for the life of me figure out why this does not work. I'm trying to add functionality that allows target A to declare target B such that any target C that depends on A will also depend on B.

AFAICT, I am treating exports of a dep exactly the same as a dep, e.g. adding B as an export of A and then having C depend on A should yield exactly the same results as just having C depend on A and B directly. However, I have tested it and this is not the case. Do you happen to see where I'm going wrong?

Thank so much for your help

Edit: relevant changes are in framework.bzl

### DIFF
--- a/for_workspace/utils_unix.sh
+++ b/for_workspace/utils_unix.sh
@@ -17,7 +17,7 @@ function path() {
 function replace_in_files() {
   if [ -d "$1" ]; then
     find -L $1 -print -type f \
-     \( -name "*.pc" -or -name "*.la" -or -name "*-config" \) \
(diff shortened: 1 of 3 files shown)


irengrig:
Wow, that's much work.
Will have a look tomorrow!

irengrig:
Hello @bbarnes52,

actually, the current implementation already passes all include (headers and include directories) and linking information (libraries and flags) transitively to the dependants.
Please have a look at the example: [cmake_synthetic](https://github.com/bazelbuild/rules_foreign_cc/tree/master/examples/cmake_synthetic).

Is the code in example similar to what you would lie to archieve?

Issue #158, opened 2018-11-28
_merge function not WAI


bbarnes52:
https://github.com/bazelbuild/rules_foreign_cc/blob/master/tools/build_defs/cmake_script.bzl#L178

If I were to add a cache_entry such as CMAKE_CXX_FLAGS, my builds will fail because the command passed to the compiler is invalid syntax. this is because the strip commands remove trailing quotations that are actually necessary.

for example, if I have CXXFLAGS="foo="bar"", and I add a cache_entry {CMAKE_CXX_FLAGS: "-Fbat}, this will yield CXXFLAGS="foo="bar -Fbat", which results in an unterminated quotation error when cmake is eventually called.

Just removing the "strip" calls worked for me, but I wanted to make sure there is not some other corner case where this would be problematic. If this is a reasonable fix, happy to submit a PR!

irengrig:
Thank you for reporting that! 
Would be happy to see the fix! Feel free to file a PR.

Yep, these strings manipulations can be tricky, I will need to run some checks that other cases are fine.

irengrig:
(Fixed)

PR #163, opened 2018-12-01
removes strip calls from string concatenation to fix bug where compil…

bbarnes52:
…er invocation failed due to invalid syntax

@irengrig can't add you as a reviewer so tagging here as FYI


irengrig:
Sorry for delay! Pushed.
I added a test for this use case: #167.

note: this closes #158

PR #181, opened 2018-12-19
adds bootstrap_options attribute for passing command line args to bootstrap.sh

### LINKED ISSUE #165: unable to find shared object files
disclaimer: not a bug or FR, hoping for assistance troubleshooting, let me know if there is a better forum

I am trying to run a simple hello_world `cc_binary` target that depends on some `cmake_external` targets. the build completes successfully, but I get the following linking error when the target is `blaze run`:

```
/blaze/a08c2e4811c846650b733c6fc815a920/execroot/__main__/bazel-out/k8-fastbuild/bin/src/examples/hello_world/hello_world: error while loading shared libraries: libconsole_bridge.so.0.4: cannot open shared object file: No such file or directory
```

libconsole_bridge.so.0.4 is created by the build, and is located at $EXT_BUILD_ROOT/bazel-out/k8-fastbuild/genfiles/third_party/ros/console_bridge/copy_console_bridge/console_bridge/ in my filesystem, as expected.

would love to get your expertise on how to proceed @irengrig . in general, how can I ensure the linker can find .so files generated by `cmake external` targets? simply adding -L<path_to_dynamic_lib> to `linkopts` does not resolve the issue.

bbarnes52:
@irengrig 

This is needed to fix #165 , whose root cause was that different compilers were being used to compile boost and cmake_external targets, resulting in unresolved symbol errors.

IMHO also a useful attribute to have in general.


irengrig:
Thank you for the pull request! Sorry that it took me so long:(

PR #184, opened 2018-12-26
Fix check for interface library artifacts.

bigdaddy:


irengrig:
Thank you so much for the pull request!
Since there is no CLA signed, and actually some corrections required, I just did the similar changes in #226.
I hope it's ok with you.

PR #188, opened 2019-01-02
Reorganize examples to be in a separate workspace

irengrig:
This will help to define native tools toolchains for examples


irengrig:
Will be possible only after https://github.com/bazelbuild/continuous-integration/pull/435 is there.

PR #192, opened 2019-01-07
Extract cmake and ninja into a toolchain; provide defaults.

irengrig:
In the workspace-level rule registration function, default toolchain
implementations are registered for all platforms, which are calling
the cmake and ninja just by name, assuming they are preinstalled.

Also, the user can pass the custom toolchains in the initialization
function, they will be registered before the defaults.

Include relevant updated into README.

### DIFF
--- a/tools/build_defs/native_tools/native_tools_toolchain.bzl
+++ b/tools/build_defs/native_tools/native_tools_toolchain.bzl
@@ -1,0 +1,37 @@
+ToolInfo = provider(
+    doc = "Information about the native tool",
+    fields = {
+        "path": """Absolute path to the tool in case the tool is preinstalled on the machine.
+Relative path to the tool in case the tool is built as part of a build; the path should be relative
+to the bazel-genfiles, i.e. it should start with the name of the top directory of the built tree
+artifact. (Please see the example "//examples:built_cmake_toolchain")""",
+        "label": """If the tool is preinstalled, must be None.
+If the tool is built as part of the build, the corresponding build target, which should produce
+the tree artifact with the binary to call.""",
+    },
+)
+
+def _native_tool_toolchain(ctx):
+    if not ctx.attr.path and not ctx.attr.label:
+        fail("Either path or label (and path) should be defined for the tool.")
+    return platform_common.ToolchainInfo(data = ToolInfo(
+        path = ctx.attr.path,
+        label = ctx.attr.label,
+    ))
+
+native_tool_toolchain = rule(
+    implementation = _native_tool_toolchain,
+    attrs = {
+        "path": attr.string(mandatory = False),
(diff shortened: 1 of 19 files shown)


irengrig:
The successful CI build (from a branch in original repo): https://buildkite.com/bazel/rules-foreign-cc/builds/166

tools/build_defs/native_tools/native_tools_toolchain.bzl
```diff
@@ -0,0 +1,37 @@
+ToolInfo = provider(
+    doc = "Information about the native tool",
+    fields = {
+        "path": """Absolute path to the tool in case the tool is preinstalled on the machine.
+Relative path to the tool in case the tool is built as part of a build; the path should be relative
+to the bazel-genfiles, i.e. it should start with the name of the top directory of the built tree
+artifact. (Please see the example "//examples:built_cmake_toolchain")""",
+        "label": """If the tool is preinstalled, must be None.
```
dslomov: 1. The value of this field is a Target - let's call it "tool" or "target" instead of "label". I was confused - I thought this filed contains an actual Label object which would be incorrect.
2. Let's say I have a tool binary checked in - do we support that? (I guess that would be label + empty path)

tools/build_defs/native_tools/native_tools_toolchain.bzl
```diff
@@ -0,0 +1,37 @@
...
+    if not ctx.attr.path and not ctx.attr.label:
+        fail("Either path or label (and path) should be defined for the tool.")
+    return platform_common.ToolchainInfo(data = ToolInfo(
+        path = ctx.attr.path,
+        label = ctx.attr.label,
+    ))
+
+native_tool_toolchain = rule(
+    implementation = _native_tool_toolchain,
+    attrs = {
```
dslomov: document these attributes

tools/build_defs/native_tools/native_tools_toolchain.bzl
irengrig: 1. Renamed label to target
2. I added an example in examples/with_prebuilt_ninja_artefact, and mentioned it in README.

tools/build_defs/native_tools/native_tools_toolchain.bzl
irengrig: Done

irengrig:
Thank you for the review! Corrected everything, please have a look.

tools/build_defs/native_tools/native_tools_toolchain.bzl
dslomov: The right way to document attributes is to add a `doc = "..."` parameter to `attr.label` and `attr.string` - sorry I didn't mention that earlier

PR #193, opened 2019-01-08
Add attribute to allow additional non-library outputs

### LINKED ISSUE #126: Support for non-cc_library artifacts in configure_make workflow.
It would be nice if configure+make projects that produced binaries and other non-library artifacts were usable in this framework. This would, for example, allow adding lex/yacc or other tools into the build process hermetically.

schoppmp:
This allows users to use additional generated files in dependent targets. Fixes #126.

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,6 +286,8 @@ attrs: {
     "interface_libraries": attr.string_list(mandatory = False),
     # Optional names of the resulting binaries.
     "binaries": attr.string_list(mandatory = False),
+    # Optional names of any additional output files.
+    "additional_outputs": attr.string_list(mandatory = False),
     # Flag variable to indicate that the library produces only headers
     "headers_only": attr.bool(mandatory = False, default = False),
   }
--- a/tools/build_defs/new_11_2018/framework.bzl
+++ b/tools/build_defs/new_11_2018/framework.bzl
@@ -80,6 +80,8 @@ CC_EXTERNAL_RULE_ATTRIBUTES = {
     "interface_libraries": attr.string_list(mandatory = False),
     # Optional names of the resulting binaries.
     "binaries": attr.string_list(mandatory = False),
+    # Optional names of any additional output files.
+    "additional_outputs": attr.string_list(mandatory = False),
     # Flag variable to indicate that the library produces only headers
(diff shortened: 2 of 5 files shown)


irengrig:
Hi @schoppmp, thank you for the pull request!
Everything is fine with the code, but I wanted to ask about the usage example of the additional outputs.

The thing is, the whole directory with the built contents is available, and there is additional constraint that one can not declare outputs in parent-child relationship, i.e. nothing directly under the resulting directory can not be declared, one will need to copy anything from the resulting directory aside.
For instance, here it is in the code: https://github.com/bazelbuild/rules_foreign_cc/blob/dfd1d815f51e3dbd36b6ed19ad1f75051a4a5862/tools/build_defs/new_11_2018/framework.bzl#L227

schoppmp:
Hi @irengrig, sorry for waiting so long with an answer.
My use case is a compiler ([Obliv-C](https://github.com/samee/obliv-c)) for which I am writing [Bazel rules](https://github.com/schoppmp/rules_oblivc). Currently, I'm using a pretty ugly [genrule](https://github.com/schoppmp/rules_oblivc/blob/master/oblivc.BUILD#L106) for compiling Obliv-C, and I would like to switch to `rules_foreign_cc`.

In any case, I need some non-library files accessible after compilation as part of the runfiles, which I'm currently doing by adding the outputs of the genrule above to the [tools](https://github.com/schoppmp/rules_oblivc/blob/master/oblivc/oblivc.bzl#L85) of the compilation action for everything that gets compiled using Obliv-C. If I switch to `rules_foreign_cc`, will all generated files be available to dependent actions if passed via `tools`, just as if they were declared as outputs of a genrule? In that case you're right and this can be closed.

irengrig:
>  If I switch to rules_foreign_cc, will all generated files be available to dependent actions if passed via tools, just as if they were declared as outputs of a genrule? 

Yes, it is possible. Here is the example how to do it: https://github.com/bazelbuild/rules_foreign_cc/blob/59a618db5f5408481957c52d345fa1368a244194/examples/configure_gnuplot/BUILD#L69

schoppmp:
Okay, thanks! I'll try to make it work in that way then. Close.

PR #194, opened 2019-01-08
Fixes for Bazel --all_incompatible_changes

laurentlb:
Tested:
  `bazel test //... --nobuild --all_incompatible_changes`

### DIFF
--- a/for_workspace/repositories.bzl
+++ b/for_workspace/repositories.bzl
@@ -8,11 +8,11 @@ def repositories():
     http_archive(
         name = "bazel_skylib",
         build_file_content = _all_content,
-        sha256 = "b5f6abe419da897b7901f90cbab08af958b97a8f3575b0d3dd062ac7ce78541f",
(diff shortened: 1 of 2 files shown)


laurentlb:
That was not enough, I should have tried:
  `bazel test @rules_foreign_cc_tests//:tests --all_incompatible_changes`

Can you look at the remaining issues?

irengrig:
Sure

irengrig:
I can not do it with flag --all_incompatible_changes, since local_cc crosstool is failing, but the related flags are not gonna be flipped in the next release, so even that is not relevant.
What I did instead, I checked everything against the flags that Bazel CI found to flip for the next release, and found/fixed one problem.
Unfortunately, night build with incompatible flags failed, will keep an eye on that.

irengrig:
Well, as I said, running with the flag does not work because of parts of Bazel itself are failing.
(Let's discuss in person)

PR #204, opened 2019-01-22
Hide build output from console

Dig-Doug:
Hi,

I'm currently using these rules to build boost in my project. The build works great, however, it currently prints a ton of messages to the console, e.g. printing all of the output files. When I build my project using my CI setup, the extra messages cause me to hit the max log size very quickly. To fix this issue, I would like to be able to hide the build output from the console.

I took a stab at fixing this myself by:
1. Writing the `script_text` in `tools/build_defs/old_12_2018/framework.bzl` to a file
1. Executing the script from # 1 and redirecting the output to `/dev/null`

This seems to work fine, but I would like to make sure this is the correct approach before working on it more. If everything looks good, I will clean this PR up so it can be merged. :)

A few other TODOs / questions I have:

1. Printing the build output is useful for debugging, so I'm guessing I should make the output redirect configurable. Is adding an attribute, e.g. `verbose = True`, to the build rules the best way to expose the option?
1. Follow up to # 1, I will print `script_text` only when `verbose = True`
1. Should I also update the other `framework.bzl` files?

If you have any other suggestions, please let me know.

Thanks!


Dig-Doug:
I was thinking about this a bit more: one other alternate approach could be that output is always written to a log file. If the build fails, the location/contents of the log file could be printed out. This would eliminate the need for a `verbose` attribute.

irengrig:
Thank you for the pull request, I was also thinking how to approach this problem.
I think the idea with a separate file and only outputting details in case of failure is the best, taking another couple hours to think about it...

Dig-Doug:
@irengrig - I'm planning to take another look at this this weekend. Let me know if you've had any other ideas on how to approach this.

irengrig:
Go ahead:) The only thing I might still need to pass a flag somehow for the case when the tests are running on CI. But I think I will figure that out later.
I like the idea with writing everything (except error stream?) into a log and outputting it only in case of error.

Dig-Doug:
Hi @irengrig, I've updated my implementation to redirect to the log file. I've also added a `.gitignore`.

Currently, my implementation redirects everything to the log file. I chose to redirect stderr because when I was trying out the changes on my project, the g3log library printed a bunch of output, mostly compiler warnings. Since that code is more or less out of my control, I don't care about those warnings and would rather hide them.

Thanks!

tools/build_defs/old_12_2018/framework.bzl
```diff
@@ -266,21 +266,46 @@ def cc_external_rule_impl(ctx, attrs):
...
+    ctx.actions.write(
+        output = build_script_file,
+        content = script_text,
+        is_executable = True,
+    )
+    build_log_file = ctx.actions.declare_file("{}_Cc{}MakeRuleLog.txt".format(lib_name, attrs.configure_name.capitalize()))
+    build_command_lines = [
+        "export BUILD_SCRIPT=\"{}\"".format(build_script_file.path),
+        "export BUILD_LOG=\"{}\"".format(build_log_file.path),
+        "$$BUILD_SCRIPT$$ &> $$BUILD_LOG$$",
```
irengrig: I like the approach, but unfortunately there are couple of problems:

- this file is only for Bazel 0.21, for Bazel 0.22 and newer /new_11_2018/framework.bzl is used
- the current script contains way too much of unix scripting. I would improve the cleanup() function in the shell toolchain to be more abstract so we can use it here, and define a new function for output redirection.

I can do both things myself later, right after applying your PR, so that you do not have to go into so many details.
I am already very grateful for the initiative:)
WDYT?

.gitignore
```diff
@@ -0,0 +1,77 @@
+#####
+# Jetbrains gitignore: https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore
+#####
+# Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio and WebStorm
+# Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
+
+# User-specific stuff
```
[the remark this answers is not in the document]
Dig-Doug: Done.

tools/build_defs/old_12_2018/framework.bzl
Dig-Doug: That sounds good to me. Thanks!

bedbad:
Hi how can see the debug output running in the command line as it was?
Cmake build takes 15 min and I could catch the wrong flags in the first seconds not waiting till they failed

PR #245, opened 2019-04-08
Boost bootstrap_options not working

Dig-Doug:
Fixes a bug where bootstrap options weren't being passed.


irengrig:
Thank you for the pull request, and sorry for the huge delay:(

PR #246, opened 2019-04-26
Windows, tests: fix for native test wrapper

laszlocsomor:
In this PR:

- Update `@bazel_skylib` to 0.8.0, which has a
  Windows-compatible unittest.bzl

- Fix unittest-using tests. Correct usage of
  `unittest.end(env)` is to return its value.

Result: //test:all now passes with Bazel 0.25.0rc3
and --incompatible_windows_native_test_wrapper

### DIFF
--- a/workspace_definitions.bzl
+++ b/workspace_definitions.bzl
@@ -103,8 +103,3 @@ def rules_foreign_cc_dependencies(
             "@rules_foreign_cc//tools/build_defs:preinstalled_cmake_toolchain",
             "@rules_foreign_cc//tools/build_defs:preinstalled_ninja_toolchain",
         )
-
-    native.register_toolchains(
-        "@bazel_skylib//toolchains/unittest:cmd_toolchain",
-        "@bazel_skylib//toolchains/unittest:bash_toolchain",
-    )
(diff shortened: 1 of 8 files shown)


irengrig:
Some tests fail with "no matching toolchains found for types @bazel_skylib//toolchains/unittest:toolchain_type".

laszlocsomor:
Thanks! I'll investigate.

workspace_definitions.bzl
```diff
@@ -102,3 +103,8 @@ def rules_foreign_cc_dependencies(
             "@rules_foreign_cc//tools/build_defs:preinstalled_cmake_toolchain",
             "@rules_foreign_cc//tools/build_defs:preinstalled_ninja_toolchain",
         )
+
+    native.register_toolchains(
```
irengrig: Please move it into /examples/WORKSPACE, then it will work fine with tests.
We should not have test toolchains in rule initialization code.

laszlocsomor: Done.

laszlocsomor:
PTAL.

laszlocsomor:
ping @irengrig

PR #249, opened 2019-05-07
Fix using Bazel-built libraries by cmake_external and configure_make

irengrig:
Please see the description of changes in commit messages.


tools/build_defs/cmake.bzl
```diff
@@ -77,6 +77,7 @@ def _create_configure_script(configureParameters):
         dict(ctx.attr.cache_entries),
         dict(ctx.attr.env_vars),
         ctx.attr.cmake_options,
+        inputs.include_dirs,
```
aehlig: Don't pass keyword arguments positionally (this is also wrong for the next argument, but at least do not make it worse).

Independently of this change, the limit for the number of positional arguments that is still managable seems to be exceeded for this function.

tools/build_defs/shell_toolchain/toolchains/impl/default_commands.bzl
```diff
@@ -73,10 +73,12 @@ if [[ -f $1 ]]; then
   return 0
 fi
 
-local children=$(find $1 -maxdepth 1 -mindepth 1)
-for child in $children; do
-  ##symlink_to_dir## $child $target
-done
+if [[ -d $1 || -L $1 ]]; then
```
aehlig: `[[` is a `bash` extension; as there are already `bash`-only constructs in the `text`, I assume it is OK, but I couldn't find the documentation about whether this is called by a generic shell (like `sh`) or by `bash` specifically (the directory name `shell_toolchain` seems to hint at the former).

irengrig: Thank you for the remark; bash is assumed, will clarify that.

irengrig:
Thank you for the review!

Issue #252, opened 2019-05-08
ranlib issues with custom toolchains


keith:
Currently rules_foreign_cc pulls the linker [from cc_common](https://github.com/bazelbuild/rules_foreign_cc/blob/4c59d4d0c5a9b0f465f57d0193a138ebd692a1b4/tools/build_defs/cc_toolchain_util.bzl#L252-L259). In some cmake builds, such as [c-ares](https://github.com/c-ares/c-ares) and [libevent](https://github.com/libevent/libevent), their cmake build systems end up calling ranlib directly, which [isn't propagated](https://github.com/bazelbuild/rules_foreign_cc/blob/4c59d4d0c5a9b0f465f57d0193a138ebd692a1b4/tools/build_defs/cmake_script.bzl#L221-L222) to cmake, and therefore ends up pulling the system one, which results in issues when you're cross compiling and the system's ranlib isn't compatible with the library it's being run on.

As far as I can tell there isn't a [action name](https://github.com/bazelbuild/bazel/blob/master/tools/build_defs/cc/action_names.bzl) that translates to ranlib. In the 2 examples above, disabling ranlib all together by setting `CMAKE_RANLIB: ""` in `cache_entries` works, but is probably pretty risky.

Does anyone have any idea how we can correctly pass ranlib through to cmake when it needs it, to not fall back on the system one? 

This is kinda related to https://github.com/bazelbuild/rules_foreign_cc/issues/185 I think

irengrig:
Hi @keith, could you have a look at https://github.com/bazelbuild/rules_foreign_cc/pull/255?
Do you think it is the right thing to do?

keith:
I've submitted a PR to envoy to disable ranlib across the board to avoid this for now https://github.com/envoyproxy/envoy/pull/6991

irengrig:
Closing this issue as you confirmed the fix #293 solves the RANLIB problem.

benjamin-branchware:
FYI for anyone ending up here where CMake is being used with the Ninja generator, and you are seeing `"" <somelibrary.a>` in your error logs, that empty string is `CMAKE_RANLIB`.

It turns out an empty string probably isn't the way to go here, and a much safer bet is to set `CMAKE_RANLIB=":"`, as a colon will still work as a no-op executable command and swallow any arguments which come after it.

PR #253, opened 2019-05-08
Allow ranlib to be overridden

keith:
This allows users to set this to an empty string if they want to disable
it. Related https://github.com/bazelbuild/rules_foreign_cc/issues/252


keith:
Thanks!

PR #255, opened 2019-05-09
Fix #252 "ranlib issues with custom toolchains" for CMake

### LINKED ISSUE #252: ranlib issues with custom toolchains

irengrig:
if the static libraries archiver in the Bazel toolchain is called "ranlib", pass it in CMAKE_RANLIB
 variable,
if the static libraries archiver in the Bazel toolchain is called "ar", pass it in CMAKE_AR variable,
otherwise, do nothing.

Not sure what should be done for configure-make case. (variables like $RANLIB, $AR etc.?)

### DIFF
--- a/tools/build_defs/cmake_script.bzl
+++ b/tools/build_defs/cmake_script.bzl
@@ -217,12 +217,15 @@ def _fill_crossfile_from_toolchain(workspace_name, target_os, tools, flags):
 
     # Force convert tools paths to absolute using $EXT_BUILD_ROOT
     if tools.cc:
-        dict["CMAKE_C_COMPILER"] = _absolutize(workspace_name, tools.cc, True)
+        dict["CMAKE_C_COMPILER"] = _absolutize(workspace_name, tools.cc, False)
     if tools.cxx:
-        dict["CMAKE_CXX_COMPILER"] = _absolutize(workspace_name, tools.cxx, True)
+        dict["CMAKE_CXX_COMPILER"] = _absolutize(workspace_name, tools.cxx, False)
 
     if tools.cxx_linker_static:
-        dict["CMAKE_AR"] = _absolutize(workspace_name, tools.cxx_linker_static, True)
+        if tools.cxx_linker_static.endswith("""/ar""") or "ar" == tools.cxx_linker_static:
+            dict["CMAKE_AR"] = _absolutize(workspace_name, tools.cxx_linker_static, False)
+        if tools.cxx_linker_static.endswith("""/ranlib""") or "ranlib" == tools.cxx_linker_static:
+            dict["CMAKE_RANLIB"] = _absolutize(workspace_name, tools.cxx_linker_static, False)
 
     if tools.cxx_linker_executable and tools.cxx_linker_executable != tools.cxx:
         normalized_path = _absolutize(workspace_name, tools.cxx_linker_executable)
--- a/test/cmake_text_tests.bzl
+++ b/test/cmake_text_tests.bzl
@@ -73,7 +73,7 @@ def _fill_crossfile_from_toolchain_test(ctx):
     tools = CxxToolsInfo(
         cc = "/some-cc-value",
         cxx = "external/cxx-value",
-        cxx_linker_static = "/cxx_linker_static",
+        cxx_linker_static = "/usr/bin/ar",
         cxx_linker_executable = "ws/cxx_linker_executable",
     )
     flags = CxxFlagsInfo(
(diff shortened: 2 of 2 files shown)


tools/build_defs/cmake_script.bzl
```diff
@@ -217,12 +217,15 @@ def _fill_crossfile_from_toolchain(workspace_name, target_os, tools, flags):
...
+        dict["CMAKE_C_COMPILER"] = _absolutize(workspace_name, tools.cc, False)
     if tools.cxx:
-        dict["CMAKE_CXX_COMPILER"] = _absolutize(workspace_name, tools.cxx, True)
+        dict["CMAKE_CXX_COMPILER"] = _absolutize(workspace_name, tools.cxx, False)
 
     if tools.cxx_linker_static:
-        dict["CMAKE_AR"] = _absolutize(workspace_name, tools.cxx_linker_static, True)
+        if tools.cxx_linker_static.endswith("""/ar""") or "ar" == tools.cxx_linker_static:
+            dict["CMAKE_AR"] = _absolutize(workspace_name, tools.cxx_linker_static, False)
+        if tools.cxx_linker_static.endswith("""/ranlib""") or "ranlib" == tools.cxx_linker_static:
```
keith: I think the problem is in bazel's case `cxx_linker_static` will always be `ar`, but yet cmake will still try and use `ranlib`, so I don't think this would fix the issue

irengrig: As far as I understand, as cxx_linker_static is coming from Bazel toolchain (cpp_link_static_library), it can be something else, not just ar.
I understood the problem you were describing in #252 as ranlib passed to Bazel as the static libraries tool.
If it is not the case, CMAKE_RANLIB variable should be used with the cmake_external rule.

We may think, however, about having "CMake toolchain" with needed CMake variables - so that all cmake_external targets refer to it, and there was no need in passing it as individual attributes.
WDYT?

irengrig:
@hlopko could you please have a look if it makes sense? Have you seen usages of ranlib/libtool instead of ar with Bazel?

tools/build_defs/cmake_script.bzl
keith: > I understood the problem you were describing in #252 as ranlib passed to Bazel as the static libraries tool.

Sorry about my wording, this isn't the case, the static library is still set to `ar`, yet the cmake attempts to use `ranlib`, and since we're not setting `CMAKE_RANLIB` it auto-detects the one on the system as `/usr/bin/ranlib` (on macOS) which breakas the build if you're cross compiling on mach-o binaries. AFAICT there isn't a place on `cc_common` where we can access the toolchains `ranlib` at this point

irengrig: Ok; seems then that there are only 2 options left:
- specify CMAKE_RANLIB in cache_entries for each cmake_external target,
- or have a toolchain with some default implementation (empty dictionary), which would contain cache_entries to apply for each cmake_external target. This sounds like a quite powerful solution, I think I like it.

keith: Either sounds good to me. I do think it would be ideal to provide a _valid_ ranlib, because at least in my case of using the ndk, we do have one, I just don't see how we can fetch it from bazel using cc_common.

hlopko:
Howdy all, nope until today I didn't know ranlib existed. Is there a way to tell cmake not to use ranlib, and to only use ar/libtool as it is passed from Bazel? Is there an advantage in using ranlib?

irengrig:
@hlopko I guess it is a variant chosen by some library; we can not avoid it, we should support his case as well. I will do the CMake-variables toolchain, this will solve the problem.

keith:
We can disable it for all cmake projects by setting it to an empty string, which is what we're currently doing to work around this problem. In theory if users were using bazel with super old versions of `ar` this might break, but that seems pretty unlikely.

irengrig:
Closing as ranlib problem was resolved in #253

PR #262, opened 2019-05-15
Migrate for --incompatible_enable_cc_toolchain_resolution

hlopko:
https://github.com/bazelbuild/bazel/issues/7260


irengrig:
Should "@bazel_tools//tools/cpp:current_cc_toolchain" references be removed then?

hlopko:
Not yet, we need them for cases when --incompatible_enable_cc_toolchain_resolution is not flipped. But in the future yes, we will be able to remove them.

PR #263, opened 2019-05-16
Windows: tests now work with native test wrapper

laszlocsomor:
See https://github.com/bazelbuild/bazel/issues/6622


laszlocsomor:
Finally, all green \o/

laszlocsomor:
ping @irengrig

laszlocsomor:
> Should we probably migrate all shell tests on $location + runfiles prefix in shell script? I will do it later then...

Yes, SGTM.
Also I'll probably add a new method to runfiles.bash, to make it work with paths returned by `$(location)`.

PR #271, opened 2019-05-28
Use forward compatible version of _configure_features

hlopko:
This PR fixes the bug introduced in https://github.com/bazelbuild/rules_foreign_cc/pull/270.

hlopko@ should not be allowed to code undercaffeinated.


irengrig:
>hlopko@ should not be allowed to code undercaffeinated.
How should I ensure that?))))

hlopko:
I'm in the process of fixing that :P

PR #278, opened 2019-06-19
Introduce clear_transitive_flags list attributes,

### LINKED ISSUE #222: Ancillary binaries are confused by outer linkeropts
I've recently been chasing down why Envoy's oss-fuzz build regressed in early Jan. There have been multiple issues, but one is related to our adoption of external_cmake, this can be seen in https://oss-fuzz-build-logs.storage.googleapis.com/log-ae2debc3-c07f-47aa-93ff-7eb4c9fe21af.txt.

The problem is that we have a fairly hokey invocation of Bazel, which you can see at https://github.com/google/oss-fuzz/blob/master/projects/envoy/build.sh. This is telling Bazel to use `-lFuzzingEnginer` everywhere. Unfortunately, `zlib`, when built under `external_cmake`, tries to do this with its test executables; that linker option is only safe to use with the target fuzzer binaries.

It might be reasonable to try and disable test binaries in `external_cmake` deps, but this doesn't work in general; e.g. zlib and c-ares don't support it by default, so we need to fix upstream, maybe a dep needs some host binary, etc.

I think one work around for us is to separate out the final link from Bazel, I'll look into this, but I'm interested the thoughts of others on the cleanest way to address this via either `external_cmake` support or changes to how we are structuring our fuzzer build.

CC @Dor1s

irengrig:
for transitive flags to be filtered out and NOT APPLIED for building this target.
Closes #222.


irengrig:
Please see the linked request #222. Also, there are other use cases for Envoy to want this functionality.

irengrig:
@hlopko Marcel, I changed the code to only filter out flags from Bazel command line, like -copts.
Btw, the flags are filtered only together with values, i.e. "--flag=value" should match to be filtered out, not only "--flag" or "--flag=otherValue".

Please review:)

irengrig:
Closing as Obsolete.

PR #286, opened 2019-07-08
Build Android example on CI

keith:
Currently there aren't any actual test targets for the NDK example, but
building it is still useful to verify.

### DIFF
--- a/.bazelci/config.yaml
+++ b/.bazelci/config.yaml
@@ -3,21 +3,29 @@ tasks:
   ubuntu1804:
     platform: ubuntu1804
     working_directory: examples
+    build_targets:
+      - "//cmake_android:app"
     test_targets:
     - "//:tests"
   ubuntu1604:
     platform: ubuntu1604
     working_directory: examples
+    build_targets:
+      - "//cmake_android:app"
     test_targets:
       - "//:tests_no_synthetic"
   macos:
     platform: macos
     working_directory: examples
+    build_targets:
+      - "//cmake_android:app"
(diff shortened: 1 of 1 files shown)


keith:
Currently for me on macOS this build actually fails because of this issue https://github.com/bazelbuild/rules_foreign_cc/issues/252

keith:
Ah the real failure here won't surface until https://github.com/bazelbuild/rules_foreign_cc/pull/287 is merged

keith:
I think we should consider disabling ranlib always since bazel doesn't support it in general. This is what we did upstream in the case of the original issue https://github.com/envoyproxy/envoy/pull/6991

keith:
Ok yep, the mac build failure here is the ranlib issue:

```
external/androidndk/ndk/toolchains/arm-linux-androideabi-4.9/prebuilt/darwin-x86_64/lib/gcc/arm-linux-androideabi/4.9.x/../../../../arm-linux-androideabi/bin/ld: error: bazel-out/android-armeabi-v7a-fastbuild/bin/cmake_android/libhello/lib/libhello.a: member at 8 is not an ELF object
```

This happens because macOS's ranlib runs on this library, and then the next command fails. The windows build looks entirely broken here so I can just remove the Android build from there

keith:
@irengrig since you fixed the ranlib issue, this is green now, is enabling this on CI reasonable?

irengrig:
Great! Last time I looked it did not compile.
Merging, thank you for contribution!

PR #287, opened 2019-07-08
Switch to rules_jvm_external

keith:
The previous gmaven_rules usage was broken, this is the new supported
way to download these dependencies.


keith:
cc @jin

PR #293, opened 2019-07-12
Fix a way of wiping CMake cache entry values defined by toolchain fro…

irengrig:
…m a command line, in case user specified empty values for these cache entries explicitly in cmake_external target

- it was a problem that we did not check that there was some value in the "toolchain" dict before we pop() - reported in comment in #292
- make code which does wiping empty values more explicit

### DIFF
--- a/test/cmake_text_tests.bzl
+++ b/test/cmake_text_tests.bzl
@@ -263,6 +263,38 @@ def _create_min_cmake_script_no_toolchain_file_test(ctx):
 
     return unittest.end(env)
 
+def _create_min_cmake_script_wipe_toolchain_test(ctx):
+    env = unittest.begin(ctx)
+
+    tools = CxxToolsInfo(
+        cc = "/usr/bin/gcc",
+        cxx = "/usr/bin/gcc",
+        cxx_linker_static = "/usr/bin/ar",
+        cxx_linker_executable = "/usr/bin/gcc",
+    )
+    flags = CxxFlagsInfo(
+        cc = ["-U_FORTIFY_SOURCE", "-fstack-protector", "-Wall"],
+        cxx = ["-U_FORTIFY_SOURCE", "-fstack-protector", "-Wall"],
+        cxx_linker_shared = ["-shared", "-fuse-ld=gold"],
+        cxx_linker_static = ["static"],
+        cxx_linker_executable = ["-fuse-ld=gold", "-Wl", "-no-as-needed"],
+        assemble = ["-U_FORTIFY_SOURCE", "-fstack-protector", "-Wall"],
+    )
+    user_env = {}
+    user_cache = {
+        "CMAKE_PREFIX_PATH": "/abc/def",
+        # These two flags/CMake cache entries must be wiped,
+        # but the second is not present in toolchain flags.
+        "CMAKE_SHARED_LINKER_FLAGS": "",
+        "WIPE_ME_IF_PRESENT": "",
+    }
+
+    script = create_cmake_script("ws", "linux", "cmake", tools, flags, "test_rule", "external/test_rule", True, user_cache, user_env, ["-GNinja"])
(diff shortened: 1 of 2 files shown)


irengrig:
I added putting empty value for CMAKE_RANLIB in this PR, because it seems that the workaround people use currently with putting empty value for CMAKE_RANLIB will not work any more as it is nothing to wipe, rather the empty value should be passed to forbid CMake to autodetect ranlib.

@keith please validate this does not break you
+ it should fix #252 and #255 is not needed then.

wrowe:
Although we are running into unrelated troubles, this appears to allow CMAKE_RANLIB to pass undisturbed.

test/cmake_text_tests.bzl
```diff
@@ -258,7 +257,39 @@ def _create_min_cmake_script_no_toolchain_file_test(ctx):
...
+    user_cache = {
+        "CMAKE_PREFIX_PATH": "/abc/def",
+        # These two flags/CMake cache entries must be wiped,
+        # but the second is not present in toolchain flags.
+        "CMAKE_SHARED_LINKER_FLAGS": "",
+        "WIPE_ME_IF_PRESENT": "",
+    }
+
+    script = create_cmake_script("ws", "linux", "cmake", tools, flags, "test_rule", "external/test_rule", True, user_cache, user_env, ["-GNinja"])
+    expected = "CC=\"/usr/bin/gcc\" CXX=\"/usr/bin/gcc\" CFLAGS=\"-U_FORTIFY_SOURCE -fstack-protector -Wall\" CXXFLAGS=\"-U_FORTIFY_SOURCE -fstack-protector -Wall\" ASMFLAGS=\"-U_FORTIFY_SOURCE -fstack-protector -Wall\" cmake -DCMAKE_AR=\"/usr/bin/ar\" -DCMAKE_EXE_LINKER_FLAGS=\"-fuse-ld=gold -Wl -no-as-needed\" -DCMAKE_PREFIX_PATH=\"$EXT_BUILD_DEPS;/abc/def\" -DCMAKE_INSTALL_PREFIX=\"test_rule\" -DCMAKE_BUILD_TYPE=\"Debug\" -DCMAKE_RANLIB=\"\" -GNinja $EXT_BUILD_ROOT/external/test_rule"
```
jin: can we break `expected` up into a string over multiple lines? a long line like this is quite difficult to read and compare with the `create_cmake_script` arguments.

tools/build_defs/cmake_script.bzl
```diff
@@ -34,6 +34,12 @@ def create_cmake_script(
 
     toolchain_dict = _fill_crossfile_from_toolchain(workspace_name, target_os, tools, flags)
     params = None
+
+    keys_with_empty_values_in_user_cache = []
```
jin: nit:

```suggestion
    keys_with_empty_values_in_user_cache = [key for key in user_cache if user_cache.get(key) == ""]
```

and move this down to just before `wipe_empty_values`

keith:
@irengrig this does fix the envoy ranlib issue without us doing this ourselves, thanks!

tools/build_defs/cmake_script.bzl
irengrig: Thank you for the suggestion, I will apply it. This can NOT be moved down, as the user_cache is changed in the following calls (intentionally).

test/cmake_text_tests.bzl
irengrig: Sorry, this does not work well: it is a single string without line breaks and when I used the symbol \ at the end of the line, my editor tries to manage line offsets :/ so I get more spaces in between words and failing tests.
Actually, as there is "Soft Wrap" mode in InteliiJ Idea, it should be no problem to read this.

irengrig:
@jin thank you for the review!

@wrowe @keith thank you for confirming the status! (I was also running envoy build locally with this version to be more sure I am not breaking things)

PR #303, opened 2019-08-07
Add -H to find in symlink_contents_to_dir

### LINKED ISSUE #302: configure_in_place broken in bazel 0.28.1
This only happens to `rule_foreign_cc` builds using the configure rule with `configure_in_place` turned on. It fails with an error like this:

```
__________________________PKG_CONFIG_PATH=
bazel-out/k8-opt/bin/external/com_facebook_faiss/faiss/logs/Configure_script.sh: line 74: /tmp/tmp.83MMt3zWJ8/configure: No such file or directory

_____ END BUILD LOGS _____
```

Narrowed it down to bazel changing the http archives in `external` to a symlink, like so:

```
$ ls -ld bazel-u4/external/com_facebook_faiss
lrwxrwxrwx 1 root root 91 Aug  7 01:06 bazel-u4/external/com_facebook_faiss -> /root/.cache/bazel/_bazel_root/129a8b5e1f972fe0890644e192abc210/external/com_facebook_faiss
```

vs bazel 0.27.0:

```
$ ls -ld bazel-ubiquity/external/com_facebook_faiss
drwxr-xr-x 2 root root 4096 Aug  7 01:15 bazel-ubiquity/external/com_facebook_faiss
```

PR incoming.

slam:
This fixes #302

### DIFF
--- a/tools/build_defs/shell_toolchain/toolchains/impl/default_commands.bzl
+++ b/tools/build_defs/shell_toolchain/toolchains/impl/default_commands.bzl
@@ -74,7 +74,7 @@ if [[ -f $1 ]]; then
 fi
 
 if [[ -d $1 || -L $1 ]]; then
-  local children=$(find $1 -maxdepth 1 -mindepth 1)
+  local children=$(find -H $1 -maxdepth 1 -mindepth 1)
   for child in $children; do
     ##symlink_to_dir## $child $target
   done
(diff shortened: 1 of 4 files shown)


irengrig:
Great! Thank you so much!
I see, it works in combination with -maxdepth option - in case when the argument is a symlink.
Merging.

PR #315, opened 2019-09-09
Adapt rules_foreign_cc for working with libtool instead of ar.

irengrig:
- for CMake, pass "<TARGET>" as the output file when forming linker flags with cc_common. CMake will later replace <TARGET> with the actual output file. Fill CMAKE_C_CREATE_ARCHIVE and CMAKE_CXX_CREATE_ARCHIVE CMake variables with 'CMAKE_AR %Bazel-link-flags% <OBJECTS>' call, where CMAKE_AR will take the path to libtool.
Related documentation: https://cmake.org/cmake/help/v3.15/variable/CMAKE_LANG_ARCHIVE_CREATE.html?highlight=cmake_%20lang%20_archive_create

- for configure_make, for the libevent example, we need to skip specifying the libtool from Bazel's toolchain as a linker, because libevent script uses it's own libtool and manages to pass the output file to it, but not to the libtool we are passing. Let it do so as it is a customized script.
Do it with specifying empty string for the $AR environment variable.
Other examples with configure_make work fine.

This fixes https://github.com/bazelbuild/bazel/issues/9258

### DIFF
--- a/.bazelci/config.yaml
+++ b/.bazelci/config.yaml
@@ -1,5 +1,4 @@
 ---
-bazel: last_rc
 tasks:
   ubuntu1804:
     platform: ubuntu1804
--- a/.bazelci/config.yaml
+++ b/.bazelci/config.yaml
@@ -1,4 +1,5 @@
 ---
+bazel: last_rc
 tasks:
   ubuntu1804:
     platform: ubuntu1804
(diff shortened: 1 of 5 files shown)


irengrig:
Hi guys, happy to find the solution for that, thank you Marcel for very good idea!
Please consider reviewing it today, as it is on **Bazel 1.0 hotlist**.

Thank you!!

irengrig:
Just added the commit to use the default Bazel version with the CI.
However, previous CI run demonstrated that rules_foreign_cc works with the Bazel 1.0 rc2.

irengrig:
Thank you for the review!
Submitting, as Jin is on vacation.

PR #317, opened 2019-09-09
Fix make() rule to run make commands sequentially and not in parallel.

### LINKED ISSUE #313: make_commands and create_make_script have a race condition - making build unstable
See:
https://github.com/bazelbuild/rules_foreign_cc/blob/master/tools/build_defs/configure_script.bzl#L58

when multiple `make_commands` are passed to [`create_make_script`](https://github.com/bazelbuild/rules_foreign_cc/blob/master/tools/build_defs/configure_script.bzl#L40), they are joined with an `&`.

This causes the shell to background all commands but the last one. But when the last one completes, the shell moves on, and can exit before all the other commands completed, which in turn can cause the build to fail randomly, depending on how slow/fast those commands are. 

Further, if one of the commands but the last one fails, the error is ignored! And the build goes on.

To see the problem, you can try with a simple shell script:

    $ cat /tmp/test.sh
    #!/bin/bash
    sleep 120 & sleep 1

Run it, and you'll see it exiting in 1 second. Leaving the sleep 120 running in the background.

The most intuitive and least surprising behavior for `make_commands` would be to run the specified commands serially, and fail if any one command fails. Just replace:

    script += ["" + " & ".join(make_commands)]

with:

    script += ["" + " && ".join(make_commands)]

irengrig:
Fixes #313.

### DIFF
--- a/tools/build_defs/configure_script.bzl
+++ b/tools/build_defs/configure_script.bzl
@@ -55,7 +55,7 @@ def create_make_script(
     script += ["echo \"PKG_CONFIG_PATH=$$PKG_CONFIG_PATH$$\""]
 
     script += ["##symlink_contents_to_dir## $$EXT_BUILD_ROOT$$/{} $$BUILD_TMPDIR$$".format(root)]
-    script += ["" + " & ".join(make_commands)]
+    script += ["" + " && ".join(make_commands)]
     return "\n".join(script)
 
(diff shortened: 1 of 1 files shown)


irengrig:
Hi, could you review this, as Jin is on vacation?

irengrig:
Thank you for the review!

PR #318, opened 2019-09-17
Be resilient with Bazel-built transitive dependencies duplicates

irengrig:
- it is possible that some duplicate libraries, built with Bazel, will be passed as inputs to cmake_external or configure_make. rules_foreign_cc should filter duplicates out rather then fail.
- add a test (which just builds the target) //cmake_synthetic:lib_with_duplicate_transitive_bazel_deps

### DIFF
--- a/.bazelci/config.yaml
+++ b/.bazelci/config.yaml
@@ -5,7 +5,6 @@ tasks:
     working_directory: examples
     build_targets:
       - "//cmake_android:app"
-      - "//cmake_synthetic:lib_with_duplicate_transitive_bazel_deps"
     test_targets:
     - "//:tests"
   ubuntu1604:
--- a/.bazelci/config.yaml
+++ b/.bazelci/config.yaml
@@ -5,6 +5,7 @@ tasks:
     working_directory: examples
     build_targets:
       - "//cmake_android:app"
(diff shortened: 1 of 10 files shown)


.bazelci/config.yaml
```diff
@@ -5,6 +5,7 @@ tasks:
     working_directory: examples
     build_targets:
       - "//cmake_android:app"
+      - "//cmake_synthetic:lib_with_duplicate_transitive_bazel_deps"
```
jin: Use https://github.com/bazelbuild/bazel-skylib/blob/master/rules/build_test.bzl and make it a test target?

tools/build_defs/framework.bzl
```diff
@@ -443,6 +443,7 @@ def _copy_deps_and_tools(files):
     return lines
 
 def _symlink_contents_to_dir(dir_name, files_list):
+    files_list = collections.uniq(files_list)
```
jin: Add a comment here?

"It is possible that some duplicate libraries will be passed as inputs to cmake_external or configure_make. Filter duplicates out here."

irengrig:
Thank you so much for so quick review!
I corrected your comments.

PR #322, opened 2019-09-26
For Bazel-built dependencies, do not symlink generic directories under includes

irengrig:
For consuming Bazel-built targets as dependencies, do not use quote_includes from CcInfo, as they are not used for anything related to the particular library, but contain references to the gendir and bin directories, which seems hardly useful. (We expect Bazel targets to describe their include directories and headers explicitly).

Rules_foreign_cc also currently is not using CcInfo.quote_includes.
Having them leads to symlinking directory trees without any reason.
Tests: we already have tests for transitive Bazel dependencies (//cmake_synthetic:test_bazel_transitive_deps, //cmake_with_bazel_transitive:test, //configure_with_bazel_transitive:test)

### DIFF
--- a/tools/build_defs/framework.bzl
+++ b/tools/build_defs/framework.bzl
@@ -450,15 +450,11 @@ def _symlink_contents_to_dir(dir_name, files_list):
         return []
     lines = ["##mkdirs## $$EXT_BUILD_DEPS$$/" + dir_name]
 
-    paths_list = []
     for file in files_list:
-        paths_list += [_file_path(file).strip()]
-
-    for path in paths_list:
-        # Filter out empty subpaths
-        # (current directory may be passed as quote_includes, but we should not symlink it)
-        if path and path != ".":
-            lines += ["##symlink_contents_to_dir## $$EXT_BUILD_ROOT$$/{} $$EXT_BUILD_DEPS$$/{}".format(path, dir_name)]
+      path = _file_path(file).strip()
+      if path:
+        lines += ["##symlink_contents_to_dir## \
+$$EXT_BUILD_ROOT$$/{} $$EXT_BUILD_DEPS$$/{}".format(path, dir_name)]
 
     return lines
 
@@ -615,11 +611,11 @@ def get_foreign_cc_dep(dep):
 
 # consider optimization here to do not iterate both collections
 def _get_headers(compilation_info):
-    # NB: current directory (".") is passed by quote_includes;
-    # ignore it by now, will be filtered by symlinking code.
-    include_dirs = collections.uniq(compilation_info.system_includes.to_list() +
-                                    compilation_info.includes.to_list() +
-                                    compilation_info.quote_includes.to_list())
+    include_dirs = compilation_info.system_includes.to_list() + \
+      compilation_info.includes.to_list()
+    # do not use quote includes, currently they do not contain
+    # library-specific information
+    include_dirs = collections.uniq(include_dirs)
     headers = []
     for header in compilation_info.headers.to_list():
         path = header.path


irengrig:
Friendly ping:)

tools/build_defs/framework.bzl
```diff
@@ -615,11 +611,11 @@ def get_foreign_cc_dep(dep):
...
 def _get_headers(compilation_info):
-    # NB: current directory (".") is passed by quote_includes;
-    # ignore it by now, will be filtered by symlinking code.
-    include_dirs = collections.uniq(compilation_info.system_includes.to_list() +
-                                    compilation_info.includes.to_list() +
-                                    compilation_info.quote_includes.to_list())
+    include_dirs = compilation_info.system_includes.to_list() + \
+      compilation_info.includes.to_list()
+    # do not use quote includes, currently they do not contain
+    # library-specific information
```
hlopko: Can you explain why quote includes shouldn't be added there? I'm not sure I follow... Thank you :)

irengrig: There is the path "." in the quote includes (????!!!!), which brings the whole exec root be symlinked into our build-deps directory, which is also under exec root.
Also, there is bazel-bin directory there - we do not need it there, we go with dependencies and trying to point to them more exactly.
I believe it all should reduce the amount of directories external tools are checking when searching for libraries.

hlopko: To address your surprise about the "." - yes this is by design. Source files and headers are all symlinked into the execroot, and if we want to be able to include the headers with repository relative paths, we have to put the execroot as a quote include directory. Alternative would be to add the actual workspace directory (the one where user develops) as a quote include directory, but that has numerous problems that we happily avoid by having execroot.

We also add "bazel-bin" as an quote include directory. It is important because that's where generated headers end up. People rely on being able to include e.g. generated protobuf headers just like they were next to non-generated headers.

It is true that currently we allow using include paths like "bazel-out/foo/bar/baz.h" or "external/repo1/qux/asdf.h". Those are problems that should be (and will be) solved elsewhere. Well behaving user will use "foo/bar/baz.h" and "qux/asdf.h" paths.

Did you observe slow builds because of too many include directories? Or is there some other motivation?

irengrig: Thank you for detailed explanation, I was guessing about that.
Well, first of all, current code is creating a cyclic symlink. - so it rules_foreign_cc should use this information, then rather in flags, but not for symlinking...

Using the exec root for getting headers/libraries does not make much sense for sandboxed execution, because the sources rules_foreign_cc is dealing with rely on external build script, which has access to the sources by definition... (it is a part of it)
For *not* sandboxed execution it would link everything what was built in this workspace before, and it will makes a build at least not reproducible... - and the same is true about bazel-bin - at least the contents of it can be described in terms of bazel dependencies...
This all looks very much like non hermetic hacks...

hlopko: Caveat - I'm talking in general, not specifically about rules_foreign_cc. 

I believe adding the execroot as a quote include directory still makes sense in the sandboxed execution. Gcc convention is to look for headers first in the directory of the source file being compiled, then in the directories specified by all those command line flags, and then there is also a predefined list of system installed directories.

Important thing is that the current working directory is not implicitly added to the header search path.

It is true that rules_foreign_cc **might** not need to care about include directories if rules_foreign_cc cannot depend on other C++ rules. I don't know if they can. If all sources and headers that rules_foreign_cc deals with are part of a single target, of course the build script can make assumptions. I didn't try to reason about this specific situation.

Maybe I don't understand your point about non-sandboxed execution, is it only that non-sandboxed execution allows actions to access the entire execroot? Yes that is the case, and that's the reason why we are willing to pay the performance cost of the sandboxed execution. It is also not specific to C++, but affects all actions Bazel runs.

I don't share your perspective that this looks like hacks, but I agree that it is complicated. But upon a closer look things start to make sense and good engineering reasons are discovered.

irengrig: Thank you for the answer!
In the case of rules_foreign_cc, current directory (where the config file is (CMakeLists.txt/Makefile/configure) and gcc/clang is actually run) is not an execroot: it is the directory, where there are sources that we want to build with the external system, like the root directory of the external repository.

>if rules_foreign_cc cannot depend on other C++ rules. 

rules_foreign_cc can depend on other rules_foreign_cc and cc_library rules, but they are NOT using quote_includes, I checked that in the code.

>I don't share your perspective that this looks like hacks, but I agree that it is complicated. But upon a closer look things start to make sense and good engineering reasons are discovered.

Sorry, I meant from the perspective of a rules_foreign_cc linking everything from bazel-bin is too much of an assumption. I think it is good we are removing this.

However, by the reasons you described, we might want to point the external system into bazel-bin using the corresponding flags/environment variables, specific to the system (CMake, GNU Make etc). Currently however I am not aware of this being needed.

PR #323, opened 2019-09-27
Improve cygwin compatibility

core-process:
Use `cygpath` in case it is available, `pwd` otherwhise. `pwd -W` does not work within Cygwin.


core-process:
@googlebot I signed it!

irengrig:

PR #328, opened 2019-10-11
Prefixes bazel_version so that it doesn't conflict with other libraries.

### LINKED ISSUE #327: Conflicting bazel_version definitions
`upb` and `rules_foreign_cc` both define `bazel_version` files and they conflict with eachother.

https://github.com/bazelbuild/rules_foreign_cc/blob/a0dc109915cea85909bef586e2b2a9bbdc6c8ff5/workspace_definitions.bzl#L91

https://github.com/protocolbuffers/upb/blob/master/bazel/workspace_deps.bzl#L7

I'm guessing they should be prefixed with the library name, e.g. upb_bazel_version.

Dig-Doug:
Fixes #327.


irengrig:

PR #330, opened 2019-10-21
Create directory structure recursively in symlink_contents_to_dir

schoppmp:
This allows to have multiple bazel-built dependencies that have files in the same directories. For example, different Boost libraries built with [rules_boost](https://github.com/nelhage/rules_boost) have headers in `include/boost/detail`, and so adding those would not work as multiple `detail` symlinks would have needed to point to different directories.


schoppmp:
Fixed in 74b146dc87d37baa1919da1e8f7b8aafbd32acd9.

PR #340, opened 2019-11-18
Fix the use_malloc example so that it works on Mac & Linux

nomis52:


nomis52:
@googlebot I signed it!

nomis52:
I'll rebase this once 339 lands and include it in the `mac_tests`.

irengrig:
Thank you for contribution.
However, this example demonstrates how possibly configure_make could have been used to assemble malloc implementation, once Bazel allows inputs from targets other then cc_library (at least cc_import).
That is why removing malloc does not make sense.

PR #341, opened 2019-11-19
Mirror all http_archives.

philwo:
Context: https://github.com/bazelbuild/bazel/issues/10270

### DIFF
--- a/examples/examples_repositories.bzl
+++ b/examples/examples_repositories.bzl
@@ -7,16 +7,20 @@ def include_examples_repositories():
         name = "libevent",
         build_file_content = all_content,
         strip_prefix = "libevent-2.1.8-stable",
-        urls = ["https://github.com/libevent/libevent/releases/download/release-2.1.8-stable/libevent-2.1.8-stable.tar.gz"],
+        urls = [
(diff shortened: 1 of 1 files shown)


examples/examples_repositories.bzl
```diff
@@ -7,16 +7,20 @@ def include_examples_repositories():
...
+            "https://mirror.bazel.build/github.com/libevent/libevent/releases/download/release-2.1.8-stable/libevent-2.1.8-stable.tar.gz",
+            "https://github.com/libevent/libevent/releases/download/release-2.1.8-stable/libevent-2.1.8-stable.tar.gz",
+        ],
     )
 
     http_archive(
         name = "zlib",
         build_file_content = all_content,
-        sha256 = "4ff941449631ace0d4d203e3483be9dbc9da454084111f97ea0a2114e19bf066",
+        sha256 = "c3e5e9fdd5004dcb542feda5ee4f0ff0744628baf8ed2dd5d66f8ca1197cb1a1",
```
irengrig: Is that correct? SHA change? The version did not change.

philwo: I changed the URL from the .tar.xz to the .tar.gz archive, because we currently do not mirror the tar.xz one and due to the network outage I cannot mirror it :sweat_smile:

irengrig: ok, I see, sorry

PR #358, opened 2020-01-01
Add a new cmake option 'cmake_relative_root'.

stellaraccident:
This makes it possible to build LLVM (by passing "llvm" for this new
option).

### DIFF
--- a/tools/build_defs/cmake.bzl
+++ b/tools/build_defs/cmake.bzl
@@ -58,6 +58,8 @@ def _create_configure_script(configureParameters):
     inputs = configureParameters.inputs
 
     root = detect_root(ctx.attr.lib_source)
+    if ctx.attr.cmake_relative_root:
+        root += "/" + ctx.attr.cmake_relative_root
 
     tools = get_tools_info(ctx)
     # CMake will replace <TARGET> with the actual output file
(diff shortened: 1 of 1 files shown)


lopter:
Hi, thanks for the patch! I have also used that to build libdbus 1.12.16 which doesn't have its `CMakeLists.txt` at the root.

attilaolah:
Same thing for `ztd`, it has its `CMakeLists.txt` under `build/cmake`.

I was thinking that it would also be possible to find the most root-level `CMakeLists.txt`, similarly to how `find_root` works. That would make most of these packages work without needing to specify cmake_relative_root.

irengrig:
Hi @stellaraccident , thank you for the pull request!
Is that possible to see the example of the usage?

And also, we should change the comment about the "root" parameter in create_cmake_script in cmake_script.bzl to "working directory for invoking cmake and build tool (make/Ninja/MSBuild)"

irengrig:
Implemented in #390: https://github.com/bazelbuild/rules_foreign_cc/blob/d838bc55067fc473fa67ec619be1478e45b96e2d/tools/build_defs/cmake.bzl#L119.

Sorry, I did not connect this PR to the problem at the moment when I did #390.

PR #362, opened 2020-01-17
Double quotes are not properly escaped

slam:
Bazel passes to CFLAGS and CXXFLAGS the redacted values of `__DATE__`,
`__TIMESTAMP__`, and `__TIME__`.

Just a single backslash is not enough. It gets passed into the code
as redacted (without any quotes) and not "redacted". The compiler
complains about the symbol "redacted" not being found for packages
that actually use one of those fields.

Remove hack in apache_httpd/BUILD that was put in place because of this
bug.

Also need to turn off escaping on heredoc when generating
crosstool_bazel.cmake. Without the single quotes around EOF, the
contents of the heredoc are escaped. With single quotes, they are not.

https://stackoverflow.com/a/9870274

### DIFF
--- a/examples/apache_httpd/BUILD
+++ b/examples/apache_httpd/BUILD
@@ -31,7 +31,6 @@ configure_make(
         "--with-apr=$$EXT_BUILD_DEPS$$/apr",
         "--with-apr-util=$$EXT_BUILD_DEPS$$/apr_util",
         "--with-pcre=$$EXT_BUILD_DEPS$$/pcre",
-        "CFLAGS='-Dredacted=\"redacted\"'",
     ],
     lib_source = "@apache_httpd//:all",
     binaries = [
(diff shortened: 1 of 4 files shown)


slam:
Some tests are failing with this error:

```
/usr/bin/gcc   -U_FORTIFY_SOURCE -fstack-protector -Wall -Wunused-but-set-parameter -Wno-free-nonheap-object -fno-omit-frame-pointer -fno-canonical-system-headers -Wno-builtin-macro-redefined -D__DATE__=\;redacted\" -D__TIMESTAMP__=\"redacted\" -D__TIME__=\"redacted\"" -o CMakeFiles/cmTC_de405.dir/testCCompiler.c.o   -c testCCompiler.c
```

Note the semicolon here: `-D__DATE__=\;redacted\"`. It should have been a double quote.

It could be a bug in a version of bazel (or its toolchains) that put the semicolon there. I searched the code base and wasn't able to find it.

Would appreciate some guidance on this.

rockwotj:
@slam and @irengrig is there anything I can do to help move this PR along?

UebelAndre:
This seems like a great change, could I get you to resolve the conflicts? I'd be happy to merge otherwise 😄

UebelAndre:
This has been implemented in https://github.com/bazelbuild/rules_foreign_cc/pull/567

slam:
My bad. Totally missed the notification of your comment back in Jan. Thank you for taking care of it @UebelAndre.

PR #363, opened 2020-01-18
Copy foreign_cc dependencies. Symlink is not enough.

### LINKED ISSUE #344: Transitive configure_make builds are broken on Darwin
# Description

A chain of 3 or more libraries fails to build on Darwin. Imagine we have the following:

`libvolume` -> `libarea` -> `liblength`

Then on Darwin, `libvolume` fails to build with:

```
EXT_BUILD_DEPS=/var/folders/qh/c65tnjdn605_7hw2_85jxrtw0000gn/T/tmp.bIJ95LxY
../libtool: line 8985: cd: ${EXT_BUILD_DEPS}/liblength/lib: No such file or directory
libtool: warning: cannot determine absolute directory name of '${EXT_BUILD_DEPS}/liblength/lib'
grep: ${EXT_BUILD_DEPS}/liblength/lib/liblength.la: No such file or directory
sed: ${EXT_BUILD_DEPS}/liblength/lib/liblength.la: No such file or directory
```

# Cause

From what I can tell this is a result of the difference in the libtool `.la` files built on each platform.

After the `build` step for `libarea` completes, the `libarea.la` file contains:

Linux:
`dependency_libs=' -lstdc++ -lm -L${EXT_BUILD_DEPS}/liblength/lib'`

Darwin:
`dependency_libs=' ${EXT_BUILD_DEPS}/liblength/lib/liblength.la -lstdc++ -lm -L${EXT_BUILD_DEPS}/liblength/lib'`

When the `libvolume` build runs and we try to link, libtool (via ltmain.sh) executes the following

```
      for deplib in $dependency_libs; do
        path=
        case $deplib in
        -L*) path=$deplib ;;
        *.la)
          func_resolve_sysroot "$deplib"
          deplib=$func_resolve_sysroot_result
          func_dirname "$deplib" "" "."
         dir=$func_dirname_result                          # Line  8985 
         # We need an absolute path.
         case $dir in
            [\\/]* | [A-Za-z]:[\\/]*) absdir=$dir ;;
```

This doesn't understand env vars, in this case $EXT_BUILD_DEPS for `*.la` libs. 

# Steps to Reproduce

Use the `configure` branch of https://github.com/nomis52/rules_foreign_cc

```
cd examples/configure_transitive/
./build_tarballs.sh . # This will build tarballs for liblength, libarea & libvolume & put them in /tmp
bazel build -s --sandbox_debug  //configure_transitive:libvolume 
```

slam:
A foreign_cc-built dep could be included by multiple other foreign_cc
targets. If these other targets are built in parallel, they could update
the dep's BAZEL_GEN_ROOT values inside the .pc, .cmake, etc.,
simulataneously. This could break the build in unpredictable ways.

To fix this, each target must have a *copy* of the dep. A symlink of the
dep is not sufficient.

Make the copy owner writable. This fixes the sed permission error on
MacOS where the files in bazel cache dir are write-once only.

This also fixes #344, where ${EXT_BUILD_DEPS} env var is not expanded
by pkgconfig or cmake files.

This fix also effectively reverted #324

### DIFF
--- a/tools/build_defs/configure_script.bzl
+++ b/tools/build_defs/configure_script.bzl
@@ -18,7 +18,7 @@ def create_configure_script(
 
     script = []
     for ext_dir in inputs.ext_build_dirs:
-        script += ["##increment_pkg_config_path## $$EXT_BUILD_ROOT$$/" + ext_dir.path]
+        script += ["##increment_pkg_config_path## $$EXT_BUILD_DEPS$$/" + ext_dir.basename]
 
     script += ["echo \"PKG_CONFIG_PATH=$$PKG_CONFIG_PATH$$\""]
 
@@ -50,7 +50,7 @@ def create_make_script(
     env_vars_string = get_env_vars(workspace_name, tools, flags, user_vars, deps, inputs)
     script = []
     for ext_dir in inputs.ext_build_dirs:
-        script += ["##increment_pkg_config_path## $$EXT_BUILD_ROOT$$/" + ext_dir.path]
+        script += ["##increment_pkg_config_path## $$EXT_BUILD_DEPS$$/" + ext_dir.basename]
 
     script += ["echo \"PKG_CONFIG_PATH=$$PKG_CONFIG_PATH$$\""]
 
--- a/tools/build_defs/framework.bzl
+++ b/tools/build_defs/framework.bzl
@@ -246,6 +246,8 @@ def cc_external_rule_impl(ctx, attrs):
         "##mkdirs## $$INSTALLDIR$$",
         _print_env(),
         "\n".join(_copy_deps_and_tools(inputs)),
+        # replace placeholder with the dependencies root
+        "##define_absolute_paths## $$EXT_BUILD_DEPS$$ $$EXT_BUILD_DEPS$$",
         "cd $$BUILD_TMPDIR$$",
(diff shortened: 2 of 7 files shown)


nomis52:
@irengrig , do you mind reviewing. Apparently this fixes my issue with transitive builds.

oquenchil:
Could you please add a test for the command you added like [here](https://github.com/bazelbuild/rules_foreign_cc/blob/594bf4d7731e606a705f3ad787dd0a70c5a28b30/test/BUILD#L14)?

Thanks!

oquenchil:
Commenting here because GH doesn't allow commenting on [collapsed lines](https://github.com/isaacs/github/issues/1655).

In framework.bzl, please update comment  "symlinked install directories of the dependencies".

jheaff1:
This may fix an issue I am having also. I am trying to build log4cxx, which depends on apr and apr-util. apr-util itself depends on apr. I am using the configure_make rule and I'm getting the following error:

/bin/grep: ${EXT_BUILD_DEPS}/apr/lib/libapr-1.la: No such file or directory
/bin/sed: can't read ${EXT_BUILD_DEPS}/apr/lib/libapr-1.la: No such file or directory

jsharpe:
I believe #635 should have resolved the underlying issue that this patch was setting out to fix

PR #367, opened 2020-02-06
Build under bazel sandbox instead of /tmp, and remove redundant printf

scele:
Build artifacts under /tmp are not cleaned up after a failed build, for
the ease of debugging.  However, this can quickly fill up the /tmp partition
if the build is large.  Build under the sandbox directory instead, and
rely on standard bazel cleanup scheme.

Also remove a print from the cleanup function, which somehow causes empty
newlines to be printed in bazel build output.

### DIFF
--- a/tools/build_defs/framework.bzl
+++ b/tools/build_defs/framework.bzl
@@ -231,9 +231,9 @@ def cc_external_rule_impl(ctx, attrs):
     define_variables = [
         set_cc_envs,
         "export EXT_BUILD_ROOT=##pwd##",
-        "export BUILD_TMPDIR=##tmpdir##",
-        "export EXT_BUILD_DEPS=##tmpdir##",
         "export INSTALLDIR=$$EXT_BUILD_ROOT$$/" + empty.file.dirname + "/" + lib_name,
+        "export BUILD_TMPDIR=$$INSTALLDIR$$/_build_tmpdir",
+        "export EXT_BUILD_DEPS=$$INSTALLDIR$$/_ext_build_deps",
     ]
 
     script_lines = [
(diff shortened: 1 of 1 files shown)


tools/build_defs/framework.bzl
```diff
@@ -331,8 +333,7 @@ def wrap_outputs(ctx, lib_name, configure_name, script_text):
     cleanup_on_success_function = create_function(
         ctx,
         "cleanup_on_success",
-        """##echo## \"rules_foreign_cc: Cleaning temp directories\"
-rm -rf $BUILD_TMPDIR $EXT_BUILD_DEPS""",
+        """rm -rf $BUILD_TMPDIR $EXT_BUILD_DEPS""",
```
attilaolah, line 336: @oquenchil would it be possible to at least cherry-pick this part of the pull request into master?

Then we can deal about the rest separately. (Although the rest of this PR would also be really nice to have, the leftover temp directories make things hard to debug.)

jsharpe:
Thanks for the patch @scele.

Unfortunately this patch seems to introduce a symlink loop when copying the deps. Testing it locally on some of my builds I also see issues with path lengths in the find call in`replace_in_files`.

Is this something you could take a look at again?

scele:
Sorry, I’m no longer using rules_foreign_cc in my projects, so I don’t have an environment to work on improving this change. :( I’ll close this PR but feel free to take the changes and improve them in some other PR!

attilaolah:
I might have time to cherry-pick some of these changes into a new PR this weekend, if that's OK.

jsharpe:
That'd be great @attilaolah if you could!

tools/build_defs/framework.bzl
```diff
@@ -231,9 +231,9 @@ def cc_external_rule_impl(ctx, attrs):
     define_variables = [
         set_cc_envs,
         "export EXT_BUILD_ROOT=##pwd##",
-        "export BUILD_TMPDIR=##tmpdir##",
-        "export EXT_BUILD_DEPS=##tmpdir##",
         "export INSTALLDIR=$$EXT_BUILD_ROOT$$/" + empty.file.dirname + "/" + lib_name,
+        "export BUILD_TMPDIR=$$INSTALLDIR$$/_build_tmpdir",
+        "export EXT_BUILD_DEPS=$$INSTALLDIR$$/_ext_build_deps",
```
attilaolah, line 236: Actually I think it would be nice if this was outside of `$$INSTALLDIR$$`, within `$$EXT_BUILD_ROOT$$` (i.e. `##pwd##`). That way we don't need to care about clashing with source files. Additionally, we could still use a tempdir under this prefix.

Issue #365, opened 2020-02-05
Problems with pkg-config on transitive builds


flichtenheld:
Hi.

I have a configure_make build that tries to run pkg-config against a component built with a different configure_make. But it fails with:

```
checking for libmnl... no
configure: error: Package requirements (libmnl >= 1) were not met:

Variable 'EXT_BUILD_DEPS' not defined in '/home/flichtenheld/.cache/bazel/_bazel_flichtenheld/a7def64fd52f4e313e77afa5a69def9f/sandbox/processwrapper-sandbox/46/execroot/__main__/bazel-out/k8-fastbuild/bin/libs/mnl/copy_mnl/mnl/lib/pkgconfig/libmnl.pc'

```

I fixed it by adding an additional statement to the configure_script.bzl like this:

```
script += ["export PKG_CONFIG=\"pkg-config --define-variable=EXT_BUILD_DEPS=$$EXT_BUILD_DEPS$$\""]
```

This resolves the issue for me and looks to me like good default behavior. But before opening a pull request for this I wanted to make sure I didn't miss a better solution.

Regards,
  Frank

jheaff1:
This seems to be related to #344

rajachan:
I've stumbled into a manifestation of the same issue. Are the maintainers planning to review/merge your proposed fix, @flichtenheld?

jsharpe:
Unfortunately without a sign off on the CLA we can't merge the change. It also needs rebasing on top of main as the file structure has changed since this PR was made.

Issue #368, opened 2020-02-06
rules_foreign_cc failing with VS 2017


meteorcloudy:
We recently removed VS 2015 installation on Bazel CI's Windows machine, rules_foreign_cc started to fail with the following error:
```
__________________________CMake Error at CMakeLists.txt:6 (project):
  Failed to run MSBuild command:
 
    MSBuild.exe
 
  to get the value of VCTargetsPath:
 
    Microsoft (R) Build Engine version 15.9.21+g9802d43bc3 for .NET Framework
    Copyright (C) Microsoft Corporation. All rights reserved.
 
    Build started 2/6/2020 11:56:22 AM.
    Project "C:\tools\msys64\tmp\tmp.dfEQx2RVrd\CMakeFiles\3.16.2\VCTargetsPath.vcxproj" on node 1 (default targets).
    C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\Common7\IDE\VC\VCTargets\Microsoft.Cpp.Platform.targets(67,5): error MSB8020: The build tools for v140 (Platform Toolset = 'v140') cannot be found. To build using the v140 build tools, please install v140 build tools.  Alternatively, you may upgrade to the current Visual Studio tools by selecting the Project menu or right-click the solution, and then selecting "Retarget solution". [C:\tools\msys64\tmp\tmp.dfEQx2RVrd\CMakeFiles\3.16.2\VCTargetsPath.vcxproj]
    Done Building Project "C:\tools\msys64\tmp\tmp.dfEQx2RVrd\CMakeFiles\3.16.2\VCTargetsPath.vcxproj" (default targets) -- FAILED.
 
    Build FAILED.
````

It looks like CMake is still trying to locate MSVC 2015 (v140). 
@irengrig Is there any way to tell Cmake to use MSVC 2017 only?

irengrig:
Fixed by #370

PR #369, opened 2020-02-06
configure_script: make sure to define EXT_BUILD_DEPS for pkg-config

flichtenheld:
Because we insert that variable into .pc files we build.
So if other packages try to use them it needs to be defined.

My proposed fix that I mentioned in issue #365


flichtenheld:
My company doesn't have a Google CLA, yet. So this might take time to sort out.

flichtenheld:
FWIW, the build failure is #368 and not related to my changes

jheaff1:
Any progress on this PR? I’m happy to take it over if need be

PR #373, opened 2020-02-27
Test for detect_root function and example of usage

irengrig:


irengrig:
In #259, the problem is represented: is detect_root correctly defining the root directory for sources, if sources come from a filegroup with the glob from some directory of the same workspace (not an external repository).
Apparently, the glob should use *exclude_directories = 0* attribute specification - then also the wrapping directory is included into the glob, and detect_root finds it correctly.

Having the tests for detect_root is also very important for this ruleset.

irengrig:
Friendly ping :)

PR #377, opened 2020-03-06
Fix shell function symlink_contents_to_dir, add test

irengrig:
- test for the case when we symlink the contents of two directories with the same inner directories (include)
- the test data of the test for shell script helper also changes, as it uses the real shell toolchain implementation text
- Fixes the issue from #330

Based on code from #375


ramrengaswamy:
Are there any plans to merge this change? Hard to use `glog` and `gflags` as a dependency to any `cmake_external` rule without this fix.

PR #378, opened 2020-03-07
Detect CMakeLists.txt when building CMake projects.

attilaolah:
Instead of pointing `cmake` to the source root, detect the topmost directory that contains a `CMakeLists.txt` file and point `cmake` there. This is required for projects that do not ship a `CMakeLists.txt` in their root directory, e.g. `lz4`, `zstd` or `llvm`.

This mostly obsoletes https://github.com/bazelbuild/rules_foreign_cc/pull/358, although an argument like `cmake_relative_root` might still be required for projects that ship multiple top-level directories containing a `CMakeLists.txt` (but then again an alternative there would be to just delete the rest from the repository using `patch_cmds`.)

This should be a backwards-compatible change and no update to the documentation should be necessary.

### DIFF
--- a/tools/build_defs/cmake.bzl
+++ b/tools/build_defs/cmake.bzl
@@ -57,7 +57,7 @@ def _create_configure_script(configureParameters):
     ctx = configureParameters.ctx
     inputs = configureParameters.inputs
 
-    root = detect_root(ctx.attr.lib_source)
+    root = detect_root(ctx.attr.lib_source, contains_file="CMakeLists.txt")
 
     tools = get_tools_info(ctx)
     # CMake will replace <TARGET> with the actual output file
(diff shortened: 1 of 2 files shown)


irengrig:

I recently added working directory attribute for cmake_external in #390.
I feel like it is better to explicitly point to the build directory, as it may be several sibling subdirectories in the project with CMakeLists.txt - I would not like to make it the default behavior.

We may, however, add one more attribute 'detect_build_dir', which would enable automatic detection.
Also, it would be nice to have some tests - it should be easy to create some directory structure for the test, like here: https://github.com/bazelbuild/rules_foreign_cc/blob/d838bc55067fc473fa67ec619be1478e45b96e2d/test/BUILD#L36.

attilaolah:
Actually, the `working_directory` attribute you added in #390 works perfectly for my use case, I don't think I need this change any more.

`detect_build_dir` would be neat, but as long as it's not the default, I don't see too much value in it: if I have to specify some attribute, I might as well specify the build dir.

I'm withdrawing this PR, thanks for fixing the issue though!

PR #379, opened 2020-03-08
Add a GNU Make toolchain

### LINKED ISSUE #372: Consider adding toolchains for configure/make
The fact that `cmake` is a toolchain makes it super easy to add a drop-in replacement toolchain that allows us to use e.g. `emcmake` from the Emscripten wrapper, for example [like so](https://github.com/attilaolah/wasm/blob/master/toolchains/emcmake/BUILD.bazel).

Even `make` can be "fixed" by passing in `make_commands`:

```py
make_commands = [
    "emcmake make",
    "emcmake make install",
],
```

Although the above is kind of ugly, because now we need to `select()` and pass in the custom `make_commands` depending on some predicate that in turn depends on the toolchain being used. Hence it would be nice to register a custom toolchain of type make, just like we have for type cmake.

However, it doesn't seem to be possible to use `emconfigure /path/to/configure` instead of `configure`, since the macro only allows us to use a relative path. I think the cleanest solution here too would be to turn `configure_make` into a rule that requires the `configure` and `make` toolchains, and then we could provide our own command for either of these, as part of the toolchain implementation.

Would you be willing to accept a pull request that introduces additional toolchains?

attilaolah:
This change adds a new toolchain type, `make_toolchain`. I have also added an implementation that uses the pre-installed toolchain (no shipped `make` yet, will add that shortly too.)

The goal is to use the `make` binary path provided by the toolchain. This way, one can configure a custom toolchain that, for example, runs `emmake make` instead of `make`, to target Emscripten. (At least this is my use case). Another use case is to use a vendored Make, instead of relying on the system `make` binary in the execution environment.

TODO:

- [x] Add a bootstrap rule for vendored `make`, just like the one for `cmake` and `ninja`.

Fixes #372.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,15 +88,20 @@ load("@rules_foreign_cc//:workspace_definitions.bzl", "rules_foreign_cc_dependen
 #
 #  Args:
 #    native_tools_toolchains: pass the toolchains for toolchain types
+#      '@rules_foreign_cc//tools/build_defs:make_toolchain',
 #      '@rules_foreign_cc//tools/build_defs:cmake_toolchain' and
 #      '@rules_foreign_cc//tools/build_defs:ninja_toolchain' with the needed platform constraints.
 #      If you do not pass anything, registered default toolchains will be selected (see below).
 #  
-#    register_default_tools: if True, the cmake and ninja toolchains, calling corresponding
-#      preinstalled binaries by name (cmake, ninja) will be registered after
+#    register_default_tools: if True, the make, cmake and ninja toolchains, calling corresponding
+#      preinstalled binaries by name (make, cmake, ninja) will be registered after
 #      'native_tools_toolchains' without any platform constraints.
 #      The default is True.
-rules_foreign_cc_dependencies(["//:my_cmake_toolchain", "//:my_ninja_toolchain"])
+rules_foreign_cc_dependencies([
+    "//:my_make_toolchain",
+    "//:my_cmake_toolchain",
+    "//:my_ninja_toolchain",
+])
 
 # OpenBLAS source code repository
 http_archive(
--- a/for_workspace/make_build.bzl
+++ b/for_workspace/make_build.bzl
@@ -1,0 +1,47 @@
+""" Rule for building GNU Make from sources. """
+
+load("//tools/build_defs:detect_root.bzl", "detect_root")
+load("@rules_foreign_cc//tools/build_defs:shell_script_helper.bzl", "convert_shell_script")
+
+def _make_tool(ctx):
+    root = detect_root(ctx.attr.make_srcs)
+
+    make = ctx.actions.declare_directory("make")
+    script = [
+        "export BUILD_DIR=##pwd##",
+        "export BUILD_TMPDIR=##tmpdir##",
+        "##copy_dir_contents_to_dir## ./{} $BUILD_TMPDIR".format(root),
+        "cd $$BUILD_TMPDIR$$",
+        "./configure --prefix=$$BUILD_DIR$$/{}".format(make.path),
+        "./build.sh",
+        "./make install",
+    ]
+    script_text = convert_shell_script(ctx, script)
+
+    ctx.actions.run_shell(
+        mnemonic = "BootstrapMake",
+        inputs = ctx.attr.make_srcs.files,
+        outputs = [make],
+        tools = [],
+        use_default_shell_env = True,
+        command = script_text,
+        execution_requirements = {"block-network": ""},
+    )
+
+    return [DefaultInfo(files = depset([make]))]
+
+""" Rule for building Make. Invokes configure script and make install.
+  Attributes:
+    make_srcs - target with the Make sources
+"""
+make_tool = rule(
+    attrs = {
+        "make_srcs": attr.label(mandatory = True),
+    },
+    fragments = ["cpp"],
+    output_to_genfiles = True,
+    implementation = _make_tool,
+    toolchains = [
+        "@rules_foreign_cc//tools/build_defs/shell_toolchain/toolchains:shell_commands",
+    ],
+)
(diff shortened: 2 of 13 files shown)


attilaolah:
Looks like some of the tests require an actual `make` toolchain on the CI?

Hold on, I'll add a bootstrap for `make` so that the builds should pass, and rebase accordingly.

attilaolah:
Hmm. I'm having trouble figuring out what tests are actually run on the CI. It fails with this error message:

```
(01:34:03) ERROR: While resolving toolchains for target //cmake_android:libhello: no matching toolchains found for types @rules_foreign_cc//tools/build_defs:make_toolchain
(01:34:03) ERROR: Analysis of target '//cmake_android:app' failed; build aborted: no matching toolchains found for types @rules_foreign_cc//tools/build_defs:make_toolchain
```

However, I don't see where is the workspace that would contain this target (`//cmake_android:app`). Maybe it passes in custom toolchains, and therefore now it should pass in a custom `make` toolchain as well?

I guess that means this change isn't that much backwards compatible, at least not for the folks who use their own toolchains, as they would now need a `make` toolchain as well.

attilaolah:
Ideally the `make` toolchain should also be used for `BootstrapCMake` so that these rules could be used on a system without `cmake` or `make`, which would be the case if for example someone wants to ship and bootstrap their own compiler.

attilaolah:
Please hold on before merge, I'd like to change the history to make sure intermediate commits are also green.

tools/build_defs/cmake.bzl
```diff
@@ -19,13 +19,19 @@ load(
...
+    "get_make_data",
+)
 load("@rules_foreign_cc//tools/build_defs:shell_script_helper.bzl", "os_name")
 
 def _cmake_external(ctx):
     cmake_data = get_cmake_data(ctx)
+    make_data = get_make_data(ctx)
 
-    tools_deps = ctx.attr.tools_deps + cmake_data.deps
+    tools_deps = ctx.attr.tools_deps + cmake_data.deps + make_data.deps
```
jsharpe: This assumes make is being used which isn't always the case - the ninja generator may be being used.

attilaolah: Just noticed this comment today. I'll look into it — ideally we'd know which dependencies to add and only add those.

attilaolah: PTAL.

attilaolah:
Is there anything else you'd like me to do here? Otherwise this should be ready for a review.

achew22:
Friendly ping @jsharpe, this patch would solve a problem I've been having in my repository.

jsharpe:
@achew22 I don't have connit rights. @irengrig is the repo owner AFAIK

attilaolah:
PTAL @irengrig. I rebased to upstream head and resolved the comments.

README.md
```diff
@@ -88,15 +88,20 @@ load("@rules_foreign_cc//:workspace_definitions.bzl", "rules_foreign_cc_dependen
...
 #  
-#    register_default_tools: if True, the cmake and ninja toolchains, calling corresponding
-#      preinstalled binaries by name (cmake, ninja) will be registered after
+#    register_default_tools: if True, the make, cmake and ninja toolchains, calling corresponding
+#      preinstalled binaries by name (make, cmake, ninja) will be registered after
 #      'native_tools_toolchains' without any platform constraints.
 #      The default is True.
-rules_foreign_cc_dependencies(["//:my_cmake_toolchain", "//:my_ninja_toolchain"])
+rules_foreign_cc_dependencies([
+    "//:mi_make_toolchain",
```
oquenchil, line 101: my_make_toolchain*

for_workspace/make_build.bzl
```diff
@@ -0,0 +1,47 @@
+""" Rule for building GNU Make from sources. """
```
oquenchil, line 1: So in the equivalent cmake_build.bzl, you are still using the make from the system not this one. That's WAI isn't it?

for_workspace/make_build.bzl
```diff
@@ -0,0 +1,47 @@
...
+        execution_requirements = {"block-network": ""},
+    )
+
+    return [DefaultInfo(files = depset([make]))]
+
+""" Rule for building Make. Invokes configure script and make install.
+  Attributes:
+    make_srcs - target with the Make sources
+"""
+make_tool = rule(
```
oquenchil, line 37: Can you add an example to verify manually that it works just like we do for [cmake](https://github.com/bazelbuild/rules_foreign_cc/blob/006d5f6f8b6f6922943add61fcf038a4a1d802e1/examples/build_cmake_itself/BUILD#L3)?

oquenchil (changes requested):
This is great. Thank you very much!

I just have a couple of nits.

README.md
attilaolah: Done.

for_workspace/make_build.bzl
attilaolah: Done in 16778ca.

Turns out one needs to install `automake` and `texinfo` on Debian to get Make working, but that should be fine: the intention of this toolchain is to make it replaceable, so that I can swap in my Emscripten wrapper. When the dependencies are installed, the toolchain can be bootstrapped by Bazel.

(Tested on Debian Buster on o ChromeOS device using Crostini.)

for_workspace/make_build.bzl
attilaolah: Well, I tried to update that to use the Make toolchain, but I couldn't get the dependencies right. I think it's worth trying that once more, but for the purpose of this pull request, that can considered an additional feature.

Especially given that we cannot bootstrap Make from absolutely nothing: we still need basic build tools like `automake` and `texinfo`. (Maybe bootstrap-from-nothing is also a feature to consider, but for me it's not particularly interesting: I mostly added the bootstrapped make for completeness. The main feature is to have the Make toolchain so that it could be swapped with another toolchain.)

oquenchil (approved):
LGTM

PR #381, opened 2020-03-08
Update CMake version

attilaolah:
Update CMake to 3.17.2. Also add a `sha256` sum to the `http_archive` hich was missing.


attilaolah:
@mellery451 updated commit message as well as PR description, and updated to 3.17.2 while there. Also rebased to current upstream master.

PR #383, opened 2020-03-11
Add make_env_vars for make call as well

artemyarulin:
`env_vars_string` already used in `create_configure_script`, but in `create_make_script` function `env_vars_string` variable exists, but unused.

And overall I guess it's correct approach to pass env variables to make command as well so users can configure it's behaviour.

Use case and test example is rocksdb
```
make(
    name = "rocksdb",
    lib_source = "@rocksdb//:all",
    # Current workaround is to override make_commands and pass env variables directly
    make_commands = ["ROCKSDB_DISABLE_ZLIB=1 ROCKSDB_DISABLE_BZIP=1 make -j 6 static_lib",
                     "cp librocksdb.a $$INSTALLDIR$$/lib/librocksdb.a",
                     "cp -RL include $$INSTALLDIR$$",
    ],
    # Issue is that following env not set for make command
    make_env_vars = {
        "ROCKSDB_DISABLE_ZLIB": "1",
        "ROCKSDB_DISABLE_BZIP": "1",
    },
    static_libraries = ["librocksdb.a"],
)
```

### DIFF
--- a/tools/build_defs/configure_script.bzl
+++ b/tools/build_defs/configure_script.bzl
@@ -48,7 +48,7 @@ def create_make_script(
         make_commands,
         prefix):
     env_vars_string = get_env_vars(workspace_name, tools, flags, user_vars, deps, inputs)
-    script = []
+    script = [env_vars_string]
     for ext_dir in inputs.ext_build_dirs:
         script += ["##increment_pkg_config_path## $$EXT_BUILD_ROOT$$/" + ext_dir.path]
 


artemyarulin:
Fixed by exporting all the variables to make it accessible for all `make_commands` which user may use for multiple steps

artemyarulin:
Found another issue that this PR resolves - with this PR `CPPFLAGS` are properly forwarded, so it's possible to use dependencies, test case (btw - I can put in examples if you wish) using snappy for rocksdb:

```
cmake_external(name = "snappy", lib_source = "@snappy//:all", static_libraries = ["libsnappy.a"])

make(name = "rocksdb",
     lib_source = "@rocksdb//:all",
     static_libraries = ["librocksdb.a"],
     linkopts = ["-ldl"],
     deps = [":snappy"],
     make_commands = [
         # Solution for https://github.com/bazelbuild/rules_foreign_cc/issues/239
         "export CFLAGS=$(echo $CFLAGS | sed 's/-D__DATE__=.redacted.//')",
         "export CXXFLAGS=$(echo $CXXFLAGS | sed 's/-D__DATE__=.redacted.//')",
         "export ASFLAGS=$(echo $ASFLAGS | sed 's/-D__DATE__=.redacted.//')",
         "export CFLAGS=\"$CFLAGS $CPPFLAGS -DSNAPPY\"",
         "make -j 6 static_lib",
         # Hack - for some reason libs and headers are not included
         "cp librocksdb.a $$INSTALLDIR$$/lib/librocksdb.a", "cp -RL include $$INSTALLDIR$$"
     ],
     make_env_vars = {
         "ROCKSDB_DISABLE_ZLIB":"1",
         "ROCKSDB_DISABLE_BZIP":"1",
         # Solution for https://github.com/bazelbuild/rules_foreign_cc/issues/185
         "AR": "/usr/bin/ar",
     },
)
```

whs-dot-hk:
Hi, @artemyarulin

This may work for rocksdb, but not for most other `Makefile`'s,

because variables are set after `make`,

like `PREFIX` here
https://github.com/bazelbuild/rules_foreign_cc/blob/d54c78ab86b40770ee19f0949db9d74a831ab9f0/tools/build_defs/make.bzl#L49

UebelAndre:
Would the [env](http://localhost/#make-env) or [args](http://localhost/#make-args) attributes also solve for this?

PR #384, opened 2020-03-16
Add option to include share directory as output

mrmeku:
When trying to compile a working binary of `postgres` the binary has a runtime dep of the timezone files it generates within `share/timezonesets`. This option is to take that sort of binary runtime dep into account


jsharpe:
The `out_data_dirs` attribute now addresses this.

PR #386, opened 2020-03-25
Add sha256 to CMake archive

kruton:
In order to have reproducible builds, we want assurance that the CMake archive downloaded
is the one that is intended. Add the expected sha256 value to achieve this goal.

'bazel sync' will no longer complain about a missing hash value.

### DIFF
--- a/for_workspace/repositories.bzl
+++ b/for_workspace/repositories.bzl
@@ -27,6 +27,7 @@ def repositories():
     http_archive(
         name = "cmake",
         build_file_content = _all_content,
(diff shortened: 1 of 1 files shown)


attilaolah:
BTW I proposed the same change in https://github.com/bazelbuild/rules_foreign_cc/pull/381/commits/793a31e6e02e9968b7b909d14b8820ae903f6a3f, as part of https://github.com/bazelbuild/rules_foreign_cc/pull/381.

PR #387, opened 2020-04-09
Add specific shell command support for FreeBSD.

gbrail:
The main change is to use the versions of cp and ln from the coreutils
package, which support all the options that Bazel is expecting -- the
default versions do not.

(Please note that I'm not a BSD expert -- but I have a small project that builds and runs there and this change helped me move forward.)


irengrig:
This is great!
Unfortunately, I can not think of a quick way for me to check it - and I really want this PR to absorb my recent fixes from #377.
Could you possibly have a look whether it works after my changed (see the commit in this PR)?
Thank you!

UebelAndre:
Oof, looks like CI isn't happy about something. @gbrail if you have time to take a look, I'd be happy to merge these changes 😄

tools/build_defs/shell_toolchain/toolchains/impl/bsd_commands.bzl
```diff
@@ -0,0 +1,162 @@
...
+
+def symlink_to_dir(source, target):
+    text = """local target="$2"
+mkdir -p "$target"
+if [[ -f "$1" ]]; then
+  gln -s -t "$target" "$1"
+elif [[ -L "$1" ]]; then
+  gcp $1 $2
+elif [[ -d "$1" ]]; then
+  local children=$(find -H "$1" -maxdepth 1 -mindepth 1)
```
UebelAndre, line 93: There was a recent bugfix for this section of code on other platforms. Does this also suffer from the same issue solved in https://github.com/bazelbuild/rules_foreign_cc/pull/416 ?

PR #393, opened 2020-04-24
Add Apache Thrift example

c4milo:
Update: I got it working using cmake. 

👋 I'm sending this PR to get help building Apache Thrift and also to contribute the example. It currently fails with the following: 

```
...nerate/thrift-t_lua_generator.o src/thrift/generate/thrift-t_netcore_generator.o src/thrift/generate/thrift-t_netstd_generator.o src/thrift/generate/thrift-t_ocaml_generator.o src/thrift/generate/thrift-t_perl_generator.o src/thrift/generate/thrift-t_php_generator.o src/thrift/generate/thrift-t_py_generator.o src/thrift/generate/thrift-t_rb_generator.o src/thrift/generate/thrift-t_rs_generator.o src/thrift/generate/thrift-t_st_generator.o src/thrift/generate/thrift-t_swift_generator.o src/thrift/generate/thrift-t_xml_generator.o src/thrift/generate/thrift-t_xsd_generator.o -ll src/thrift/libparse.a -lpthread
/bin/sh: ../../libtool: No such file or directory
make[3]: *** [thrift] Error 127
make[2]: *** [all-recursive] Error 1
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2
```

Apache Thrift ships with its own libtool wrapper script but for some reason `make` cannot find it. 
cc: @werat


c4milo:
@googlebot I signed it!

fuhailin:
Is there any update about this PR?

c4milo:
Can someone please review this and merge if it makes sense?

jsharpe:
This PR needs updating to the new api changes that have happened since this was first created. Also it duplicates builds of examples which are now more completely done in the examples directory i.e. bison and OpenSSL.

As a general rule we're trying to limit the examples to something that tests a unique use case of the rules, what does thrift show that other existing examples don't? Also these examples should be enabled in the CI test suites (ideally on all platforms including RBE and windows) so they actually get built. Longer term this kind of thing may be better suited to live in the bzlmod registry repository but that project isn't at the point that it'd be ready to accept such contributions yet.

PR #396, opened 2020-05-04
Fix extension-stripping logic

shaoster:
Trivial manual test case:
   
    libpython3.8.a vs libpython3.8m.a

Existing logic erroneously reports collision on "libpython3" token.

Note: string.rpartition is available and semantically unchanged since at least [Bazel 0.23](https://docs.bazel.build/versions/0.21.0/skylark/lib/string.html#rpartition).

Please advise if there's any good places for me to add a unit test for this private function.


shaoster:
(Somebody else with the CLA stuff signed, feel free to take this change.)

shaoster:
@googlebot I signed it!

shaoster:
@oquenchil: Able to take a look at this?

PR #407, opened 2020-06-05
Handle case where LDFLAGS is undefined

errordeveloper:
I am cross-compiling Envoy and for some reason hitting a case where `LDFLAGS` is not set. I don't quite understand the root cause yet, but clearly this code shouldn't assume that `vars`  has `LDFLAG` key set to something, especially because it's set conditionally in the first place:

https://github.com/bazelbuild/rules_foreign_cc/blob/74b146dc87d37baa1919da1e8f7b8aafbd32acd9/tools/build_defs/configure_script.bzl#L144-L147


irengrig:
Thank you for the contribution!

Issue #405, opened 2020-06-04
cmake_external: cp failed: -r not specified; omitting directory 


ukai:
Not sure how to reproduce, but rules_foreign_cc failed cp in symlink_to_dir in tools/build_defs/shell_toolchain/toolchains/impl/linux_commands.bzl
```
cp: -r not specified; omitting directory ...
```

```
def symlink_to_dir(source, target):
    text = """local target="$2"                                                 
mkdir -p "$target"                                                              
if [[ -f "$1" ]]; then                                                          
  ln -s -f -t "$target" "$1"                                                    
elif [[ -L "$1" ]]; then                                           
  cp $1 $2                                                                      
elif [[ -d "$1" ]]; then 
```

if $1 is symlink to dir, it fails.
```
$ mkdir dir
$ ln -s dir symlink-dir
$ mkdir targt
$ if [[ -L symlink-dir ]]; then
  cp symlink-dir target
fi
cp: -r not specified; omitting directory 'symlinkdir'
$ echo $?
1
```

[[ -L "$1" && ! -d "$1" ]] or check [[ -d "$1" ]] first?

Issue #406, opened 2020-06-04
cmake_external: failed to copy file if filename contains space.


ukai:
I found the following error in CMake.log

```
Can not copy /src/workspace/.cache/bazel/_bazel_ukai/a08c2e4811c846650b733c6fc815a920/sandbox/processwrapper-sandbox/1/execroot/test/external/cmake_linux/cmake-3.16.2-Linux-x86_64/doc/cmake/html/_sources/generator/Eclipse
Can not copy CDT4.rst.txt
Can not copy /src/workspace/.cache/bazel/_bazel_ukai/a08c2e4811c846650b733c6fc815a920/sandbox/processwrapper-sandbox/1/execroot/test/external/cmake_linux/cmake-3.16.2-Linux-x86_64/doc/cmake/html/_sources/generator/NMake
Can not copy Makefiles
Can not copy JOM.rst.txt
```

tools/build_defs/shell_toolchain/toolchains/impl/linux_commands.bzl uses
```
  local children=$(find -H "$1" -maxdepth 1 -mindepth 1)
  for child in $children; do
```

but it should be something like
```
 SAVEIFS="$IFS"
 IFS=$'\n'
 local children=($(find -H "$1" -maxdepth 1 -mindepth 1))
 IFS="$SAVEIFS"
 for child in "${children[@]}"; do
```

also, split_argument in tools/build_defs/shell_script_helper.bzl drops quote, so
 cp "$1" "$2"
becomes
 cp $1 $2
in actual script.

```
 parts.append(quoted)
```
should be
```
 parts.append("\"" + quoted + "\"")
```

?

attilaolah:
I just hit my first repo that contains filenamse whit spaces 😢 

Would be awesome to fix this!

kcherneha:
Have the same issue with cmake_external for libblepp library.
What is the current status of this issue? When it will be fixed on master branch?

attilaolah:
I think when #416 is merged this should be fixed.

attilaolah:
#416 was just merged in in 78dd474, this should now be fixed.

PR #416, opened 2020-06-22
support filename containing space

ukai:
handle filename containg space.
cmake prebuilt archive contains such filename in docs etc.

also fix symlink to a directory. [ -L "$1" ] returns success
for it, but it would fail "cp $1 $2" as
 cp: -r not specified; omitting directory ...

Fixes: #405, #406


attilaolah:
@oquenchil any chance this could be cherry-picked? Several folks (including me) are hitting #406 

Looking at the code, this seems to be more complex than I'd like it to be, but I also don't know a better solution.

tools/build_defs/shell_toolchain/toolchains/access.bzl
```diff
@@ -34,8 +34,13 @@ def call_shell(shell_context, method_, *args):
 
     return result
 
+def _quoted(arg):
+    return arg.startswith("\"") and arg.endswith("\"")
```
attilaolah, line 38: Technically, `arg.endswith("\"")` could also be true for non-quoted literals, if we also have `arg.endswith("\\\"")`. But in that case the arg could be half-quoted?

Also I'm not sure how this deals with mixed quotation and whatnot.

tools/build_defs/shell_toolchain/toolchains/access.bzl
```diff
@@ -34,8 +34,13 @@ def call_shell(shell_context, method_, *args):
...
+
 def _wrap_if_needed(arg):
-    return "\"" + arg + "\"" if arg.find(" ") >= 0 else arg
+    if arg.find(" ") >= 0 and not _quoted(arg):
+        return "\"" + arg + "\""
+    return arg
```
attilaolah, line 43: 👍 for making this multi-line.

tools/build_defs/shell_toolchain/toolchains/impl/default_commands.bzl
```diff
@@ -72,8 +72,11 @@ elif [[ -L "$1" ]]; then
   local actual=$(readlink "$1")
   ##symlink_contents_to_dir## "$actual" "$target"
 elif [[ -d "$1" ]]; then
-  local children=$(find -H "$1" -maxdepth 1 -mindepth 1)
-  for child in $children; do
+  SAVEIFS=$IFS
+  IFS=$'\n'
+  local children=($(find -H "$1" -maxdepth 1 -mindepth 1))
+  IFS=$SAVEIFS
```
attilaolah, line 78: First time I see this trick… I looked into it a bit and found [this article describing a similar approach](https://www.cyberciti.biz/tips/handling-filenames-with-spaces-in-bash.html).

Overall 👍.

attilaolah:
OK overall this looks good to me, most of the changes are the `IFS/SAVEIFS` trick for `find`, and the relevant test cases are updated too.

I'd vote for merging this (or cherry-picking if the PR does not get rebased).

oquenchil:
Yes, sorry for the delay, I will review the backlog of rules_foreign_cc PRs tomorrow.

UebelAndre (approved):
I think it's important to note the `IFS` trick. The article linked in an [earlier comment](https://github.com/bazelbuild/rules_foreign_cc/pull/416#discussion_r525102431) seems like it gives good insight to something I think is gonna be unclear to most readers so thanks for finding that!

Other than that this seems great! Thank you!

test/expected/inner_fun_text.txt
```diff
@@ -18,14 +22,18 @@ local target="$2"
...
 ln -s -f -t "$target" "$1"
-elif [[ -L "$1" ]]; then
-cp $1 $2
+elif [[ -L "$1" && ! -d "$1" ]]; then
+cp "$1" "$2"
 elif [[ -d "$1" ]]; then
-local children=$(find -H "$1" -maxdepth 1 -mindepth 1)
+SAVEIFS=$IFS
+IFS=$'
+'
```
djmarcin, line 30: Technically newline is a legal (but highly unusual) file name character. Find supports -print0 to emit null characters which are the only illegal characters in file names (on Unix at least)

UebelAndre: I was thinking of making a similar change to this and yours but was going to use the following which handles spaces just fine.
```bash
find -H "$1" -maxdepth 1 -mindepth 1 | while read child; do
```

Do you have a sense of which approach might be better? Is there an additional change that should be made related to this PR?

djmarcin: It really depends how pedantic you want to be about accepting valid but unusual file names. :-)

Your method covers the vast majority of cases, but it does miss some. It doesn’t properly handle trailing spaces or new lines in the file name. The null separator is something of a de facto Unix standard for handling this — several tools have built in support — but it can still be awkward to work with (see the weird input redirection in #443)

```
test$ find . | while read child; do echo "Child: [$child]"; ls -b "$child" ; done
Child: [.]
file1\ has\ whitespace                file4\ has\ trailing\ spaces\ \
file2\ \ has\ \ double\ \ whitespace  file5\ has\nnewline
\ \ file3\ has\ leading\ spaces       file\ has\ nonprintable\ \177
Child: [./file4 has trailing spaces]
ls: cannot access './file4 has trailing spaces': No such file or directory
Child: [./  file3 has leading spaces]
./\ \ file3\ has\ leading\ spaces
Child: [./file has nonprintable ]
./file\ has\ nonprintable\ \177
Child: [./file5 has]
ls: cannot access './file5 has': No such file or directory
Child: [newline]
ls: cannot access 'newline': No such file or directory
Child: [./file2  has  double  whitespace]
./file2\ \ has\ \ double\ \ whitespace
Child: [./file1 has whitespace]
./file1\ has\ whitespace
```

Issue #320, opened 2019-09-25
Feature request: additional outputs


mattgodbolt:
I'm using foreign rules to build things like `postgres` to run a test instance locally during my build. `postgres` needs files in `$PREFIX/shared` to be preserved. It would be really useful if I could tell `configure_make()` to also preserve more subfolders than:

```
$ ls bazel-out/k8-fastbuild/bin/tools/third_party/postgres
bin/  include/  lib/  logs/
```

Is there something I'm overlooking that allows me to do this already? Thanks in advance!

mattgodbolt:
I've seen that all the files are available in `.../copy_postgres/...` but I'm uncertain how to appropriately refer to these later on (or whether I can rely on such).

irengrig:
Hi @mattgodbolt, right now you can select the copy_postgres directory with the filegroup, specifying the output_group as "gen_dir". 
(and write a custom Starlark rule for selecting the certain files by name/subpath.
I quickly sketched one: https://github.com/bazelbuild/bazel-skylib/pull/194 - probably I will make it a pull request.)

Here is the example how a binary file can be selected: https://github.com/bazelbuild/rules_foreign_cc/blob/14a9f35611c3256496d260ba56c1d602a8deab51/examples/configure_gnuplot/BUILD#L61

Here is the code of creating output groups of cmake_external/configure_make:
https://github.com/bazelbuild/rules_foreign_cc/blob/14a9f35611c3256496d260ba56c1d602a8deab51/tools/build_defs/framework.bzl#L384

- currently the separate file groups are created for the "install" directory, i.e. what is produced by the build, and for each of the specified binary files.
Unfortunately, I have the fixed attributes for each of the directories bin, include etc.

mattgodbolt:
Thanks @irengrig

jschaf:
@mattgodbolt, off-topic, so forgive me in advance:

Would you mind posting your bazel rules for Postgres? This thread is currently the top hit for a Google search of "rules_foreign_cc postgres". I've started to write my own but have hit a couple snags. I'll open a separate bug for those.

mattgodbolt:
Oh! I'm afraid we switched away from Bazel a long time back after hitting too many issues (unlrelated to this). We never got the postgres thing working, sorry.

Issue #241, opened 2019-03-30
Shared library dependencies aren't properly handled


rnburn:
I'm setting up `configure_build` with something like the following:

```
configure_make(
    name = "pcre",
    lib_source = "@org_pcre//:all",
    static_libraries = ["libpcre.a"],
    shared_libraries = ["libpcre.so.1"],
)

configure_make(
    name = "apache_httpd",
    configure_options = [
        "--with-apr=$EXT_BUILD_DEPS/apr",
        "--with-apr-util=$EXT_BUILD_DEPS/apr_util",
        "--with-pcre=$EXT_BUILD_DEPS/pcre",
    ],
    lib_source = "@org_apache_httpd//:all",
    binaries = [
        "httpd",
    ],
    deps = [
        ":apr",
        ":apr_util",
        ":pcre",
    ],
    visibility = [
        "//visibility:public",
    ],
)
```

If I then create a small test script that runs `apache_httpd/bin/httpd -h`

```
filegroup(
    name = "httpd_binary",
    srcs = [
        ":apache_httpd",
    ],

)

sh_test(
    name = "t",
    srcs = [
        "t.sh",
    ],
    data = [
        ":httpd_binary",
        ":apr",
        ":apr_util",
        ":pcre",
    ],
)
```

I get this error when it runs:
```
apache_httpd/bin/httpd: error while loading shared libraries: libpcre.so.1: cannot open shared object file: No such file or directory
```

irengrig:
Hi @rnburn, thank you so much for this issue.
I tried to modify RPATH for the generated binary ([relevant article](https://amir.rachum.com/blog/2016/09/17/shared-libraries/)), but did not succeed, because the configure script also spoils RPATH with the current build directory value.

So please have a look at a not-so-nice workaround in #247 with setting LD_LIBRARY_PATH to contain needed shared libraries. 
Please note the filegroups with output_group attribute - especially for httpd - as this way we are able to select the originally build directory as a whole.
I was thinking of wrapping it into a rule/function.
I would appreciate if you could suggest a nicer solution.

PR #417, opened 2020-06-25
Set LD_LIBRARY_PATH in configure scripts.

bshi:
In attempting to write some rules for PostgreSQL that would build on a
both MacOS and a bare Debian 10 installation with only Bazel and
build-essential installed, I ran into [2] on Linux.  This attempts to
generalize approach taken in [1] to address [2]

This change also adds a foreign_runtime rule which replaces the
filegroup(..., output_group = "gen_dir").  It bundles all transitive
ForeignCcArtifact dependencies and produces a manifest file suitable
to build an LD_LIBRARY_PATH environment variable at runtime.  An
"exec.sh" wrapper then uses the shell runfiles support in @bazel_tools
to build the right environment variable.

Notably, all this was unnecessary on MaxOS.

[1] https://github.com/bazelbuild/rules_foreign_cc/pull/247
[2] https://github.com/bazelbuild/rules_foreign_cc/issues/241

Tested:

Confirm that binaries find the right library dependencies at runtime
on Linux.

    $ bazel run postgres:check_ldd
    ...
    INFO: Build completed successfully, 4 total actions
    	linux-vdso.so.1 (0x00007ffe4a991000)
    	libpq.so.5 => /home/vagrant/.cache/bazel/_bazel_vagrant/1e1e282f2d2054fbbe63d9fcfcb08d64/execroot/rules_foreign_cc_tests/bazel-out/k8-fastbuild/bin/postgres/check_ldd.runfiles/rules_foreign_cc_tests/postgres/copy_postgres/postgres/lib/libpq.so.5 (0x00007f1d67f30000)
    	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f1d67da7000)
    	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f1d67c24000)
    	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f1d67c03000)
    	libreadline.so.8 => /home/vagrant/.cache/bazel/_bazel_vagrant/1e1e282f2d2054fbbe63d9fcfcb08d64/execroot/rules_foreign_cc_tests/bazel-out/k8-fastbuild/bin/postgres/check_ldd.runfiles/rules_foreign_cc_tests/postgres/copy_readline/readline/lib/libreadline.so.8 (0x00007f1d67ba5000)
    	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f1d67b9b000)
    	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f1d679da000)
    	libssl.so.1.1 => /home/vagrant/.cache/bazel/_bazel_vagrant/1e1e282f2d2054fbbe63d9fcfcb08d64/execroot/rules_foreign_cc_tests/bazel-out/k8-fastbuild/bin/postgres/check_ldd.runfiles/rules_foreign_cc_tests/postgres/copy_openssl/openssl/lib/libssl.so.1.1 (0x00007f1d67927000)
    	libcrypto.so.1.1 => /home/vagrant/.cache/bazel/_bazel_vagrant/1e1e282f2d2054fbbe63d9fcfcb08d64/execroot/rules_foreign_cc_tests/bazel-out/k8-fastbuild/bin/postgres/check_ldd.runfiles/rules_foreign_cc_tests/postgres/copy_openssl/openssl/lib/libcrypto.so.1.1 (0x00007f1d675df000)
    	/lib64/ld-linux-x86-64.so.2 (0x00007f1d68062000)
    	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f1d675c3000)
    	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f1d675be000)

---
This change is [Reviewable](https://reviewable.io/reviews/bazelbuild/rules_foreign_cc/417)

### DIFF
--- a/examples/postgres/BUILD
+++ b/examples/postgres/BUILD
@@ -1,0 +1,122 @@
+package(default_visibility = ["//visibility:public"])
+
+load("@rules_foreign_cc//tools/build_defs:configure.bzl", "configure_make")
+load("@rules_foreign_cc//tools/build_defs:framework.bzl", "foreign_runtime")
+
+# This example tests a somewhat complex dependency scheme::
+#
+#     ncurses <- readline <- postgres
+#                 openssl <--/
+
+CONFIGURE_ENV_VARS = select({
+    # See https://github.com/bazelbuild/rules_foreign_cc/issues/338
+    "@bazel_tools//src/conditions:darwin": {"AR": ""},
+    "//conditions:default": {},
+})
+
+sh_library(
+    name = "exec",
+    srcs = ["exec.sh"],
+)
+
+configure_make(
+    name = "ncurses",
+    configure_env_vars = CONFIGURE_ENV_VARS,
+    configure_options = [
+        "--without-manpages",
+        "--without-debug",
+        "--without-ada",
+        "--without-tests",
+        "--without-tack",
+        "--disable-home-terminfo",
+        "--disable-db-install",
+        "--enable-interop",
+    ] + select({
+        "@bazel_tools//src/conditions:darwin": [],
+        "//conditions:default": [
+            # The configure script checks this value and when run under Bazel on
+            # bare-bones Debian 10 it gets a value of "unknown" which triggers
+            # downstream errors.  When run in Debian 10 in a user shell, this is
+            # the result (and it's also a hard coded value in the configure
+            # script).
+            "ARFLAGS=-curvU",
+        ],
+    }),
+    lib_source = "@ncurses//:all",
+    static_libraries = ["libncurses.a"],
+)
+
+configure_make(
+    name = "readline",
+    configure_env_vars = CONFIGURE_ENV_VARS,
+    lib_source = "@readline//:all",
+    static_libraries = [
+        "libreadline.a",
+        "libhistory.a",
+    ],
+    deps = [":ncurses"],
+)
+
+configure_make(
+    name = "openssl",
+    configure_command = "config",
+    configure_env_vars = CONFIGURE_ENV_VARS,
+    lib_source = "@openssl//:all",
+    static_libraries = [
+        "libssl.a",
+        "libcrypto.a",
+    ],
+)
+
+configure_make(
+    name = "postgres",
+    binaries = [
+        "psql",
+        "pg_ctl",
+        "postgres",
+        "initdb",
+    ],
+    configure_env_vars = CONFIGURE_ENV_VARS,
+    configure_options = ["--with-openssl"],
+    lib_source = "@postgres//:all",
+    static_libraries = ["libpq.a"],
+    deps = [
+        ":openssl",
+        ":readline",
+    ],
+)
+
+foreign_runtime(
+    name = "postgres.runtime",
+    srcs = [":postgres"],
+)
+
+runtime_data = select({
+    "@bazel_tools//src/conditions:darwin": [],
+    "//conditions:default": [":postgres.runtime"],
+})
+
+filegroup(
+    name = "psql_binary",
+    srcs = [":postgres"],
+    output_group = "psql",
+)
+
+sh_binary(
+    name = "psql",
+    srcs = [":exec"],
+    args = ["$(location :psql_binary)"],
+    data = [":psql_binary"] + runtime_data,
+    deps = ["@bazel_tools//tools/bash/runfiles"],
+)
+
+sh_binary(
+    name = "check_ldd",
+    srcs = [":exec"],
+    args = [
+        "ldd",
+        "$(location :psql_binary)",
+    ],
+    data = [":psql_binary"] + runtime_data,
+    deps = ["@bazel_tools//tools/bash/runfiles"],
+)
--- a/examples/postgres/exec.sh
+++ b/examples/postgres/exec.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+
+# --- begin runfiles.bash initialization v2 ---
+# See https://github.com/bazelbuild/bazel/blob/master/tools/bash/runfiles/runfiles.bash
+set -uo pipefail; f=bazel_tools/tools/bash/runfiles/runfiles.bash
+source "${RUNFILES_DIR:-/dev/null}/$f" 2>/dev/null || \
+  source "$(grep -sm1 "^$f " "${RUNFILES_MANIFEST_FILE:-/dev/null}" | cut -f2- -d' ')" 2>/dev/null || \
+  source "$0.runfiles/$f" 2>/dev/null || \
+  source "$(grep -sm1 "^$f " "$0.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/null || \
+  source "$(grep -sm1 "^$f " "$0.exe.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/null || \
+  { echo>&2 "ERROR: cannot find $f"; exit 1; }; f=; set -e
+# --- end runfiles.bash initialization v2 ---
+
+set -euo pipefail
+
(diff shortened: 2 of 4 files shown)


bshi:
Hi @oquenchil @irengrig this is an attempt at generalizing some comments that @irengrig made in #241; if the approach looks sound to y'all, I can shore up the documentation and tests.

tools/build_defs/framework.bzl
```diff
@@ -309,6 +313,33 @@ def cc_external_rule_impl(ctx, attrs):
...
+    [0,1,2].  Some [1] configure scripts actually compile test programs to test
+    system configuration and on Linux, without LD_LIBRARY_PATH set these
+    configure scripts fail.  I (bshi) haven't had issues with this on MacOS and
+    purportedly this is only an issue on Linux so ideally we'd constrain to
+    Linux but see [3].
+
+    [0] https://github.com/bazelbuild/rules_foreign_cc/issues/241
+    [1] https://www.postgresql.org/message-id/23645.1437968936%40sss.pgh.pa.us
+    [2] https://github.com/bazelbuild/rules_foreign_cc/pull/247
+    [3] https://github.com/bazelbuild/bazel/issues/11107.
```
UebelAndre, line 329: This seems like something that's now possible

tools/build_defs/framework.bzl
```diff
@@ -674,3 +705,26 @@ def _collect_libs(cc_linking):
             if library:
                 libs.append(library)
     return collections.uniq(libs)
+
+def _foreign_runtime(ctx):
```
UebelAndre, line 709: nit: rename to `_foreign_runtime_impl`

tools/build_defs/framework.bzl
```diff
@@ -674,3 +705,26 @@ def _collect_libs(cc_linking):
...
+    ld_manifest = ctx.actions.declare_file(ctx.attr.name + ".LD_LIBRARY_PATH")
+    ctx.actions.write(
+        output = ld_manifest,
+        content = "\n".join(lib_dirs),
+    )
+    all_deps.append(ld_manifest)
+
+    return [DefaultInfo(runfiles = ctx.runfiles(files = all_deps))]
+
+foreign_runtime = rule(
```
UebelAndre, line 725: Can you add `doc` files here?

UebelAndre (changes requested):
I think in general this seems good but could use a second pair of eyes @jsharpe 🙏 

Other than that, a few comments and can you include `foreign_runtime` in [docs](https://github.com/bazelbuild/rules_foreign_cc/tree/master/docs)?

You can regenerate docs by navigating into that directory and running `bazel run :generate_docs`

examples/examples_repositories.bzl
```diff
@@ -246,3 +246,35 @@ def include_examples_repositories():
...
+        sha256 = "30306e0c76e0f9f1f0de987cf1c82a5c21e1ce6568b9227f7da5b71cbea86c9d",
+    )
+
+    http_archive(
+        name = "openssl",
+        urls = ["https://github.com/openssl/openssl/archive/OpenSSL_1_1_1g.tar.gz"],
+        strip_prefix = "openssl-OpenSSL_1_1_1g",
+        build_file_content = all_content,
+        sha256 = "281e4f13142b53657bd154481e18195b2d477572fdffa8ed1065f73ef5a19777",
+    )
```
jsharpe, line 280: nit: update to 1.1.1i

tools/build_defs/framework.bzl
jsharpe: Yes but linux isn't the only platform that uses `LD_LIBRARY_PATH` e.g. freeBSD so I don't think we should put this comment in as is as its not accurate. On macOS the env var is `DYLD_LIBRARY_PATH` but I don't fully understand when this is required.

bshi: I've updated the language of the comment to clarify that this is a known issue on but potentially not restricted to Linux.

tools/build_defs/framework.bzl
bshi: Done.

examples/examples_repositories.bzl
bshi: Done

bshi:
This should resolve https://github.com/bazelbuild/rules_foreign_cc/issues/320 as well.

examples/postgres/BUILD
```diff
@@ -0,0 +1,122 @@
...
+    "//conditions:default": [":postgres.runtime"],
+})
+
+filegroup(
+    name = "psql_binary",
+    srcs = [":postgres"],
+    output_group = "psql",
+)
+
+sh_binary(
```
UebelAndre, line 105: Is there any way to use something like this in a `sh_test`? I feel like this could potentially get stale since it's not going to get run in CI.

UebelAndre (changes requested):
I think this is looking good. I'll have to take more time to dig into it to get a better feel for the changes but if @jsharpe is good with this then I am too.
 
Can you also add `foreign_runtime` to https://github.com/bazelbuild/rules_foreign_cc/blob/master/docs/docs.bzl#L25 ?

examples/postgres/exec.sh
```diff
@@ -0,0 +1,27 @@
...
+if [ "$( uname )" == "Linux" ]; then
+  export LD_LIBRARY_PATH=
+  for manifest in $( cut -f2 -d' ' "${RUNFILES_MANIFEST_FILE}" | grep -E '.LD_LIBRARY_PATH$' ); do
+    for rel in $( cat "$manifest" ); do
+      LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$PWD/$rel"
+    done
+  done
+fi
+
+exec $@
```
jsharpe, line 27: I feel this script is fairly fundamental to using `foreign_runtime` - shouldn't this be included in `tools/build_defs` rather than in the `examples` `WORKSPACE`.

examples/postgres/BUILD
```diff
@@ -0,0 +1,122 @@
...
+    name = "psql",
+    srcs = [":exec"],
+    args = ["$(location :psql_binary)"],
+    data = [":psql_binary"] + runtime_data,
+    deps = ["@bazel_tools//tools/bash/runfiles"],
+)
```
jsharpe, line 111: This feels like a fair amount of boiler plate to setup an executable. I feel like we should be able to generate these targets automatically from the list of `binaries` given to the framework?

tools/build_defs/framework.bzl
```diff
@@ -324,6 +324,10 @@ def cc_external_rule_impl(ctx, attrs):
         else:
             make_commands.append(line)
 
+    ld_library_path = _define_ld_library_path(attrs.deps)
+    if len(ld_library_path) > 0:
+        define_variables.append("export LD_LIBRARY_PATH=\"{}\"".format(":".join(ld_library_path)))
+
```
jsharpe, line 330: This needs to add a default of appending to LD_LIBRARY_PATH that may already be set via `action_env` or via `cc_env` from the toolchain.

Something like:
```suggestion
    ld_library_path = _define_ld_library_path(attrs.deps)
    if len(ld_library_path) > 0:
        define_variables.append("export LD_LIBRARY_PATH=\"{}:$${LD_LIBRARY_PATH:-}$$\"".format(":".join(ld_library_path)))

```

bshi: Sorry for ghosting out on this!  I'll try to carve out some time this week or next to freshen things up.

PR #422, opened 2020-08-06
Add tools_deps_env_vars

illicitonion:
Some external build tools allow overriding which version of a binary will be used by tool-specific environment variables. This is already done for a hard-coded set of tools here: https://github.com/bazelbuild/rules_foreign_cc/blob/9eb30f8c5a214799b73707666ca49e7b7a35978f/tools/build_defs/configure_script.bzl#L170-L175

This PR allows users to specify that additional environment variables should be set for tools in `${EXT_BUILD_DEPS}/bin` - these are already added to the `$PATH` but some build tools require env vars rather than presence on the `$PATH` to do this overriding.

### DIFF
--- a/tools/build_defs/framework.bzl
+++ b/tools/build_defs/framework.bzl
@@ -105,6 +105,13 @@ CC_EXTERNAL_RULE_ATTRIBUTES = {
         cfg = "host",
         default = [],
     ),
+    "tools_deps_env_vars": attr.string_dict(
+        doc = (
+            "Optional environment variables which will point to the absolute path of the file in " +
+            "tools_deps dir with the value's name. " +
+            "e.g. if this is set as {\"PERL\": \"perl6\"} the action will have the env var PERL=/path/to/tools/deps/bin/perl6"
+        ),
+    ),
     "out_include_dir": attr.string(
         doc = "Optional name of the output subdirectory with the header files, defaults to 'include'.",
         mandatory = False,
@@ -303,7 +310,7 @@ def cc_external_rule_impl(ctx, attrs):
         "export BUILD_TMPDIR=##tmpdir##",
         "export EXT_BUILD_DEPS=##tmpdir##",
         "export INSTALLDIR=$$EXT_BUILD_ROOT$$/" + empty.file.dirname + "/" + lib_name,
-    ]
+    ] + ["export " + key + "=${EXT_BUILD_DEPS}/bin/" + value for key, value in ctx.attr.tools_deps_env_vars.items()]
 
     make_commands = []
     for line in attrs.make_commands:


illicitonion:
@UebelAndre any chance I can nudge you at this? I have a small follow-up to send out which depends on it, too :) Thanks!

tools/build_defs/framework.bzl
```diff
@@ -105,6 +105,13 @@ CC_EXTERNAL_RULE_ATTRIBUTES = {
         cfg = "host",
         default = [],
     ),
+    "tools_deps_env_vars": attr.string_dict(
```
UebelAndre, line 108: What do you think of making this use [label_keyed_string_dict](https://docs.bazel.build/versions/master/skylark/lib/attr.html#label_keyed_string_dict)? This way users can set `{"PERL": "@perl"}` and have the rule hook it up automatically by querying the executable of that target?

UebelAndre (changes requested):
I think this seems fine. I'm wondering if there's a way we can extract this information from a target and not need users to specify a string. I wonder if that's too constraining though. Would love to hear your thoughts 😄

tools/build_defs/framework.bzl
illicitonion: I don't think this can work unless the target is either exactly one file, or contains a file which happens to be named the same as the env var, neither of which is necessarily the case

UebelAndre: Why would the target need to be named the same as the env var? My understanding is you need a way to map an "arbitrary" env var to some binary. Would users always have to specify `${EXT_BUILD_DEPS}/bin` before the name of the executable? My concern with that is it might not be obvious and it's open to users trying to shoehorn this new attr into doing things it's not expected to do. How do you feel about that?

illicitonion: > Why would the target need to be named the same as the env var? My understanding is you need a way to map an "arbitrary" env var to some binary.

I'm saying one or the other would need to be the case - basically, the rule would need to be able to work out which file was intended to be the correct file (two ways of doing that are "there's only one file" or "the file happens to have the same name as the env var", but I don't think there's a general way?).

Though actually `label_keyed_string_dict` goes the wrong way around - that would look like `{"@perl": "PERL"}` which definitely has the cardinality the wrong way around, and AFAIK there's no string_keyed_label_dict.

> Would users always have to specify `${EXT_BUILD_DEPS}/bin` before the name of the executable?

No, the usage would literally be:
```python
tools_deps_env_vars = {
    "PERL": "perl",
}
```

and as long as someone has already added a file named `perl` as a `tools_deps`, `$PERL` will point to an absolute path of the binary named `perl` which is in `tools_deps`.

> My concern with that is it might not be obvious and it's open to users trying to shoehorn this new attr into doing things it's not expected to do. How do you feel about that?

I think it's pretty narrow-scoped and hard to abuse... Can you think of a better docstring to make clear what it's for / how it works?

UebelAndre: I think this is fine then. I still think I'd prefer something like `label_keyed_string_dict` because it maps a string to a target and Bazel will warn users when that target is not configured correctly, but there's nothing to warn about string to string mappings. Users will be stuck having to debug the make build to find out why their strings don't match up or will never be told that one doesn't make sense anymore (eg: `{"MY_BIN": "bin_not_being_provided"}`). But I suppose this is already an existing patter and something users are expected to deal with.

UebelAndre: I think the current changes are good and by necessity. It could be the case that a tool provides multiple binaries you want to create mappings for. so LGTM

UebelAndre (approved):

tools/build_defs/framework.bzl
jsharpe: Is it not better off simply having a env attr that sets the environment globally and not specific to tools_deps? So long as we pass it through `ctx.expand_location` then you can do 
```
env = {
    "PERL": $(location :perl)
    }
```
which I think then sidesteps various issues that the copying introduces in terms of ensuring run files etc are all copied and setup correctly (which they currently aren't).  This would also address #383 too in a general way?

illicitonion: That would be fine by me... For my use case I need the binaries in both `$PATH` and specific env vars, so conflating the two made sense for me, but I can tease them apart if you'd prefer. Though I guess it would mean that `$PATH` and the other env vars would be pointing at distinct-but-identical binaries, which may be problematic in some cases...

illicitonion:
This was obsoleted by other features.

PR #423, opened 2020-08-06
Expand depset to list

illicitonion:
Otherwise if I put anything in additional_tools I get an error about
depset not being iterable.

This matches the for loop just above, which does the to_list conversion.


UebelAndre (approved):

UebelAndre:
@illicitonion Thanks! 🙏

PR #424, opened 2020-08-25
Fix failing configure_make builds on macOS

crncnnr:
Hey thanks for this project, it's amazing.

I recently had `configure_make` builds fail on macOS, but the logs showed the compile was successful. I tracked the bug down to the generated shell script, which was throwing an error on copying the results.

There was a slight error with the bash syntax, a local variable was being created *and* expanded at the same time, but this fails as the variable doesn't yet exist.

Removing this variable expansion fixed my builds.

Something like `set -o nounset` or `set -e` would've caught this, but I wasn't confident enough with my bash skills to add that here, out of fear of causing errors elsewhere.

### DIFF
--- a/tools/build_defs/shell_toolchain/toolchains/impl/osx_commands.bzl
+++ b/tools/build_defs/shell_toolchain/toolchains/impl/osx_commands.bzl
@@ -74,7 +74,7 @@ for child in "${children[@]}"; do
   if [[ -f "$child" ]]; then
     cp "$child" "$target"
   elif [[ -L "$child" ]]; then
-    local $actual=$(readlink "$child")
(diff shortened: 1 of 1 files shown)


UebelAndre:
continuing from https://github.com/bazelbuild/rules_foreign_cc/pull/487#issuecomment-772750912

Just a simple test, if you can add a test that fails before your change and pass after it, that'd be all that's necessary.

UebelAndre:
It seems like the changes here are tested in https://github.com/bazelbuild/rules_foreign_cc/pull/500#issuecomment-776327492 so I think it's now good to merge these changes. Thanks for opening this! Hopefully we'll be able to be more responsive to more PRs 😄

PR #428, opened 2020-09-17
Replace target attribute with targets

attilaolah:
This is so that multiple labels could be provided to a toolchain as sources. An example is `emcmake`, which needs both `emscripten` and `binaryen` to run.

Another reason for this is to pass in config files that might be needed for the toolchain. The config files can be passed in as a label and will be available in `${EXT_BUILD_DEPS}/bin`, just like any other directory.

### DIFF
--- a/examples/with_prebuilt_ninja_artefact/BUILD
+++ b/examples/with_prebuilt_ninja_artefact/BUILD
@@ -9,7 +9,7 @@ native_tool_toolchain(
     # In the path, we also start with the name of the directory,
     # in this case the external repository.
     path = "ninja_artefact/ninja",
-    target = "@ninja_artefact//:all",
+    targets = ["@ninja_artefact//:all"],
     visibility = ["//visibility:public"],
 )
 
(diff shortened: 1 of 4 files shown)


attilaolah:
Actually, let's hold off with this: it should be possible to use a file group or something similar to get this to work.

jsharpe:
> Actually, let's hold off with this: it should be possible to use a file group or something similar to get this to work.

Yes that's what I was also thinking; the other way is a `sh_binary` wrapper script which sets up the `PATH` env before calling out to the target to reference the tools stored in a `data` attribute on the `sh_binary`?

Also see the changes I'm making in #437 - enabling use of `$(execpath)` in the `path` attribute - does that help?

attilaolah:
Yes, let me revert this back to a draft PR while I try to plumb something together on my side.

attilaolah:
I'm trying to get this to work, but I think it won't be possible until #437 is merged.

- If I pass in a `filegroup` as the `target` label, none of the sources of the filegroup are even built.
- If I pass in a `sh_binary` rule, it seems to work but then I lose all the runfiles (which are built for the host/exec environment, but not added to the `.runfiles` dir. I think #437 should fix this.

attilaolah:
I finally managed to work around this, but I basically had to inject all required tools into `tools_deps`, which is annoying a bit but I have macros for each rule anyway.

It's not great but should work until #437 is done, and so I don't need this change any more.

PR #429, opened 2020-09-30
Add prefix script attribute

whs-dot-hk:
This pr add `prefix_script` to attribute, unlike `postfix_script`, `prefix_script` will run before the configure command, which open up new opposites, like making changes to the makefiles or accepting a license.

This new attribute will works for `configure_make`, `cmake_external`, etc.

I also provided an example of building ffmpeg.

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,6 +270,9 @@ attrs: {
     # Not used by the shell script part in cc_external_rule_impl.
     "additional_tools": attr.label_list(mandatory = False, allow_files = True, default = []),
     #
+    # Optional part of the shell script to be added before the configure command
+    "prefix_script": attr.string(mandatory = False),
+    #
     # Optional part of the shell script to be added after the make commands
     "postfix_script": attr.string(mandatory = False),
     # Optinal make commands, defaults to ["make", "make install"]
--- a/examples/build_ffmpeg_with_prefix_script/BUILD.ffmpeg
+++ b/examples/build_ffmpeg_with_prefix_script/BUILD.ffmpeg
@@ -1,0 +1,31 @@
+load("@rules_foreign_cc//tools/build_defs:configure.bzl", "configure_make")
+
+filegroup(
+    name = "all",
+    srcs = glob(["**"]),
+)
+
+configure_make(
+    name = "library",
+    lib_source = ":all",
+    configure_options = [
+        "--disable-static",
+        "--enable-shared",
+        "--enable-libx264",
+        "--enable-gpl",
+    ],
+    deps = [
+        "@x264//:library",
+    ],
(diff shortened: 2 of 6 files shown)


whs-dot-hk:
@googlebot I signed it!

jsharpe:
Is this actually needed? Can't this always be done either by a `genrule` to create the input to `lib_sources` or by a patch on the repository rule?

whs-dot-hk:
No, a `genrule` cannot put the files in places...

But, I have second through on capability, so I created `new_configure_make`
https://github.com/whs-dot-hk/whs_rules_foreign_cc/blob/9a507e13c7973e6f77e14a1aaa61164de1c02638/tools/build_defs/configure.bzl#L168

UebelAndre:
>  and patch doesn't work for `new_git_repository`...

It doesn't? It has documentation for it ([new_git_repository::patches](https://docs.bazel.build/versions/master/repo/git.html#new_git_repository-patches)). What about it doesn't work?

> or accepting a license.

Can you elaborate on this a bit?

whs-dot-hk:
@UebelAndre For git repositories, if I stated `master` it always updates, so patches will not work

Currently, we cannot run commands before `configure`, `autoreconf`, etc, so lacks the flexibility to ready the code before building

UebelAndre:
>  For git repositories, if I stated `master` it always updates, so patches will not work

You can set `patch_tool = "bash"` and have the `patches` attribute be a list of bash scripts to mutate the repository you're cloning. I think this would give you the same flexibility as a `prefix_script`. Would that help?

matzipan:
There's a use-case for this: 

Feeding a pkgconfig library like openssl to a CMake library. Your pkgconfig willl look like this:

```
prefix=${EXT_BUILD_DEPS}/openssl
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: OpenSSL
Description: Secure Sockets Layer and cryptography libraries and tools
Version: 1.1.1b
Requires: libssl libcrypto
```

But because you can't call `replace_in_files` before the CMake command, CMake ingests the `${EXT_BUILD_DEPS}`, tries to expand it and fails.

Trying to figure out if this can be achieved by any other means, but I'm not sure...

jsharpe:
Related to the above is https://gitlab.kitware.com/cmake/cmake/-/issues/19254. 
The way I solve this is to have a wrapper script for pkgconfig that sets define-prefix to replace the EXT_BUILD_DEPS variable.
The real fix for this though is to sort out a proper toolchain configuration for pkg-config so that the wrapper script can be included as part of these rules to avoid the need for such hacks.

PR #432, opened 2020-10-21
configure: add autoconf to configure

otan:
We have run into a library that works with autoconf (jemalloc) but not autoreconf.
As such, I've introduced a new rule to use autoconf as the build script.


irfansharif:
Bump, @oquenchil. We've been using this in cockroachdb/cockroach for bit now and "it works".

UebelAndre (approved):
The changes here look good to me! Can you rebase and update the documentation?

otan (commented):
done!

UebelAndre:
Thanks! Much appreciated 😄

PR #437, opened 2020-11-04
Resolve tools to include runfiles

### LINKED ISSUE #433: The tools_deps targets don't include runfiles
Where a tools_deps is a build target that includes runfiles the run files are not made available for the tool to use and so the tool fails to run.

jsharpe:
Fixes #433


jsharpe:
@googlebot I signed it!

jsharpe:
@oquenchil is there anything you'd like me to change to get this merged?

tools/build_defs/framework.bzl
```diff
@@ -276,14 +276,17 @@ def cc_external_rule_impl(ctx, attrs):
     if "requires-network" in ctx.attr.tags:
         execution_requirements = {"requires-network": ""}
 
+    inputs_from_tools, manifests_from_tools = ctx.resolve_tools(tools = ctx.attr.tools_deps)
```
jsharpe, line 279:
```suggestion
    inputs_from_tools, manifests_from_tools = ctx.resolve_tools(tools = attrs.tools_deps)
```

tools/build_defs/framework.bzl
```diff
@@ -73,6 +73,7 @@ CC_EXTERNAL_RULE_ATTRIBUTES = {
         ),
         mandatory = False,
         allow_files = True,
+        cfg = "host",
```
UebelAndre, line 76: you probably want `cfg = "exec"` 😄

jsharpe: Do you know what version that was introduced? I was trying to find it but failed!

UebelAndre: Hmm, let me do some digging. https://docs.bazel.build/versions/master/skylark/rules.html#configurations is where I learned about this (just for documentations sake)

UebelAndre: And it appears to have first appeared in the [3.6.0](https://docs.bazel.build/versions/3.6.0/skylark/rules.html) documentation.

tools/build_defs/native_tools/BUILD
```diff
@@ -18,7 +18,7 @@ make_tool(
 
 native_tool_toolchain(
     name = "built_make",
-    path = "make/bin/make",
+    path = "$(execpath :make_tool)/bin/make",
```
UebelAndre, line 21: seems a bit odd. I'd expect `$(execpath :make_tool)` to be a path to the tool and `target = ":make_tool"` should be sufficient. Do you think it'd be difficult to update this to not require both `target` and `path`?

tools/build_defs/cmake_script.bzl
```diff
@@ -73,6 +77,10 @@ def create_cmake_script(
     if not params.cache.get("CMAKE_RANLIB"):
         params.cache.update({"CMAKE_RANLIB": ""})
 
+    if len(make_path.split('/')) != 1:
```
UebelAndre, line 80: what is this accounting for?

UebelAndre (changes requested):
A couple of questions/comments

tools/build_defs/cmake_script.bzl
jsharpe: This is for the case that its using a builtin tool that isn't linked into the sandbox e.g. the system installed `cmake`

tools/build_defs/native_tools/BUILD
jsharpe: `$(execpath :make_tool)` is the output from a `rules_foreign_cc` build which doesn't generate an executable target directly ( you have to do the dance of creating a `filegroup` and an executable target via a wrapper or another `cc_*` rule. This issue is along the lines of the same problem that #428 is looking to address too; there may be multiple tools required in the `PATH` but I guess in the emscripten case that this could be done by a shell wrapper that sets up the path so that `binaryen` is present in the context that `emscripten` is expecting it to be.

Also note that I'd made this observation in the `tool_access.bzl` file that it'd be easier if this was an executable target than a freeform path as we need to special case between absolute and relative paths.

tools/build_defs/framework.bzl
```diff
@@ -518,25 +525,15 @@ def _correct_path_variable(env):
...
     lines = []
     lines += _symlink_contents_to_dir("lib", files.libs)
     lines += _symlink_contents_to_dir("include", files.headers + files.include_dirs)
 
     if files.tools_files:
         lines.append("##mkdirs## $$EXT_BUILD_DEPS$$/bin")
-    for tool in files.tools_files:
-        lines.append("##symlink_to_dir## $$EXT_BUILD_ROOT$$/{} $$EXT_BUILD_DEPS$$/bin/".format(tool))
+    for tool in files.tools_files.to_list():
+        lines.append("##symlink_to_dir## $$EXT_BUILD_ROOT$$/{} $$EXT_BUILD_DEPS$$/".format(_file_path(tool)))
```
jsharpe, line 536: I actually think that by using `$(execpath)` in places that this code can be removed. I need to add calls to `ctx.expand_location` for the command strings so they can use the tools provided here directly. 
I also think that introducing a global `env` attribute, as we've discussed in other PRS, and then allowing Make variable substitution of these variables into the commands would be a good idea.

tools/build_defs/framework.bzl
```diff
@@ -680,37 +677,25 @@ def _define_inputs(attrs):
...
-        for file_list in tool.files.to_list():
-            tools_files += _list(file_list)
-
     for input in attrs.additional_inputs:
         for file_list in input.files.to_list():
-            input_files += _list(file_list)
+            input_files += list(file_list)
 
-    # These variables are needed for correct C/C++ providers constraction,
+    tools_files, manifests_from_tools = ctx.resolve_tools(tools = ctx.attr.tools_deps + attrs.additional_tools)
```
jsharpe, line 685: I'm not sure what effect not propagating `manifests_from_tools` here has - its marked as an experimental parameter on the API at the moment in the bazel docs.

UebelAndre, line 685: is `manifests_from_tools` intentionally unused here?

UebelAndre (changes requested):
One last question but I think it otherwise looks good.

tools/build_defs/framework.bzl
jsharpe: It is at the moment but probably should be plumbed through InputFiles to the point at which a `ctx.actions` call using `tools_files` is made and passed in there. I don't actually know what the `manifests_from_tools` argument does - its marked as experimental on the API in the bazel docs.

UebelAndre: Well, can you prefix it with `_` if it's not being used and add a comment for some of this context?

jsharpe: I've now updated the code to propagate it to the `run_shell` call.

tools/build_defs/configure.bzl
```diff
@@ -69,7 +69,7 @@ def _create_configure_script(configureParameters):
         autoreconf_options = ctx.attr.autoreconf_options,
         autoreconf_env_vars = ctx.attr.autoreconf_env_vars,
         autogen = ctx.attr.autogen,
-        autogen_command = ctx.attr.autogen_command,
+        autogen_command = ctx.expand_location(ctx.attr.autogen_command),
```
jsharpe, line 72: Probably also needs `targets=ctx.attr.tools_deps+ctx.attr.additional_tools` as above

UebelAndre:
Is https://github.com/bazelbuild/rules_foreign_cc/issues/490 related to this in any way?

jsharpe:
> Is #490 related to this in any way?

Not directly; but it'd be natural to fix up #490 after landing this.

whs-dot-hk:
Good job, this is it... It will put an end to this repo

UebelAndre:
> Good job, this is it... It will put an end to this repo

Can you explain why you feel that way?

PR #441, opened 2020-11-24
symlink_to_dir: reordered if statement

filippobrizzi:
## Context

To compile `opencv` with `opencv_contrib` I put `opencv_contrib` as `tools_deps` for `opencv`. 
I switched to the latest master and got this error:
`cp: omitting directory opencv_contrib`
Basically, Bazel was failing to copy the `opencv_contrb` files in `${EXT_BUILD_DEPS}/bin/com_github_opencv_opencv_contrib`.

## Debug
I started debugging and I found that my issue was introduced in https://github.com/bazelbuild/rules_foreign_cc/pull/377.
I run some test and found out that re-ordering the `if` statements solve my issue.

However, now CI is failing and I am not sure why.

Can anyone help here?
Thanks!


filippobrizzi:
Hi @irengrig could you have a look at this PR? Thank you!

filippobrizzi:
@googlebot I signed it!

tools/build_defs/shell_toolchain/toolchains/impl/osx_commands.bzl
```diff
@@ -106,17 +106,13 @@ fi
 def symlink_to_dir(source, target):
     text = """local target="$2"
 mkdir -p "$target"
-if [[ -f "$1" ]]; then
+if [[ -d "$1" ]]; then
```
filippobrizzi, line 109: I needed to revert to previous otherwise it doesn't work for me

UebelAndre:
> Can anyone help here?

Hi! Sorry for the delay. The failure you're getting is
```output
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
Executing tests from @rules_foreign_cc//test:shell_method_symlink_contents_to_dir_test
-----------------------------------------------------------------------------
2,5d1
< 
< aaa/include:
< header1.h
< header2.h
FAIL: files "external/rules_foreign_cc/test/expected/out_symlinked_dirs_osx.txt" and "external/rules_foreign_cc/test/out_symlinked_dirs.txt" differ
```
So you'll have to also update the `@rules_foreign_cc//test:shell_method_symlink_contents_to_dir_test` and `@rules_foreign_cc//test:shell_script_inner_fun_test` targets as part of this change, it would seem.

filippobrizzi:
@UebelAndre thanks a lot for the hint, I managed to fix tests!

test/expected/out_symlinked_dirs_osx.txt
```diff
@@ -1,5 +1 @@
 include
-
-aaa/include:
```
UebelAndre, line 3: Does this mean these headers aren't made being made available after this change? Can you explain why this changed?

UebelAndre (changes requested):
On question about the changes and, would it not be clearer to pass `opecv_contrib` as `additional_input`? Why is it considered a tool?

filippobrizzi:
> On question about the changes and, would it not be clearer to pass `opecv_contrib` as `additional_input`? Why is it considered a tool?

@UebelAndre I need the `opencv_contrib` files to be copied somewhere where `opencv` can access them as `opencv` is going to compile them directly. From my understanding and tests `tools_deps` is the only parameter that allows for this, but if it's not true I'll be happy to change.

UebelAndre:
> @UebelAndre I need the `opencv_contrib` files to be copied somewhere where `opencv` can access them as `opencv` is going to compile them directly. From my understanding and tests `tools_deps` is the only parameter that allows for this, but if it's not true I'll be happy to change.

Can you link a BUILD and WORKSPACE file that builds `opecv` so I can do some testing?

filippobrizzi:
Hi @UebelAndre, here is the WORKSPACe https://gist.github.com/filippobrizzi/0579beec46ad5174eba9bf57b70d7913
and the BUILD https://gist.github.com/filippobrizzi/227eada1bb09006a985e3cbdae566887 files to compile OpenCV with opencv_contrib

tools/build_defs/shell_toolchain/toolchains/impl/osx_commands.bzl
UebelAndre: Sorry, I don't think I understand what you're referring to here. Can you elaborate a bit?

UebelAndre:
> Hi @UebelAndre, here is the WORKSPACe https://gist.github.com/filippobrizzi/0579beec46ad5174eba9bf57b70d7913
> and the BUILD https://gist.github.com/filippobrizzi/227eada1bb09006a985e3cbdae566887 files to compile OpenCV with opencv_contrib

Sorry, I still don't think I understand the impacts of the changes here. When I build locally using your WORKSPACE and BUILD file as-is, it works for me on MacOS (haven't tried other systems). Can you link the failure you were getting?

filippobrizzi:
@UebelAndre I found out something very interesting, building with the default spawn strategy: `--spawn_strategy=sandbox` the compilation works fine.
But building with `--spawn_strategy=local` gives the following error:
```
cp: /private/var/tmp/_bazel_fbrizzi/16d3af4a7aff65af89bd2fe5a20e806d/execroot/opencv_bazel_test/external/com_github_opencv_opencv_contrib is a directory (not copied).

```

UebelAndre:
> @UebelAndre I found out something very interesting, building with the default spawn strategy: `--spawn_strategy=sandbox` the compilation works fine.
> But building with `--spawn_strategy=local` gives the following error:
> 
> ```
> cp: /private/var/tmp/_bazel_fbrizzi/16d3af4a7aff65af89bd2fe5a20e806d/execroot/opencv_bazel_test/external/com_github_opencv_opencv_contrib is a directory (not copied).
> ```

Hmmm, very interesting. I'll have to dig into that. I know we don't build in the Bazel sandbox so maybe the behavior here is related to that? https://github.com/bazelbuild/rules_foreign_cc/issues/478

PR #443, opened 2020-12-11
Use null delimited find for looping

djmarcin:
Looping over the results of find does not work if the filenames contain spaces.


UebelAndre:
Hey! Thanks for opening a PR for this. Can you take a look at https://github.com/bazelbuild/rules_foreign_cc/pull/416 real quick and let me know if this also solves the same issue for you?

PR #456, opened 2021-01-16
toolchain_mappings: remove unnecessary arch constraints

jesec:
This change allows rules_foreign_cc to work on architectures other
than x86_64.

Bug: bazelbuild/bazel#11628
Test: manual. no longer fails with "ln: illegal option -- t" on darwin_arm64


jesec:
@googlebot I signed it!

whs-dot-hk:
Look very legit to me the extra `x86_64` condition was add during #196

jsharpe (approved):

PR #458, opened 2021-01-22
Enable Buildifier in CI

UebelAndre:
The purpose of this PR is to enable Buildifier in CI. This is done by mostly ignoring existing defects but still allows for future changes to benefit from this check.

New changes should not "disable" any defects.

### DIFF
--- a/examples/cmake_android/BUILD
+++ b/examples/cmake_android/BUILD
@@ -1,3 +1,4 @@
+load("@rules_cc//cc:defs.bzl", "cc_library")
 load("@rules_android//android:rules.bzl", "android_binary", "android_library")
 load("@rules_foreign_cc//tools/build_defs:cmake.bzl", "cmake_external")
 
--- a/.bazelci/config.yaml
+++ b/.bazelci/config.yaml
@@ -124,3 +124,5 @@ tasks:
       - "//cmake_android:app"
     test_targets:
       - "//:tests_no_synthetic"
+
+buildifier: latest
--- a/cc_configure_make/def.bzl
+++ b/cc_configure_make/def.bzl
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# buildifier: disable=module-docstring
 _DEPRECATION_NOTICE = """\
 `cc_configure_make` is deprecated and will soon be removed. Please use \
 configure_make rule defined in `@rules_foreign_cc//tools/build_defs:configure.bzl` \
@@ -77,7 +78,9 @@ _cc_configure_make_rule = rule(
     deprecation = _DEPRECATION_NOTICE,
 )
 
+# buildifier: disable=function-docstring
 def cc_configure_make(name, configure_flags, src, out_lib_path):
+    # buildifier: disable=print
     print("WARNING: " + _DEPRECATION_NOTICE)
     name_cmr = "_{}_cc_configure_make_rule".format(name)
     _cc_configure_make_rule(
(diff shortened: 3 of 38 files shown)


jsharpe (approved):

attilaolah:
It might be neater to do this in two steps:

1. First just run `buildifier` to fix formatting / whitespace-related issues.
2. Then run with `-lint=fix` to do the more substantial changes, like replacing `navice` rules with the Starlark ones from `@rules_cc`.

Note also that CI is failing in `'//cmake_android:app'`.

examples/cmake_android/BUILD
```diff
@@ -1,3 +1,5 @@
+load("@rules_cc//cc:defs.bzl", "cc_library")
+load("@rules_android//android:rules.bzl", "android_binary", "android_library")
```
attilaolah, line 2: This seems to be [failing on BuildKite](https://buildkite.com/bazel/rules-foreign-cc/builds/1282#0428871d-995f-45d2-8b16-662c8c404077):

```
(14:10:54) ERROR: error loading package 'cmake_android': Unable to find package for @rules_android//android:rules.bzl: The repository '@rules_android' could not be resolved.
```

jsharpe:
I'm actually not sure that depending on `rules_cc` is the right thing to do yet as the transition path to starlark cc rules isn't yet well defined. We've actually done what you suggested already and merged a PR to fix up just the formatting without lint fixes.

UebelAndre:
Yeah, this has been mostly solved for in https://github.com/bazelbuild/rules_foreign_cc/pull/459 but I want to see what else we can do here so users can freely run buildifier on this repo and not run into any issues.

UebelAndre:
Ok, this is in much better shape. I'll split out some PRs from this but feel free to take a look here and let me know what you think. @attilaolah @jsharpe 

> I'm actually not sure that depending on `rules_cc` is the right thing to do yet as the transition path to starlark cc rules isn't yet well defined.

I think the intent that the `rules_cc` repo be used is pretty clear. Most (if not all) the symbols in that repo are wrappers around the native call so this shouldn't be a big change. Additionally, *all* uses of `rules_cc` are in examples so there's very little risk to introducing it. Do you concur?

examples/cmake_android/BUILD
UebelAndre: fixed

jsharpe:
> Ok, this is in much better shape. I'll split out some PRs from this but feel free to take a look here and let me know what you think. @attilaolah @jsharpe
> 
> > I'm actually not sure that depending on `rules_cc` is the right thing to do yet as the transition path to starlark cc rules isn't yet well defined.
> 
> I think the intent that the `rules_cc` repo be used is pretty clear. Most (if not all) the symbols in that repo are wrappers around the native call so this shouldn't be a big change. Additionally, _all_ uses of `rules_cc` are in examples so there's very little risk to introducing it. Do you concur?

That was the original intent for `rules_cc` but the original migration project failed. See https://docs.google.com/document/d/1GW7UVo1s9X0cti9OMgT3ga5ozKYUWLPk9k8c4-34rC4/edit# for the current thinking on how the migration should proceed. Also see issue https://github.com/bazelbuild/rules_cc/issues/92.

However as you say with all uses being in the examples repo then this doesn't introduce an extra dependency on the end user and so if it gets us to the goal of being able to run Buildifier in lint mode as a CI check then I'm for it. 👍

UebelAndre:
Huzzah! This PR is now ready 😄

jsharpe (changes requested):
Can we put the docs update into a separate PR?

UebelAndre:
> Can we put the docs update into a separate PR?

Yeah: https://github.com/bazelbuild/rules_foreign_cc/pull/473

jsharpe (approved):

PR #461, opened 2021-01-22
Added `data` attr to common framework

UebelAndre:
This adds a [`data`](https://docs.bazel.build/versions/master/be/common-definitions.html#typical-attributes) attribute to rules rules built on `framework.bzl` (`boost_build`, `cmake_external`, `configure_make`, and `make`).

Examples have been added to `@rules_foreign_cc_tests//cmake_with_data/...`

Note, the `@rules_foreign_cc_tests//cmake_with_data:test_with_data` examples will not pass without [--enable_runfiles](https://docs.bazel.build/versions/master/command-line-reference.html#flag--enable_runfiles) enabled. This will need to be explicitly set on windows machines

### DIFF
--- a/tools/build_defs/framework.bzl
+++ b/tools/build_defs/framework.bzl
@@ -386,7 +386,7 @@ def cc_external_rule_impl(ctx, attrs):
     return [
         DefaultInfo(
             files = depset(direct = rule_outputs + wrapped_files),
-            runfiles = ctx.runfiles(attrs.data),
+            runfiles = ctx.runfiles(ctx.files.data),
         ),
         OutputGroupInfo(**output_groups),
         ForeignCcDeps(artifacts = depset(
--- a/tools/build_defs/framework.bzl
+++ b/tools/build_defs/framework.bzl
@@ -370,6 +370,12 @@ def cc_external_rule_impl(ctx, attrs):
         env = cc_env,
     )
 
+    # Gather runfiles transitively as per the documentation in:
+    # https://docs.bazel.build/versions/master/skylark/rules.html#runfiles
+    runfiles = ctx.runfiles(files = ctx.files.data)
+    for target in [ctx.attr.lib_source] + ctx.attr.additional_inputs + ctx.attr.deps + ctx.attr.data:
+        runfiles = runfiles.merge(target[DefaultInfo].default_runfiles)
+
     externally_built = ForeignCcArtifact(
         gen_dir = installdir_copy.file,
         bin_dir_name = attrs.out_bin_dir,
@@ -386,7 +392,7 @@ def cc_external_rule_impl(ctx, attrs):
     return [
         DefaultInfo(
             files = depset(direct = rule_outputs + wrapped_files),
-            runfiles = ctx.runfiles(ctx.files.data),
+            runfiles = runfiles,
         ),
         OutputGroupInfo(**output_groups),
         ForeignCcDeps(artifacts = depset(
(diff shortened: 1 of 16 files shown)


UebelAndre:
@jsharpe do you have a moment to look at this? 🙏

tools/build_defs/framework.bzl
```diff
@@ -378,7 +384,10 @@ def cc_external_rule_impl(ctx, attrs):
     ]
     output_groups[attrs.configure_name + "_logs"] = wrapped_files
     return [
-        DefaultInfo(files = depset(direct = rule_outputs + wrapped_files)),
+        DefaultInfo(
+            files = depset(direct = rule_outputs + wrapped_files),
+            runfiles = ctx.runfiles(ctx.files.data),
```
jsharpe, line 389: I think you also need to grab the transitive run files from ctx.files.data and ctx.files.deps as per the example at https://docs.bazel.build/versions/master/skylark/rules.html#runfiles?

jsharpe (changes requested):
Mostly LGTM. Just the one question about if the runfiles collection is correct - I think we need to transitively merge the `runfiles` from both `deps` and `data` to ensure we get all of the required runfiles.

tools/build_defs/framework.bzl
UebelAndre: Good call, updated.

UebelAndre: I've also added the default runfiles of `ctx.attr.lib_source` to the set being returned in the `DefaultInfo` provider here. I think that's correct as well(?)

jsharpe: Yes I think that is correct; do we also need `ctx.attr.additional_inputs` run files too?

UebelAndre: Another good call, updated.

jsharpe (approved):
LGTM

whs-dot-hk:
I think this is quite useless, without the control of where the data should be placed

UebelAndre:
> I think this is quite useless, without the control of where the data should be placed

@whs-dot-hk The motivation here was so the data would be available to downstream targets however Bazel would traditionally make it available. The example added in this PR is the use case I have for it and I think it's valuable. But maybe I'm missing something? Can you elaborate a bit more?

whs-dot-hk:
Because, you need to modify the code to use `data`, and with cmake, configure make, etc, you don't mess with the source code most of the time

UebelAndre:
> Because, you need to modify the code to use `data`, and with cmake, configure make, etc, you don't mess with the source code most of the time

Ah, I think that's a good call out. There have been discussions about having something like an `env` attribute for defining arbitrary values that map to arbitrary locations (similar discussion on https://github.com/bazelbuild/rules_foreign_cc/pull/422) but that wasn't really my intent for this PR. I'm not using the files I pass to `data` in my build. They're intended for use at runtime. I'm, in particular, using this attribute to pass [cc_import](https://docs.bazel.build/versions/master/be/c-cpp.html#cc_import) targets to be used at runtime which has worked out perfectly because, again, this data is not needed until runtime where I can use existing Bazel patterns to locate the file in either a rule or with the understanding of where Bazel would put otherwise put runfiles. I do not think the existence of this attribute requires anyone to modify external source code they're trying to build with this project. Can you provide a more concrete example where that might be required? I'm happy to try and fix any issues I introduce or work to figure out what the missing features are.

whs-dot-hk:
What about something like a `Dockerfile`'s `COPY`?

UebelAndre:
> What about something like a `Dockerfile`'s `COPY`?

I don't think I understand what you mean by that. Can I ask you to elaborate in a new issue? 

It sounds to me that you feel there's a lack of flexibility here. Am I correct in saying that?

PR #469, opened 2021-01-25
Add tests for checking the minimum supported version to CI

UebelAndre:
This adds new builds to CI to test the root of the repo as well as the examples. Among those new builds are two checks for the minimum supported Bazel version.

In the future it might be better to consolidate all these tests into smaller tests cases such that we only have jobs for testing the root, examples, and docs directories. But I think making sure everything is being tested is a better place to be in.

### DIFF
--- a/.bazelci/config.yaml
+++ b/.bazelci/config.yaml
@@ -55,13 +55,13 @@ tasks:
       - "//:win_tests"
   ubuntu2004_flags:
     name: Flags
-    platform: ubuntu1804
+    platform: ubuntu2004
     working_directory: test/standard_cxx_flags_test
     test_targets:
       - "//:flags_test"
   ubuntu1804_flags:
     name: Flags
-    platform: ubuntu1604
+    platform: ubuntu1804
     working_directory: test/standard_cxx_flags_test
     test_targets:
       - "//:flags_test"
@@ -79,13 +79,13 @@ tasks:
       - "//:flags_test"
   ubuntu2004_detect_root:
     name: Detect root
-    platform: ubuntu1804
+    platform: ubuntu2004
     working_directory: test/detect_root_test
     test_targets:
       - "//:tests"
   ubuntu1804_detect_root:
     name: Detect root
-    platform: ubuntu1604
+    platform: ubuntu1804
     working_directory: test/detect_root_test
     test_targets:
       - "//:tests"
--- a/.bazelci/config.yaml
+++ b/.bazelci/config.yaml
@@ -1,69 +1,102 @@
 ---
 tasks:
-  ubuntu1804:
+  ubuntu2004:
+    platform: ubuntu2004
+    build_targets:
+      - "//..."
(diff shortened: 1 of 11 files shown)


.bazelci/config.yaml
```diff
@@ -1,69 +1,102 @@
...
     platform: windows
     working_directory: examples
     test_flags:
       - "--enable_runfiles"
     test_targets:
       - "//:win_tests"
-  ubuntu1804_flags:
+  ubuntu2004_flags:
+    name: Flags
     platform: ubuntu1804
```
jsharpe, line 32: Needs updating to 2004

.bazelci/config.yaml
```diff
@@ -1,69 +1,102 @@
...
     working_directory: test/standard_cxx_flags_test
     test_targets:
       - "//:flags_test"
-  ubuntu1604_flags:
+  ubuntu1804_flags:
+    name: Flags
     platform: ubuntu1604
```
jsharpe, line 37: Needs update to 1804

.bazelci/config.yaml
```diff
@@ -1,69 +1,102 @@
...
   windows_flags:
+    name: Flags
     platform: windows
     working_directory: test/standard_cxx_flags_test
     test_targets:
       - "//:flags_test"
-  ubuntu1804_detect_root:
+  ubuntu2004_detect_root:
+    name: Detect root
     platform: ubuntu1804
```
jsharpe, line 52: Need updating to 2004

.bazelci/config.yaml
```diff
@@ -1,69 +1,102 @@
...
     working_directory: test/detect_root_test
     test_targets:
       - "//:tests"
-  ubuntu1604_detect_root:
+  ubuntu1804_detect_root:
+    name: Detect root
     platform: ubuntu1604
```
jsharpe, line 57: Needs updating to 1804

.bazelci/config.yaml
```diff
@@ -73,4 +106,21 @@ tasks:
     platform: ubuntu2004
     working_directory: docs
     build_targets:
-    - //...
\ No newline at end of file
+    - //...
+  min_supported_version:
+    name: "Minimum Supported Version"
+    bazel: "3.4.0"
+    platform: ubuntu1604
```
jsharpe, line 113: I assume this is 1604 because this is the only place where 3.4.0 is available in CI?

jsharpe (changes requested):

.bazelci/config.yaml
UebelAndre: I don't think so. I didn't want to drop support for 1604 here so I use it for the min version tests. If we decide we don't aren't going to test on that environment, I can easily change it but wanted that to be a well communicated decision.

UebelAndre:
Thanks for the review, updated

jsharpe (approved):

PR #474, opened 2021-01-25
Add boost examples to tests and fix bootstrap parameter

jsharpe:
Actually test `boost_build` rule in CI build.

Fix incorrect parameter passed to bootstrap: `s/toolchain/toolset/`


.bazelci/config.yaml
```diff
@@ -43,6 +47,8 @@ tasks:
     working_directory: examples
     build_targets:
       - "//cmake_android:app"
+      - "//boost:boost_regex"
+      #- "//boost:boost_filesystem"
```
UebelAndre, line 51: What happened with this test?

UebelAndre (changes requested):
Is this something we want to enable? I had the same thought for the `cmake_build` and `make_build` targets and they were pretty expensive to build. I agree they should be tested but I'd hate to trigger [20 minute PR builds](https://buildkite.com/bazel/rules-foreign-cc/builds/1487#2ef6bbe6-6167-4c14-a68e-ee90021d0431) unnecessarily. 

These targets have also got me wondering why there isn't just a `http_file` rule for fetching a prebuilt binary/bundle for things like `cmake`, `ninja`, and potentially `boost`. Are these build rules really that beneficial to people?

.bazelci/config.yaml
jsharpe: It was passing invalid arguments to the build; because on macOS clang masquerades as g++ by default then boost's build system is picking the wrong compiler flags. As a result this target fails to build on macOS

PR #475, opened 2021-01-26
Fix test target on macOS with Bazel@HEAD

meteorcloudy:
supports_dynamic_linker feature was removed from cc toolchain for macOS,
cc_binary should be used to build shared library.

Context: bazelbuild/bazel#4341 (comment)

### DIFF
--- a/examples/cmake_with_data/lib_b/BUILD
+++ b/examples/cmake_with_data/lib_b/BUILD
@@ -1,9 +1,11 @@
-load("@rules_cc//cc:defs.bzl", "cc_library")
+load("@rules_cc//cc:defs.bzl", "cc_binary")
 
-cc_library(
+cc_binary(
     name = "lib_b",
-    srcs = ["src/lib_b.cpp"],
-    hdrs = ["src/lib_b.h"],
-    linkstatic = False,
+    srcs = [
+        "src/lib_b.cpp",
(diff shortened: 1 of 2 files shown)


jsharpe:
Thanks @meteorcloudy. This patch doesn't seem to work on windows though, could you take another look?

meteorcloudy:
@jsharpe Fixed now! The shared library on Windows should end with ".dll" ;)

jsharpe:
Thanks @meteorcloudy - yes strange that it was working with loading the interface lib instead of dll!

UebelAndre:
My bad 😅

Do you know why a `cc_binary` produces that?

meteorcloudy:
@UebelAndre  Do you mean why does a cc_binary output a dll?

UebelAndre:
> @UebelAndre Do you mean why does a cc_binary output a dll?

Yeah

meteorcloudy:
Oh, it's just the way to build a shared binary with Bazel, you can use a `cc_binary` rule with `linkshared = True`, then instead of building an executable binary, it builds a shared library (.so for Linux, .dll for Windows)
Check the `linkshared` attribute here: https://docs.bazel.build/versions/master/be/c-cpp.html#cc_binary

UebelAndre:
Thanks! Good to know 😄

PR #480, opened 2021-02-01
Ignore Vim swap files

attilaolah:
Minor change to the `.gitignore` to make my life easier.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,6 +68,9 @@ fabric.properties
 # Android studio 3.1+ serialized cache file
 .idea/caches/build_file_checksums.ser
 
+# Vim swap files
+.*.sw?
+
 #####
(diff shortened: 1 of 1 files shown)


UebelAndre (approved):
I actually think it's better to keep IDE patterns out if repos and to setup a [global gitignore](https://stackoverflow.com/questions/7335420/global-git-ignore) but seems like that ship has long since passed for this repo 😆 

Plus, you've been making regular updates so happy to accommodate. Thanks! 🙏

attilaolah:
Fair point. I just committed https://github.com/attilaolah/dotfiles/commit/19cb5d4b30755d6e03ce7153919125a304c445b1 which should now properly set up my global `.gitignore`. So no need for future changes like this.

But thanks for accepting anyway :)

whs-dot-hk:
This is insane, how about emacs, visual studio code, etc, I think you guys are spam merging

cc @google-admin

PR #487, opened 2021-02-03
Fix osx_commands shell script

jlaxson:
There is a typo in the shell script generated for copying a symlink after build.

Without this, `$actual` is empty and so bash interprets `local =$(readlink "$child")` which is invalid.

### DIFF
--- a/.bazelci/config.yaml
+++ b/.bazelci/config.yaml
@@ -18,7 +18,7 @@ tasks:
       - "//..."
     test_targets:
       - "//..."
-      # TODO: TODO make issue
+      # TODO: https://github.com/bazelbuild/rules_foreign_cc/issues/495
       - "-//test:shell_method_symlink_contents_to_dir_test"
       - "-//test:shell_script_inner_fun_test"
   ubuntu2004_examples:
@@ -27,6 +27,7 @@ tasks:
     working_directory: examples
     build_targets:
       - "//cmake_android:app"
+      - "//cmake_defines:all"
     test_targets:
       - "//:tests"
   ubuntu1804_examples:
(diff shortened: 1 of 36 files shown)


UebelAndre:
Thanks! Is there an easy test case you could add so this doesn't regress?

crncnnr:
This seems to be a duplicate of the same bug I noticed in #424. What sort of tests are you wanting? To check that this change doesn't break anything? Or a test to stop this happening again?

UebelAndre:
> This seems to be a duplicate of the same bug I noticed in #424. What sort of tests are you wanting? To check that this change doesn't break anything? Or a test to stop this happening again?

Yeah, I think these scripts are pretty easy to break since it's a lot of boiler plate and would hate for another change to be made to introduce a regression. It seems like you have a use case that exercises a code path not included in testing. If your use case was simple, I'd love to have it added.

I think this is a good change and want to get it merged, just always on the lookout to prevent regressions.

Aside from that, seems like CLA needs to be addressed and your branch updated 😄

crncnnr:
I've updated #424 and the CLA should be fine over there.

It looks like the problematic code path is `copy_dir_contents_to_dir`, and looking at `test/` only `symlink_contents_to_dir` is tested.  I could add a similar test, or were you thinking a real-world test? I noticed that there's a bunch of testing happening in `examples/`.

UebelAndre:
>  I could add a similar test, or were you thinking a real-world test? I noticed that there's a bunch of testing happening in `examples/`.

Just a simple test case that would fail before your change and pass after your change.

jlaxson:
I'm still working on validating this change.  There is at minimum one more thing broken - the `find` call below needs a trailing slash or some other way of explicitly traversing symlinks, otherwise it sometimes fails to copy sources of dependencies into position.

UebelAndre:
https://github.com/bazelbuild/rules_foreign_cc/pull/424 was merged which contains the same change. Thanks for all the PR though!

PR #488, opened 2021-02-03
Fixed incorrect detection of lib_source root

UebelAndre:
This fixes some inconsistent behavior I was seeing in how the source root was detected. Using `exclude_directories=0` on [glob](https://docs.bazel.build/versions/master/be/functions.html#glob) returns incorrect results in that directories will have [File.is_source](https://docs.bazel.build/versions/master/skylark/lib/File.html#is_source) set to `True`. This should be `False` and [File.is_directory](https://docs.bazel.build/versions/master/skylark/lib/File.html#is_directory) should be `True` instead.

The original implementation seems to have been introduced in https://github.com/bazelbuild/rules_foreign_cc/pull/57 and seemed to try and account for the use of `exclude_directories=0` without the correct [File.is_directory](https://docs.bazel.build/versions/master/skylark/lib/File.html#is_directory) behavior. I would rather file an upstream bug and have this setup to use the correct behavior so users don't run into unexpected behavior.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,4 +79,6 @@ fabric.properties
 /bazel-*
 /docs/bazel-*
 /examples/bazel-*
+/test/detect_root_test/bazel-*
+/test/standard_cxx_flags_test/bazel-*
 .ijwb/
--- a/examples/cmake_working_dir/BUILD
+++ b/examples/cmake_working_dir/BUILD
@@ -3,10 +3,7 @@ load("@rules_foreign_cc//tools/build_defs:cmake.bzl", "cmake_external")
 
 filegroup(
     name = "sources",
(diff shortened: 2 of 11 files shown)


UebelAndre:
I've opened https://github.com/bazelbuild/bazel/issues/12954 to track the root issue upstream.

whs-dot-hk:
How this help when it is reducing functionality?

jsharpe:
> How this help when it is reducing functionality?

Has this change caused a regression for you @whs-dot-hk ?

whs-dot-hk:
This is a typical case of changing a test for the code... (Happens when someone don't really understand the underlying code)

PR #489, opened 2021-02-03
Allow setting env vars

illicitonion:
Also, bring all runfiles of data deps along for execution - currently
data deps which consist of more than one file aren't copied into the
action, which means that they will fail at runtime when they try to use
their runfiles.


tools/build_defs/framework.bzl
```diff
@@ -87,6 +87,14 @@ CC_EXTERNAL_RULE_ATTRIBUTES = {
         allow_files = True,
         default = [],
     ),
+    "env": attr.string_dict(
+        doc = (
+            "Environment variables to set during the build. " +
+            "$(location) macros may be used to point at files which are listed as data deps, but unlike with other rules, " +
```
UebelAndre, line 93: For my own curiosity, is not not recommended to use something like `execpath` and `rootpath`?
https://docs.bazel.build/versions/master/be/make-variables.html#predefined_label_variables

I personally have always used `location` but don't know if that's really "legacy" behavior. Is there something "modern" we could write in the docs?

UebelAndre (approved):
LGTM but wanna get @jsharpe to take a look too 😄

tools/build_defs/framework.bzl
illicitonion: Good call - I always forget that there are other macros! Switched to only support `execpath` because that's the thing we're really clearly actually supporting (by literally prepending the exec root!)

UebelAndre:
Oh! One thing real quick, can you `cd docs` and run `bazel run :generate_docs`? 🙏

tools/build_defs/framework.bzl
```diff
@@ -309,12 +317,21 @@ def cc_external_rule_impl(ctx, attrs):
...
         "export EXT_BUILD_ROOT=##pwd##",
         "export BUILD_TMPDIR=##tmpdir##",
         "export EXT_BUILD_DEPS=##tmpdir##",
         "export INSTALLDIR=$$EXT_BUILD_ROOT$$/" + empty.file.dirname + "/" + lib_name,
+    ] + [
+        "export {key}={value}".format(
+            key=key,
+            # Prepend the exec root to each $(execpath ) lookup because the working directory will not be the exec root.
+            value=ctx.expand_location(value.replace("$(execpath ", "$$EXT_BUILD_ROOT$$/$(execpath "), data_dependencies)
+        )
```
jsharpe, line 333: The potential issues I can see with this is:

1.  by encoding the full path we are reducing the chances of getting reproducible builds which reduces the effectiveness of caching for downstream targets - but this is a general problem with `rules_foreign_cc` rules so I'm not expecting this to actually be resolved here but:
2.  these are `data` deps being expanded and so these should not be executed as part of the build as they will be for `target` not `exec` configurations and as such all that can be done with such env variables that point at executable code is for the path to be baked into outputs and as such reduce the reproducibility as described in `1.`
3. We should also expand targets from `ctx.attr.tools_deps` and `ctx.attr.additional_tools` (as an aside: why are there two attributes for build tools? - and why is it not called `ctx.attr.tools` as per other rules?) as these are the dependencies that are feasible to be ran as part of the build process. 

Having said the above this is definitely heading in the right direction IMO and am happy with this if the target list is updated to add the tools attributes too.

tools/build_defs/framework.bzl
```diff
@@ -87,6 +87,14 @@ CC_EXTERNAL_RULE_ATTRIBUTES = {
         allow_files = True,
         default = [],
     ),
+    "env": attr.string_dict(
+        doc = (
+            "Environment variables to set during the build. " +
+            "$(execpath) macros may be used to point at files which are listed as data deps, but unlike with other rules, " +
```
jsharpe, line 93: Add tools to this doc string - as per comment below.

tools/build_defs/framework.bzl
```diff
@@ -365,7 +382,7 @@ def cc_external_rule_impl(ctx, attrs):
             empty.file,
             wrapped_outputs.log_file,
         ],
-        tools = depset([wrapped_outputs.script_file] + ctx.files.data),
+        tools = depset([wrapped_outputs.script_file], transitive = [data[DefaultInfo].default_runfiles.files for data in data_dependencies]),
```
jsharpe, line 385: Just checking - does `default_runfiles.files` include the data files themselves?

It'd be good to have a test that actually exercises the runfiles from data as I think the existing test will only check the data files itself and only at runtime. I haven't checked but I don't think the current test for `data` uses the files at buildtime? @UebelAndre?

jsharpe (changes requested):
Generally agree with the direction of this.
Need to add in expansion of execpath for tools too though and updating the generated docs as requested by @UebelAndre.

whs-dot-hk:
Hi @illicitonion, The `configure_env_vars` in `configure_make` and `cache_entries` in `cmake` aren't enough?

illicitonion:
> Hi @illicitonion, The `configure_env_vars` in `configure_make` and `cache_entries` in `cmake` aren't enough?

Unfortunately not, I am building some code which requires an env var to be set for both `configure` and `make` (and which uses `make` not `cmake`).

tools/build_defs/framework.bzl
illicitonion: > The potential issues I can see with this is:
> 
> 1. by encoding the full path we are reducing the chances of getting reproducible builds which reduces the effectiveness of caching for downstream targets - but this is a general problem with `rules_foreign_cc` rules so I'm not expecting this to actually be resolved here but:

👍 

> 3. We should also expand targets from `ctx.attr.tools_deps` and `ctx.attr.additional_tools`

Done

> (as an aside: why are there two attributes for build tools? - and why is it not called `ctx.attr.tools` as per other rules?) [...]  as these are the dependencies that are feasible to be ran as part of the build process.

I'm not sure... `additional_tools` doesn't transition away from the target configuration, though... (Add `tools_deps` is currently `host` rather than `exec`, which we should probably clean up at some point too :))

> Having said the above this is definitely heading in the right direction IMO and am happy with this if the target list is updated to add the tools attributes too.

Done - thanks!

tools/build_defs/framework.bzl
illicitonion: As far as I can tell, it only does for _rules_ - if you have a dependency on a _file_ label, it isn't included.

This surfaced when I wrote that test you asked for :)

illicitonion:
@jsharpe I think this is now good to go? Let me know if you want any other changes :)

jsharpe (approved):
LGTM

Issue #301, opened 2019-07-31
Propagating compiler -D'efines to dependent native bazel cc_library components from cmake_external dependencies?


wrowe:
Simple case, the grpc/grpc component compiles against the c-ares project, built using
foreign_cc with the following syntax, and we have listed CARES_STATICLIB as defines to
prevent the symbols from being imported when linking monolithicly;

```
envoy_cmake_external(
    name = "ares",
    cache_entries = {
        "CARES_SHARED": "no",
        "CARES_STATIC": "on",
        "CARES_MSVC_STATIC_RUNTIME": "on",
        "CMAKE_BUILD_TYPE": "RelWithDebInfo",
    },
    copy_pdb = True,
    lib_source = "@com_github_c_ares_c_ares//:all",
    pdb_name = "c-ares",
    static_libraries = select({
        "//bazel:windows_x86_64": ["cares.lib"],
        "//conditions:default": ["libcares.a"],
    }),
    defines = ["CARES_STATICLIB"],
)
```

This is documented in https://github.com/bazelbuild/rules_foreign_cc/blob/master/README.md
```
    # Optional compilation definitions to be passed to the dependencies of this library.
    # They are NOT passed to the compiler, you should duplicate them in the configuration options.
    "defines": attr.string_list(mandatory = False, default = []),
```
Looking at the behavior of core bazel logic, it makes sense that these defines are not passed to the object they are declared in, but to all the dependent components and should be in the compiler options of those dependencies. E.g. a dynamic library object compilation might declare an EXPORT stub in building the component. But all dependent components must use a corresponding (different) IMPORT stub, which is one purpose of this defines=[] list.

In bazel/envoy_build_system.bzl envoy_cmake_external(), the **kwargs argument including any defines list is passed on to cmake_external().

This component is consumed then by a grpc/grpc rule which compiles against the resulting ares.h header file and requires that define noted above;
```
def _com_github_grpc_grpc():
    _repository_impl("com_github_grpc_grpc")
    native.bind(
        name = "cares",
        actual = "//external:ares",
    )
```

In spite of this, we see the following error indicating that the grpc_ares_wrapper.cc compilation during the com_github_grpc_grpc component is not given the -DCARES_STATICLIB flag as listed in the dependencies' defines;
```
ERROR: C:/workspace/envoy/source/exe/BUILD:27:1: Linking of rule '//source/exe:envoy-static' failed (Exit
4217)
[...]

Issue #421, opened 2020-08-05
rules_foreign_cc doesn't add -fPIC when --force_pic is specified


lizan:
Bazel adds -fPIC for other cc targets. This results link failure.

wrowe:
Note it would also help if envoy_cmake_external supported copts. It already supports linkopts, and this is a shortcoming relative to cc_library or cc_binary. This would make implementing workarounds simpler.

While it does support cache_entries, which can include CMAKE_CXX_FLAGS, and also supports env_vars, which can include CXXFLAGS, these are very difficult to manipulate using select().

jsharpe:
I guess one of the tricky issues here is that for native bazel targets it builds static libraries with both `-fPIC` and without depending upon the context within which it is linked. Now when you are relying on an external build system to replicate that behaviour it very much depends on the third party build system as to whether it natively adds (or indeed strips off) `-fPIC` from its builds. 
I guess a workaround to this could be to set a feature on the target that passes `-fPIC` from the configure cc toolchain.

wrowe:
Why has this report been ignored and allowed to expire? Please reopen, it's pretty basic and trivial

UebelAndre:
Sorry the report has gone unaddressed as long as it has. Unfortunately, `rules_foreign_cc` has no official support and those who do work on it do so in their free time. If the solution is simple, I'd be happy to collaborate on a pull request if someone would be willing to open one 😄 

I'll otherwise re-open for now though.

wrowe:
I will again ask on behalf of bazel consumers that you please do not close. If @lizan or I had assurance a patch might be applied, one of us might even volunteer to fix, but we are less than optimistic based on a few years of observation

PR #498, opened 2021-02-04
Propagate defines into CFLAGS and CXXFLAGS passed to the framework

jsharpe:
(Hopefully) Addresses issue highlighted in #301.

I suspect there are similar issues in propagating includes from deps `CcInfo.compilation_context`.

### DIFF
--- a/.bazelci/config.yaml
+++ b/.bazelci/config.yaml
@@ -27,6 +27,7 @@ tasks:
     working_directory: examples
     build_targets:
       - "//cmake_android:app"
+      - "//cmake_defines:all"
     test_targets:
       - "//:tests"
   ubuntu1804_examples:
@@ -35,6 +36,7 @@ tasks:
     working_directory: examples
     build_targets:
       - "//cmake_android:app"
+      - "//cmake_defines:all"
     test_targets:
       - "//:tests_no_synthetic"
   macos_examples:
@@ -43,6 +45,7 @@ tasks:
     working_directory: examples
     build_targets:
       - "//cmake_android:app"
+      - "//cmake_defines:all"
     test_targets:
       - "//:mac_tests"
   windows_examples:
--- a/examples/cmake_defines/BUILD
+++ b/examples/cmake_defines/BUILD
@@ -1,0 +1,25 @@
+load("@rules_foreign_cc//tools/build_defs:cmake.bzl", "cmake_external")
+
+cmake_external(
+    name = "lib_a",
+    lib_source = ":lib_a_sources",
+    static_libraries = ["liblib_a.a"],
+    deps = [":lib_b"],
+)
+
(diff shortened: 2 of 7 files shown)


UebelAndre:
Can we also incorporate the one of the simpler examples in that issue into this PR? I think it'd be valuable.

jsharpe:
> Can we also incorporate the one of the simpler examples in that issue into this PR? I think it'd be valuable.

The examples as is in the issue is incomplete (missing at least openssl from the deps for CURL), also a compilation only example wouldn't necessarily catch a regression - we'd need to explicitly check that curl was in fact being built with the correct defines set.

UebelAndre:
Wouldn't something like the following work?
```
                    +---------------+
  +->  lib_b  +-->  | lib_a         |
  |                 +---------------+
  +                 |               |
-DFOO               | #ifndef FOO   |
                    | #error NoFoo  |
                    | #endif        |
                    |               |
                    +---------------+

```

wrowe:
This appears to resolve #301 after inspecting the .obj.params intermediate input files (still building, so can't entirely confirm.)

The question I've had, for some time, is how to use select at line 305 much as I did at line 310 below. Obviously, select can't be used for dictionaries, only for lists. My thought is that adding copts accepting a list, same as linkopts, gets around this headache. Any other thoughts? 
https://github.com/wrowe/envoy/blob/9b46e213ddb8a32d838732da045f71a4af40653c/bazel/foreign_cc/BUILD#L305

wrowe:
So, might this also address the missing -fPIC flag? Observed by @lizan in https://github.com/bazelbuild/rules_foreign_cc/issues/421

jsharpe:
> So, might this also address the missing -fPIC flag? Observed by @lizan in #421

Unfortunately not - I think that #421 a separate issue that needs some investigation - we might need a `config_setting` to lookup the value of `--force_pic` and manually enable the `pic` feature from the toolchain, although this might be toolchain specific if you're not using the default toolchain.

jsharpe:
@UebelAndre PTAL - I've added a build test along the lines you suggested and verified it fails if I remove this patch.

.bazelci/config.yaml
```diff
@@ -27,6 +27,7 @@ tasks:
     working_directory: examples
     build_targets:
       - "//cmake_android:app"
+      - "//cmake_defines:all"
```
UebelAndre, line 30: Can't wait to upgrade the min version to 4.0 so we can just build `//...`

UebelAndre (approved):
I didn't know that `examples/cmake_defines/lib_b/lib_b.cpp` could be empty. I probably wrote more boiler plate than I had to in my time 😆 

Looks good! Thanks for adding the tests 🙏

PR #499, opened 2021-02-09
Remove version from generated script

jsharpe:
Having the version in the generated script will cause cache invalidations any time the version number is bumped even if the change causing the version bump doesn't affect the build being executed.

### DIFF
--- a/tools/build_defs/framework.bzl
+++ b/tools/build_defs/framework.bzl
@@ -303,7 +303,7 @@ def cc_external_rule_impl(ctx, attrs):
     if execution_os_name != "osx":
         set_cc_envs = "\n".join(["export {}=\"{}\"".format(key, cc_env[key]) for key in cc_env])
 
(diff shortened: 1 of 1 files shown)


tools/build_defs/framework.bzl
```diff
@@ -304,7 +303,7 @@ def cc_external_rule_impl(ctx, attrs):
     if execution_os_name != "osx":
         set_cc_envs = "\n".join(["export {}=\"{}\"".format(key, cc_env[key]) for key in cc_env])
 
-    version_and_lib = "Bazel external C/C++ Rules #{}. Building library '{}'\\n".format(VERSION, lib_name)
+    version_and_lib = "Bazel external C/C++ Rules. Building library '{}'\\n".format(lib_name)
```
UebelAndre, line 306: Can you also rename this variable?

jsharpe: Ah yes

UebelAndre (approved):
Looks good!

PR #503, opened 2021-02-11
Move location of generated logs and scripts

### LINKED ISSUE #483: gen_dir output group includes log files
The `gen_dir` output_group includes the log file and build script used. Its unlikely that the enduser wants these when consuming them via `gen_dir` we should remove them from the output_group.

jsharpe:
Fixes #483 by placing the logs outside of the install root which gets copied into a `directory` output.
Also removes the log files and build script from the `DefaultInfo` provider so that logs aren't propagated to downstream targets.


jsharpe:
@UebelAndre PTAL

UebelAndre (approved):

Issue #497, opened 2021-02-04
Add repository rules for fetching prebuilt build tools (`cmake`, `ninja`, etc...)


UebelAndre:
Currently users can either use there's a set of tools toolchains that either rely on a pre-installed build tool or compile one from source. In general, it's not always convenient to expect one of these tools to be installed on the host and it can be very time consuming to have to compile these tools. In addition to the following
https://github.com/bazelbuild/rules_foreign_cc/blob/14520d220f93f073bd1cf1d0c5dfbfd982d70b11/tools/build_defs/native_tools/BUILD#L7-L62

A new rule should be added that will download [cmake](https://cmake.org/download/) and [ninja](https://github.com/ninja-build/ninja/releases) for the current platform and make it available for users to register. A similar example of what I'm hoping for can be seen in [@rules_rust//rust:repositories.bzl](https://github.com/bazelbuild/rules_rust/blob/main/rust/repositories.bzl)

attilaolah:
It is true that compiling these from source, while seems like a good idea, can take forever.

Not to mention the rabbit hole it produces: compiling `make` requires `autotools`, `m4` and whatnot. Then building those too from source can result in a bootstrapping problem…

I could give this a try, but we first need to find out if these tools provide binaries for the major platforms. (I don't want to be in the business of having to maintain a binary release…)

UebelAndre:
My thought for this change was to write a basic shells script that generates a bzl file with repository definitions for a predefined set of cmake and ninja versions. Then update the interface for `rules_foreign_cc_dependencies` to be something like
```python
def rules_foreign_cc_dependencies(
        native_tools_toolchains = [],
        register_default_tools = True,
        cmake_version = "3.18.0",
        ninja_version = "1.10.2",
        additional_shell_toolchain_mappings = [],
        additional_shell_toolchain_package = None):
```

Where users can pick from our built in set of supported releases that will be downloaded from cmake/github/google with all the appropriate `sha256` values already defined.

Though I'm not sure what you mean with 

> I don't want to be in the business of having to maintain a binary release…

That said though, I haven't had the time to work on this but think it'd help immensely in making builds more hermetic. If you wanted to work on this I'd greatly appreciate it! 😄

attilaolah:
> Though I'm not sure what you mean with

What I meant there is that this would work as long as cmake/ninja/etc. provide binaries somewhere, on GitHub or some place else. I.e. we don't need to build and release binaries.

[...]

PR #507, opened 2021-02-15
Toolchains to download prebuilt cmake and ninja

jsharpe:

PR #485, opened 2021-02-02
Made buildifier checks more aggressive

UebelAndre:
Fixed all buildifier defects and updated CI to enforce changes remain buildifier compliant.

Users can run `buildifier -lint=fix -mode=fix -warnings=all -r ./` to make sure their changes are buildifier compliant

PR #508, opened 2021-02-15
Fetch prebuilt cmake and ninja binaries as registered default toolchains

UebelAndre:
closes https://github.com/bazelbuild/rules_foreign_cc/issues/497

This updates the default registered toolchains to use prebuilt cmake and ninja binaries published on github. This makes builds more hermetic

Note the `//toolchains:prebuilt_toolchains.py` script might not be needed but I think it helps streamline the exposure of new cmake/ninja releases


toolchains/prebuilt_toolchains.py
```diff
@@ -0,0 +1,370 @@
...
+    path = "$(execpath :cmake_bin)",
+    target = ":cmake_bin",
+)
+
+toolchain(
+    name = "cmake_toolchain",
+    exec_compatible_with = {{exec_compatible_with}},
+    toolchain = ":cmake_tool",
+    toolchain_type = "@rules_foreign_cc//tools/build_defs:cmake_toolchain",
+)
```
jsharpe, line 138: The issue with having the toolchain definition inside the repository is that this causes bazel to fetch tools for all platforms on all platforms. Can this be moved into `@rules_foreign_cc` as I have done in #507?

toolchains/prebuilt_toolchains.py
```diff
@@ -0,0 +1,370 @@
...
+from textwrap import indent
+import urllib.request
+import hashlib
+
+
+CMAKE_SHA256_URL_TEMPLATE = "https://cmake.org/files/v{minor}/cmake-{full}-SHA-256.txt"
+CMAKE_URL_TEMPLATE = "https://github.com/Kitware/CMake/releases/download/v{full}/{file}"
+
+CMAKE_VERSIONS = [
+    "3.19.4",
```
jsharpe, line 13: Nit: 3.19.5 was released just yesterday.

examples/deps/repositories.bzl
```diff
@@ -225,6 +225,7 @@ def repositories():
         name = "apr_util",
         build_file_content = all_content,
         strip_prefix = "apr-util-1.6.1",
+        sha256 = "b65e40713da57d004123b6319828be7f1273fbc6490e145874ee1177e112c459",
```
jsharpe, line 228: Just an observation - I'm not sure that Buildifier is checking the `examples` WORKSPACE. The ordering of where `sha256` is in this file is inconsistent which suggests Buildifier isn't running on this file.

toolchains/prebuilt_toolchains.py
```diff
@@ -0,0 +1,370 @@
...
+                    bin=bin,
+                    exec_compatible_with=CMAKE_TARGETS[plat_target],
+                )
+            )
+            version_toolchains.append("@{}//:cmake_toolchain".format(name))
+
+        archives.append("\n".join(
+            [
+                "    if \"{}\" == version:".format(version),
+            ] + [indent(archive, " " * 8) for archive in version_archives])
```
jsharpe, line 252: The version_archives expansion doesn't need to be protected by the if. There's no harm in defining all the repositories unconditionally and this allows an end user to refer to these to setup their own toolchain if necessary using these repositories e.g. they may want to setup a toolchain with an additional exec constraint so that they can select an older version of a tool for a particular build if the newest version doesn't work for them.

toolchains/prebuilt_toolchains.py
```diff
@@ -0,0 +1,370 @@
...
+                    bin="ninja.exe" if "win" in target else "ninja",
+                    exec_compatible_with=NINJA_TARGETS[target],
+                )
+            )
+            version_toolchains.append("@{}//:ninja_toolchain".format(name))
+
+        archives.append("\n".join(
+            [
+                "    if \"{}\" == version:".format(version),
+            ] + [indent(archive, " " * 8) for archive in version_archives])
```
jsharpe, line 325: Ditto as above

tools/build_defs/native_tools/BUILD
```diff
@@ -1,68 +1,65 @@
...
-native_tool_toolchain(
+alias(
     name = "built_cmake",
-    path = "cmake/bin/cmake",
-    target = ":cmake_tool",
+    actual = "//toolchains:built_cmake",
+    deprecation = _DEPRECATION_NOTICE,
+    tags = ["manual"],
     visibility = ["//visibility:public"],
 )
```
jsharpe, line 43: Happy to leave these here but I don't think anyone can be using these as I don't think the `built_*` targets for toolchains actually work.

toolchains/prebuilt_toolchains.py
```diff
@@ -0,0 +1,370 @@
...
+        "@platforms//os:windows",
+    ],
+}
+
+NINJA_URL_TEMPLATE = "https://github.com/ninja-build/ninja/releases/download/v{full}/ninja-{target}.zip"
+
+NINJA_TARGETS = {
+    "linux": [
+        "@platforms//os:linux",
+    ],
```
jsharpe, line 52: This needs constraining to x86_64 as this incorrectly gets selected on aarch64

toolchains/prebuilt_toolchains.py
```diff
@@ -0,0 +1,370 @@
...
+    "mac": [
+        "@platforms//os:macos",
+    ],
+    "win": [
+        "@platforms//os:windows",
+    ],
```
jsharpe, line 58: I also assume that this build is x86_64 (or maybe x86_32?) only and wouldn't work on arm64 windows.

toolchains/prebuilt_toolchains.py
```diff
@@ -0,0 +1,370 @@
...
+
+NINJA_URL_TEMPLATE = "https://github.com/ninja-build/ninja/releases/download/v{full}/ninja-{target}.zip"
+
+NINJA_TARGETS = {
+    "linux": [
+        "@platforms//os:linux",
+    ],
+    "mac": [
+        "@platforms//os:macos",
+    ],
```
jsharpe, line 55: I also assume this isn't a universal binary for all releases and so needs a x86_64 constraint.

workspace_definitions.bzl
```diff
@@ -45,8 +54,4 @@ def rules_foreign_cc_dependencies(
 
     native.register_toolchains(*native_tools_toolchains)
     if register_default_tools:
-        native.register_toolchains(
-            "@rules_foreign_cc//tools/build_defs:preinstalled_cmake_toolchain",
-            "@rules_foreign_cc//tools/build_defs:preinstalled_ninja_toolchain",
-            "@rules_foreign_cc//tools/build_defs:preinstalled_make_toolchain",
-        )
+        prebuilt_toolchains(cmake_version, ninja_version)
```
jsharpe, line 57: I think you also need to call `preinstalled_toolchains` here so that if there isn't a matching exec constraint for the platform e.g. ninja on aarch64 then we at least try to fallback to the preinstalled version. This should probably also be a parameter to the repo macro such as `use_preinstalled_tools`

jsharpe (changes requested):
Definitely heading the correct direction with this but still needs a bit of work to ensure we don't break some of the platforms not tested in CI e.g. aarch64 or FreeBSD.

I think I'd also like to eventually add the option to the repository macro to select using a built version of the tools but I've not actually got this working correctly yet and that can be done after this PR has landed.

toolchains/prebuilt_toolchains.py
UebelAndre: Good call, updated.

examples/deps/repositories.bzl
UebelAndre: I've made a TODO for myself to fix this. But I'd like to do it in a separate ticket. Forgot I even included this change in this PR.

toolchains/prebuilt_toolchains.py
UebelAndre: This is a universal binary. The previous releases used `Darwin-XX_XX` as the binary name. Anything `macos-universal` is universal

workspace_definitions.bzl
UebelAndre: Added an option to register preinstalled toolchains

toolchains/prebuilt_toolchains.py
UebelAndre: Would you be ok with leaving this as a later improvement?

UebelAndre: moving conversation from https://github.com/bazelbuild/rules_foreign_cc/pull/508#discussion_r576669270 here

toolchains/prebuilt_toolchains.py
UebelAndre: Moving conversation to https://github.com/bazelbuild/rules_foreign_cc/pull/508#discussion_r576668650

toolchains/prebuilt_toolchains.py
UebelAndre: At least for CMake. This has been added to ninja

examples/deps/repositories.bzl
UebelAndre: TODO is here: https://github.com/bazelbuild/rules_foreign_cc/pull/485#issuecomment-779887617

UebelAndre:
> Definitely heading the correct direction with this but still needs a bit of work to ensure we don't break some of the platforms not tested in CI e.g. aarch64 or FreeBSD.

Preinstalled tools will once again be used by default to reduce breakages.

tools/build_defs/native_tools/BUILD
UebelAndre: I'll remove this after the release

jsharpe (approved):

PR #512, opened 2021-02-17
Ensure symlink_contents_to_dir traverses symlinks on OSX

jlaxson:
If the `source` arg to symlink_contents_to_dir is itself a symlink, the `find` call wouldn't traverse it.  I believe I encountered this when trying to chain `configure_make` rules where the first rule generated a shared library. 

Another solution would be to pass `-L` to find, but that would make it traverse all symlinks, not just the root, and this seemed closer to original intent.

### DIFF
--- a/tools/build_defs/shell_toolchain/toolchains/impl/osx_commands.bzl
+++ b/tools/build_defs/shell_toolchain/toolchains/impl/osx_commands.bzl
@@ -99,7 +99,7 @@ elif [[ -L "$1" && ! -d "$1" ]]; then
 elif [[ -d "$1" ]]; then
   SAVEIFS=$IFS
   IFS=$'\n'
-  local children=($(find "$1" -maxdepth 1 -mindepth 1))
+  local children=($(find "$1/" -maxdepth 1 -mindepth 1))
   IFS=$SAVEIFS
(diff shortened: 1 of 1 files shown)


tools/build_defs/shell_toolchain/toolchains/impl/osx_commands.bzl
```diff
@@ -99,7 +99,7 @@ elif [[ -L "$1" && ! -d "$1" ]]; then
 elif [[ -d "$1" ]]; then
   SAVEIFS=$IFS
   IFS=$'\n'
-  local children=($(find "$1" -maxdepth 1 -mindepth 1))
+  local children=($(find "$1/" -maxdepth 1 -mindepth 1))
```
UebelAndre, line 102: Is this not an issue on Linux as well?

UebelAndre (changes requested):
One question so far and could I get you to somehow add a test for this functionality so it doesn't regress in the future? 🙏

UebelAndre:
I think this might have been solved for in https://github.com/bazelbuild/rules_foreign_cc/pull/603. @jlaxson could you try updating and see if your problem is resolved?

jsharpe:
This change was addressed in #603

PR #513, opened 2021-02-17
Example for CMake with Transitive Shared Libraries

jlaxson:
The existing example shows how things work for static libs; this expands it a bit to show building a shared library, injecting it into CMake, and getting everything just right so a final binary built on top of a CMake shared lib works.

### DIFF
--- a/examples/cmake_with_bazel_transitive/BUILD
+++ b/examples/cmake_with_bazel_transitive/BUILD
@@ -11,6 +11,7 @@ cmake_external(
         # as currently we copy all libraries, built with Bazel, into $EXT_BUILD_DEPS/lib
         # and the headers into $EXT_BUILD_DEPS/include
         "LIBA_DIR": "$EXT_BUILD_DEPS",
+        "LIBA_PATH": "$EXT_BUILD_DEPS/lib/liba.a",
     },
     cmake_options = ["-GNinja"],
     lib_source = "//cmake_with_bazel_transitive/libb:b_srcs",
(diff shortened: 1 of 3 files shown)


examples/cmake_with_bazel_transitive/BUILD
```diff
@@ -23,6 +24,30 @@ cmake_external(
...
+        #"LIBA_PATHx": "$EXT_BUILD_DEPS/lib/liba.so",
+    },
+    cmake_options = ["-GNinja", "--trace", "--debug-find"],
+    lib_source = "//cmake_with_bazel_transitive/libb:b_srcs",
+    make_commands = [
+        "ninja",
+        "ninja install",
+    ],
+    shared_libraries = ["libb.so"],
+    deps = ["//cmake_synthetic/liba:lib_a_so"],
```
UebelAndre, line 48: Is there a way to avoid the cross example boundary?

examples/cmake_with_bazel_transitive/BUILD
```diff
@@ -23,6 +24,30 @@ cmake_external(
...
+        "BUILD_SHARED_LIBS": "ON",
+        "CMAKE_SHARED_LIBRARY_SUFFIX_CXX": ".so", # easier consistency on macos
+        "CMAKE_MACOSX_RPATH": "True",
+        # CMake's find_package wants to find cmake config for liba,
+        # which we do not have -> disable search
+        "CMAKE_DISABLE_FIND_PACKAGE_LIBA": "True",
+        # as currently we copy all libraries, built with Bazel, into $EXT_BUILD_DEPS/lib
+        # and the headers into $EXT_BUILD_DEPS/include
+        "LIBA_DIR": "$EXT_BUILD_DEPS",
+        #"LIBA_PATHx": "$EXT_BUILD_DEPS/lib/liba.so",
```
UebelAndre, line 39: Is this comment still communicating something valuable?

examples/cmake_with_bazel_transitive/libb/src/CMakeLists.txt
```diff
@@ -2,8 +2,8 @@ cmake_minimum_required(VERSION 2.8.4)
 
 set(LIBB_SRC libb.cpp libb.h)
 
-add_library(libb_static STATIC ${LIBB_SRC})
-set_target_properties(libb_static PROPERTIES OUTPUT_NAME "libb")
+add_library(libb_static ${LIBB_SRC})
```
UebelAndre, line 5: Is there a reason to remove the explicit use of `STATIC`?

UebelAndre (changes requested):
Sorry for the delay! Thank you so much for adding an example! A couple of comments but so far it's looking pretty good 😄

Issue #505, opened 2021-02-13
configure_in_place causes (near?) infinite symlink recursion


attilaolah:
In #479 I've added [an exception](https://github.com/bazelbuild/rules_foreign_cc/pull/479/files#diff-3796b11de709dc4560c4cc5892d3385d3cdaac834856ce2aa33a05ad1fce3273L103-R102) for `.ext_build_deps`, maybe we need one for `.build_tempdir` as well, for the case when `configure_in_place` is set.

Also maybe a failing test case would be nice.

whs-dot-hk:
Maybe revert #479 ? Clean up function is always there...

jsharpe:
configure_with_bazel_transitive on macOS seems to exhibit this failure - its not enabled in CI though.

attilaolah:
It's odd that it would be platform-specific though. In the meantime, I found out that in my case the error was caused by something else, and not the symlink tree (there just happened to be a very deep directory layout, generated by `./configure` & `make`, for some reason.)

UebelAndre:
You should check out `tools/build_defs/shell_toolchain/toolchains/impl/osx_commands.bzl` in https://github.com/bazelbuild/rules_foreign_cc/pull/514. There's a snippet on line 128 that was there for every platform but MacOS.

attilaolah:
Oops, you're right! Looks like I missed adding that exception to `osx_commands.bzl`. Thanks for adding it!

UebelAndre:
Do you think that would cause the recursive behavior we're seeing?

attilaolah:
I think so.

PR #514, opened 2021-02-18
Fixed unbound variable in commands and recursive symlinks

UebelAndre:
closes https://github.com/bazelbuild/rules_foreign_cc/issues/505

- 0122473a8c59682a193427f5589ebf772e5fb57e fixes some unbound variables. This change is proven by the change in between [this build](https://buildkite.com/bazel/rules-foreign-cc/builds/1761#83c5581e-433d-40d2-a8ee-a5a245dd2bb3) and [this build](https://buildkite.com/bazel/rules-foreign-cc/builds/1764#8fb42e4c-6e09-4254-b48a-aecf111f42c2) where this change was cherry picked and fixed an error
- f9f230385cbcf2e854304815c44ccca4447d5f4b fixes recursive symlink issues on macos. This change is proven by the change in behavior between [this build](https://buildkite.com/bazel/rules-foreign-cc/builds/1764#0e264415-91c9-47dc-a721-3296d472607c) and [this build](https://buildkite.com/bazel/rules-foreign-cc/builds/1765#4ebd633f-e873-4c39-bb66-c165aaa780a3) where the change was cherry picked.

### DIFF
--- a/test/expected/inner_fun_text.txt
+++ b/test/expected/inner_fun_text.txt
@@ -12,7 +12,7 @@ IFS=$'
 '
 local children=($(find -H "$1" -maxdepth 1 -mindepth 1))
 IFS=$SAVEIFS
-for child in "${children[@]}"; do
+for child in "${children[@]:-}"; do
 symlink_to_dir "$child" "$target"
 done
 fi
@@ -32,7 +32,7 @@ local children=($(find -H "$1" -maxdepth 1 -mindepth 1))
 IFS=$SAVEIFS
 local dirname=$(basename "$1")
 mkdir -p "$target/$dirname"
-for child in "${children[@]}"; do
+for child in "${children[@]:-}"; do
 if [[ "$dirname" != *.ext_build_deps ]]; then
 symlink_to_dir "$child" "$target/$dirname"
 fi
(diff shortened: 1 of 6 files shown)


UebelAndre:
@attilaolah this seems to solve for the recursive symlinks issue. Do you have a test case you could run too?

attilaolah:
No, sadly not: My failing test case was `libgrass`, but since then it turned out to be some linking problem, and I couldn't reproduce this issue (but I also cannot build `libgrass` until I find the right incantation of linker flags.)

test/expected/inner_fun_text.txt
```diff
@@ -32,8 +32,8 @@ local children=($(find -H "$1" -maxdepth 1 -mindepth 1))
 IFS=$SAVEIFS
 local dirname=$(basename "$1")
 mkdir -p "$target/$dirname"
-for child in "${children[@]}"; do
-if [[ "$dirname" != *.ext_build_deps ]]; then
+for child in "${children[@]:-}"; do
+if [[ "$dirname" != *.ext_build_deps && "$dirname" != *.build_tempdir ]]; then
```
attilaolah, line 36: This is a great addition, although I'm a little worried that we're adding these as blanket exception. One day someone will decide to name their files like this and we'll just skip them here.

It would be nice if we had the full path and passed it in to the Starlark function that generates the shell function, that would be more future proof.

But for now this is a great addition, thanks for adding it here!

UebelAndre: I've actually decided to drop this particular change `&& "$dirname" != *.build_tempdir` since it wasn't necessary to fix the problem I was able to reproduce

jsharpe (approved):

PR #521, opened 2021-02-24
Built tools

jsharpe:
Now that we can build make from source enable that as a default option.
Also enables built_cmake and built_ninja where a prebuilt toolchain option doesn't exist.


UebelAndre:
One thing I was thinking of doing was also creating a file similar to `prebuilt_toolchains.bzl` but instead of having definitions for prebuilt bundles, it has definitons for the raw source code. `built_toolchains.bzl` or something, and that'd also respect the `cmake_version`, `ninja_version`, and additionally `make_version` for grabbing the requested version of the source code.

.bazelci/config.yaml
```diff
@@ -39,6 +39,15 @@ tasks:
       - "//cmake_defines:all"
     test_targets:
       - "//:tests_no_synthetic"
+  rbe_ubuntu1604_examples:
```
UebelAndre, line 42: 🙌 Oh snap! 🙌

toolchains/toolchains.bzl
```diff
@@ -5,6 +5,15 @@ load(":prebuilt_toolchains.bzl", _prebuilt_toolchains = "prebuilt_toolchains")
 # Re-expose the prebuilt toolchains macro
 prebuilt_toolchains = _prebuilt_toolchains
 
+# buildifier: disable=unnamed-macro
+def built_toolchains():
```
UebelAndre, line 9: Can you also make a `built_toolchains.bzl` file that has a similar structure to `prebuilt_toolchains.bzl` and defines all the source locations for the various build tools? 🙏

for_workspace/make_build.bzl
```diff
@@ -50,6 +53,7 @@ make_tool = rule(
             doc = "target with the Make sources",
             mandatory = True,
         ),
+        "_cc_toolchain": attr.label(default = Label("@bazel_tools//tools/cpp:current_cc_toolchain")),
```
UebelAndre, line 56: Shouldn't this just go in the list of toolchains below?

jsharpe: No; see https://docs.bazel.build/versions/master/integrating-with-rules-cc.html

.bazelci/config.yaml
jsharpe: I'm going to remove this from this PR and handle enabling this separately.

for_workspace/cmake_build.bzl
```diff
@@ -38,10 +38,11 @@ cmake_tool = rule(
     attrs = {
         "cmake_srcs": attr.label(mandatory = True),
     },
-    fragments = ["cpp"],
+    host_fragments = ["cpp"],
```
UebelAndre, line 41: What does `fragments` -> `host_fragments` do?

for_workspace/make_build.bzl
```diff
@@ -1,28 +1,41 @@
...
 load("@rules_foreign_cc//tools/build_defs:shell_script_helper.bzl", "convert_shell_script")
 load("//tools/build_defs:detect_root.bzl", "detect_root")
 
 def _make_tool(ctx):
     root = detect_root(ctx.attr.make_srcs)
 
+    cc_toolchain = find_cpp_toolchain(ctx)
+
+    # we need this fictive file in the root to get the path of the root in the script
+    empty = fictive_file_in_genroot(ctx.actions, ctx.label.name)
```
UebelAndre, line 17: Can you elaborate a bit more on what's happening here?

UebelAndre (changes requested):
Just a couple of questions but otherwise looks good to me 😄

for_workspace/make_build.bzl
jsharpe: This is a copy of what is done in `framework.bzl` basically its creating an empty output file so that the script can detect the folder to use for INSTALLDIR.

for_workspace/cmake_build.bzl
jsharpe: This will access the `cpp` fragments in the `host` configuration rather than `target` since this rule will always be building make for the `host` (or `exec`) configuration. Which actually means that I've got the `"_cc_toolchain"` attribute wrong now I think about it - it should be `current_cc_host_toolchain` rather than `current_cc_toolchain`?

I wasn't 100% sure on this though; we may want to rely on the fact that this will referenced via a host configuration and so the toolchain should be set correctly anyway and we shouldn't need to explicitly select the host toolchain. 

Either way these rules aren't actually properly using the toolchain so this area does need some improvement.

PR #522, opened 2021-02-24
Update default cmake to 3.19.6

jsharpe:
I've left 3.19.5 in the repo list - I'm not sure if this is the correct thing to do or not in terms of not breaking users when they upgrade if they've explicitly listed a version?


UebelAndre (approved):
I think it's fine to leave it. And we should so we don't break users.

PR #523, opened 2021-02-24
Make generate_crosstool=True the default for cmake.

jsharpe:
This was previously setting `CMAKE_SYSTEM_NAME` in the generated
toolchain file but setting this explicitly enables `CMAKE_CROSSCOMPILING`
(see https://cmake.org/cmake/help/v3.19/variable/CMAKE_CROSSCOMPILING.html#variable:CMAKE_CROSSCOMPILING)
which breaks projects that use `TRY_RUN`


UebelAndre:
What's the risk to users not having set this at all? Should things still be expected to work without issue?

jsharpe:
Without `CMAKE_SYSTEM_NAME` set I think this should behave the same as setting the values via the command line. The impact to users is that they no longer have to do the select dance on this variable to support windows.

attilaolah:
I think this may have broken the use of the Emscripten toolchain from https://github.com/emscripten-core/emsdk/blob/master/bazel/emscripten_toolchain.

Before this CL, I'm not exactly sure how CMake was doing the linking, but the linker was invoked in some different way, or maybe the linker configs were not passed down correctly from the C++ toolchain? I'm not sure.

But since this change (I think), CMake now tries to use a link command like this:

```txt
-- Check for working CXX compiler: /root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/sandbox/processwrapper-sandbox/55/execroot/wasm/external/emsdk/emscripten_toolchain/emcc.sh                                                                                            
-- Check for working CXX compiler: /root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/sandbox/processwrapper-sandbox/55/execroot/wasm/external/emsdk/emscripten_toolchain/emcc.sh - broken                                                                                   
CMake Error at /root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/external/cmake-3.19.6-Linux-x86_64/share/cmake-3.19/Modules/CMakeTestCXXCompiler.cmake:59 (message):                                                                                                       
  The C++ compiler                                                                                                                           
                                                                                                                                                                                                                                                                                          
    "/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/sandbox/processwrapper-sandbox/55/execroot/wasm/external/emsdk/emscripten_toolchain/emcc.sh"                                                                                                                         
                                                                                                                                                                                                                                                                                          
  is not able to compile a simple test program.                                                                                                                                                                                                                                           
                                                                                                                                             
  It fails with the following output:
                                                                      
    Change Dir: /root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/sandbox/processwrapper-sandbox/55/execroot/wasm/bazel-out/wasm32-fastbuild/bin/lib/zstd/zstd_lib.build_tmpdir/CMakeFiles/CMakeTmp                                                                         
                                                                                                                                             
    Run Build Command(s):/usr/bin/gmake cmTC_49c94/fast && /usr/bin/gmake  -f CMakeFiles/cmTC_49c94.dir/build.make CMakeFiles/cmTC_49c94.dir/build
    gmake[1]: Entering directory '/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/sandbox/processwrapper-sandbox/55/execroot/wasm/bazel-out/wasm32-fastbuild/bin/lib/zstd/zstd_lib.build_tmpdir/CMakeFiles/CMakeTmp'
    Building CXX object CMakeFiles/cmTC_49c94.dir/testCXXCompiler.cxx.o                                                                      
    /root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/sandbox/processwrapper-sandbox/55/execroot/wasm/external/emsdk/emscripten_toolchain/emcc.sh --sysroot=/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/sandbox/processwrapper-sandbox/55/execroot/wasm/
external/emscripten/emscripten/cache/sysroot   --sysroot=/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/sandbox/processwrapper-sandbox/55/execroot/wasm/external/emscripten/emscripten/cache/sysroot -fdiagnostics-color -fno-exceptions -fno-strict-aliasing -funsigned-
char -no-canonical-prefixes -std=gnu++17 -nostdinc -nostdinc++ -fomit-frame-pointer -O2 -Wall -iwithsysroot/include/c++/v1 -iwithsysroot/include/compat -iwithsysroot/include -isystem /root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/sandbox/processwrapper-sandbox/55/e
xecroot/wasm/external/emscripten/lib/clang/12.0.0/include -Wno-builtin-macro-redefined -D__DATE__="redacted" -D__TIMESTAMP__="redacted" -D__TIME__="redacted" -Werror  -o CMakeFiles/cmTC_49c94.dir/testCXXCompiler.cxx.o -c /root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f560
98/sandbox/processwrapper-sandbox/55/execroot/wasm/bazel-out/wasm32-fastbuild/bin/lib/zstd/zstd_lib.build_tmpdir/CMakeFiles/CMakeTmp/testCXXCompiler.cxx
    Linking CXX executable cmTC_49c94 
    /root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/external/cmake-3.19.6-Linux-x86_64/bin/cmake -E cmake_link_script CMakeFiles/cmTC_49c94.dir/link.txt --verbose=1
    /root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/sandbox/processwrapper-sandbox/55/execroot/wasm/external/emsdk/emscripten_toolchain/emcc_link.sh --sysroot=/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/sandbox/processwrapper-sandbox/55/execroot/
wasm/external/emscripten/emscripten/cache/sysroot -fdiagnostics-color -fno-exceptions -fno-strict-aliasing -funsigned-char -no-canonical-prefixes -std=gnu++17 -nostdinc -nostdinc++ -fomit-frame-pointer -O2 -Wall -iwithsysroot/include/c++/v1 -iwithsysroot/include/compat -iwithsysroo
t/include -isystem /root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/sandbox/processwrapper-sandbox/55/execroot/wasm/external/emscripten/lib/clang/12.0.0/include -Wno-builtin-macro-redefined -D__DATE__="redacted" -D__TIMESTAMP__="redacted" -D__TIME__="redacted" -Werro
r  --sysroot=/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/sandbox/processwrapper-sandbox/55/execroot/wasm/external/emscripten/emscripten/cache/sysroot -fdiagnostics-color -fno-strict-aliasing -funsigned-char -no-canonical-prefixes -s PRINTF_LONG_DOUBLE=1 -O2  CMa
keFiles/cmTC_49c94.dir/testCXXCompiler.cxx.o -o cmTC_49c94                                                                                                                                                                                                                                
      File "/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/sandbox/processwrapper-sandbox/55/execroot/wasm/external/emsdk/emscripten_toolchain/link_wrapper.py", line 24, in <module>
        assert sys.argv[1][0] == '@'                                                                                                                                                                                                                                                      
    AssertionError                                                                                                                                                                                                                                                                        
    gmake[1]: *** [CMakeFiles/cmTC_49c94.dir/build.make:106: cmTC_49c94] Error 1
    gmake[1]: Leaving directory '/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/sandbox/processwrapper-sandbox/55/execroot/wasm/bazel-out/wasm32-fastbuild/bin/lib/zstd/zstd_lib.build_tmpdir/CMakeFiles/CMakeTmp'
    gmake: *** [Makefile:140: cmTC_49c94/fast] Error 2
```

I think the reason this fails is because the Emscripten toolchain expets Bazel to pass a single argument to the linker, containing only the linker param file: https://github.com/emscripten-core/emsdk/blob/master/bazel/emscripten_toolchain/crosstool.bzl.

Then the "link wrapper", https://github.com/emscripten-core/emsdk/blob/master/bazel/emscripten_toolchain/link_wrapper.py, checks that there is only a single argument, and that it starts with "@" and assert-fails right there.

I think a workaround for this is to set `"CMAKE_TRY_COMPILE_TARGET_TYPE": "STATIC_LIBRARY"` in `cache_entries`, but setting that everywhere just to work around this link flag seems bad, since Emscripten can in fact produce a working Binary: a JS file that can be executed with NodeJS, and CMake should be able to detect that. (It somehow worked before.)

Alternatively, I could try and generate a linker param file myself in the shell script that acts as the "linker" in the Bazel toolchain: https://github.com/emscripten-core/emsdk/blob/master/bazel/emscripten_toolchain/emcc_link.sh. I'm not sure how linker param files look like though, but it could be a way to support this use case.

In any case, don't revert this change: I just need to find out how to work around the issue, and documenting the problem here in case someone else bumps into the same problem.

attilaolah:
Just confirmed, either of the solutions fixes my breakage:

Option 1: Pass `cache_entries = {"CMAKE_TRY_COMPILE_TARGET_TYPE": "STATIC_LIBRARY"}`.
Option 2: Pass `generate_crosstool_file = False`.

In any case, I think the solution going forward will be to fix that link wrapper shell script to pass the arguments as a temporary file, or even better, to make the Python script understand regular `sys.argv` args.

UebelAndre:
Can you elaborate on the solution a bit more?

attilaolah:
If I pass either of those parameters from [my previous comment](#issuecomment-786964026) to `cmake_external()`, the Emscripten build passes. I think it is simply because it either does not try to link any binary in the first case, or in the second case, it somehow figures out what linker to use based on my `emcmake cmake` invocation, instead of using the linker provided by the CC toolchain, which in my case is broken.

So there are several things at work here:

1. Bazel CC toolchain, using Clang by default.
2. Bazel CC toolchain for `--cpu=wasm32` which uses an Emscripten-based toolchain from https://github.com/emscripten-core/emsdk/tree/master/bazel. This by default is somewhat broken for my use case, for several reasons (see below).
3. rules_foreign_cc CMake toolchain using whatever default CC toolchain is configured.
4. My custom toolchain implementing the cmake_toolchain type, that calls the Emscripten wrapper for CMake (i.e. it calls `emcmake cmake …`).

Now, without any toolchain magic, when I cross-compile my `cmake_external()` rule, I end up calling `emcmake cmake`, which runs `cmake` with a bunch of environmment variables and args that tell it to use the Emscripten toolchain.

However, when we write a CMake toolchain, rules_foreign_cc will tell CMake to use a CMake toolchain file that it writes based on the Emscripten (CC) toolchain, which in my case is somewhat broken (but I wasn't using it directly anyway). So that is why it fails. Before that, I was sort of side-stepping some of the Bazel CC toolchain tools, like the linker, and use whatever `emcmake` passed down to `cmake`, which was working for me.

Now, the way forward for me would be to fix the Emscripten-based Bazel CC toolchain, and then use the CMake toolchain file that you set here as the default. However, for that I need to fix the Emscripten toolchain first.

The reason it is broken is (a) because it assumes some absolute ptahs, because it wants me to merge that workspace into my own, which I'm not doing. I filed https://github.com/emscripten-core/emsdk/issues/696 to fix that, but I haven't gotten to it yet.

Another reason is that the Emscripten toolchain assumes there is a system Python in the `PATH`, which is not the case for me: I build Python myself, using Bazel of course :) And pass that down as `tools_deps` to the `cmake_external()` rule. (Not the most elegant solution, but hey, it works.)

Hope this helps. I'll work on making the CC Emscripten toolchain work, then this should be solved hopefully. That would be a useful thing to do anyway.

attilaolah:
Specifically, this line in the generated CMake toolchain file is causing trouble:

```
set(CMAKE_CXX_LINK_EXECUTABLE "$EXT_BUILD_ROOT/external/emsdk/emscripten_toolchain/emcc_link.sh <FLAGS> <CMAKE_CXX_LINK_FLAGS> <LINK_FLAGS> <OBJECTS> -o <TARGET> <LINK_LIBRARIES>")
```

attilaolah:
So the correct way to invoke the linker would be to cat those flags to a linker param file, and then invoke the linker with a single argument, `@path/to/param/file`. But I'm not sure whether CMake can work with linker param files.

attilaolah:
OK I managed to patch the Emscripten SDK to work for me, now also with the CMake toolchain file.

For prosperity, [here is my current patch](https://github.com/attilaolah/wasm/blob/e65ca7343d25b7344eb1af170f5ed156fa1d4b35/tools/emscripten/emsdk.patch), which I'm hoping to upstream (at least partially) at some point, but for now I'm only patching the remote Bazel repo.

PR #526, opened 2021-02-27
Reduce noise in build logs

UebelAndre:
This PR cleans up the error outputs so they're reasonably parseable. This is done by replacing the use of [ctx.actions.run_shell](https://docs.bazel.build/versions/master/skylark/lib/actions.html#run_shell) with [ctx.actions.run](https://docs.bazel.build/versions/master/skylark/lib/actions.html#run) in the core framework and instead writing the code to a file.

Windows is a bit unique in that it also builds a `.bat` script to wrap the `.sh` script and, like Bazel ([Using Bazel without Bash msys2](https://docs.bazel.build/versions/3.5.0/windows.html#using-bazel-without-bash-msys2)), it expects bash to be available on the host.

One other notable change here is that the `echo` alias is now actually `echo` and no longer `printf`. This significantly cleans up the code required for formatting generated scripts and build logs.

Finally, this PR also fixes a bug where when a build failure occurs, the error was getting swallowed up and the only issues surfaced to users were that the `static_libraries`, `shared_libraries`, `binaries`, etc... targets they requested were not created. Now users will see the full log of the build failure so they don't have to go digging for the output build log (which is not very easy to do for builds in CI).

Examples of the new errors can be seen here: https://buildkite.com/bazel/rules-foreign-cc/builds/1821#a565cc8e-7125-4bea-b8df-0fa99a024e9d

### DIFF
--- a/tools/build_defs/configure_script.bzl
+++ b/tools/build_defs/configure_script.bzl
@@ -7,8 +7,7 @@ def _pkgconfig_script(ext_build_dirs):
     script = []
     for ext_dir in ext_build_dirs:
         script.append("##increment_pkg_config_path## $$EXT_BUILD_DEPS$$/" + ext_dir.basename)
-
-    script.append("echo \"PKG_CONFIG_PATH=$${PKG_CONFIG_PATH:-}$$\"")
+        script.append("echo \"PKG_CONFIG_PATH=$${PKG_CONFIG_PATH:-}$$\"")
 
     script.append("##define_absolute_paths## $$EXT_BUILD_DEPS$$ $$EXT_BUILD_DEPS$$")
 
@@ -99,7 +98,7 @@ def create_make_script(
     script = _pkgconfig_script(ext_build_dirs)
 
     script.append("##symlink_contents_to_dir## $$EXT_BUILD_ROOT$$/{} $$BUILD_TMPDIR$$".format(root))
-    script.append("" + " && ".join(make_commands))
+    script.extend(make_commands)
     return "\n".join(script)
 
 # buildifier: disable=function-docstring
--- a/tools/build_defs/configure_script.bzl
+++ b/tools/build_defs/configure_script.bzl
@@ -5,8 +5,9 @@ load(":framework.bzl", "get_foreign_cc_dep")
 def _pkgconfig_script(ext_build_dirs):
     """Create a script fragment to configure pkg-config"""
     script = []
-    for ext_dir in ext_build_dirs:
-        script.append("##increment_pkg_config_path## $$EXT_BUILD_DEPS$$/" + ext_dir.basename)
+    if ext_build_dirs:
+        for ext_dir in ext_build_dirs:
+            script.append("##increment_pkg_config_path## $$EXT_BUILD_DEPS$$/" + ext_dir.basename)
         script.append("echo \"PKG_CONFIG_PATH=$${PKG_CONFIG_PATH:-}$$\"")
 
     script.append("##define_absolute_paths## $$EXT_BUILD_DEPS$$ $$EXT_BUILD_DEPS$$")
--- a/tools/build_defs/framework.bzl
+++ b/tools/build_defs/framework.bzl
@@ -303,7 +303,7 @@ def cc_external_rule_impl(ctx, attrs):
     if execution_os_name != "osx":
         set_cc_envs = "\n".join(["export {}=\"{}\"".format(key, cc_env[key]) for key in cc_env])
 
-    lib_header = "Bazel external C/C++ Rules. Building library '{}'\\n".format(lib_name)
+    lib_header = "Bazel external C/C++ Rules. Building library '{}'".format(lib_name)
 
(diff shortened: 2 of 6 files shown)


tools/build_defs/configure_script.bzl
```diff
@@ -7,8 +7,7 @@ def _pkgconfig_script(ext_build_dirs):
     script = []
     for ext_dir in ext_build_dirs:
         script.append("##increment_pkg_config_path## $$EXT_BUILD_DEPS$$/" + ext_dir.basename)
-
-    script.append("echo \"PKG_CONFIG_PATH=$${PKG_CONFIG_PATH:-}$$\"")
+        script.append("echo \"PKG_CONFIG_PATH=$${PKG_CONFIG_PATH:-}$$\"")
```
jsharpe, line 10: nit: Do we really need to echo PKG_CONFIG_PATH for each entry in `ext_build_dirs`?

tools/build_defs/framework.bzl
```diff
@@ -377,19 +380,40 @@ def cc_external_rule_impl(ctx, attrs):
...
         ],
-        tools = depset([wrapped_outputs.script_file] + ctx.files.data + ctx.files.tools_deps + ctx.files.additional_tools, transitive = [data[DefaultInfo].default_runfiles.files for data in data_dependencies]),
+        tools = depset(
+            [wrapped_outputs.script_file] + extra_tools + ctx.files.data + ctx.files.tools_deps + ctx.files.additional_tools,
+            transitive = [data[DefaultInfo].default_runfiles.files for data in data_dependencies],
+        ),
         # We should take the default PATH passed by Bazel, not that from cc_toolchain
         # for Windows, because the PATH under msys2 is different and that is which we need
         # for shell commands
         use_default_shell_env = execution_os_name != "osx",
```
jsharpe, line 415: nit: Can we update this comment to be consistent with the behaviour of what is actually happening here?

jsharpe (changes requested):

tools/build_defs/configure_script.bzl
UebelAndre: Ah, good point. I just didn't want it echoing unnecessarily

UebelAndre: Fixed.

tools/build_defs/framework.bzl
UebelAndre: Updated this to a `TODO` we should ideally not need this (or make it optional for users).

slsyy:
Good job. Would it be possible to display the commands being run like with a bash `set -x` flag? It is certainly good for autotools case, because of tons of possible invocations: `autogen`, `autoconf`, `autoreconf`, `configure` and `make`.

UebelAndre:
> Good job. Would it be possible to display the commands being run like with a bash `set -x` flag? It is certainly good for autotools case, because of tons of possible invocations: `autogen`, `autoconf`, `autoreconf`, `configure` and `make`.

Hmm, I thought about this as well. I opted out since adding `-x` made the logs extremely spammy and I feel that information isn't always relevant to users. I feel like users should be able to trust the rules to run what they asked it to with what they expect. The fact that that's not clear is the thing I'd like to address. It could be just that we really do need to add `-x` but I'd rather take some time to think about it more and do that as a separate PR.

jsharpe:
> > Good job. Would it be possible to display the commands being run like with a bash `set -x` flag? It is certainly good for autotools case, because of tons of possible invocations: `autogen`, `autoconf`, `autoreconf`, `configure` and `make`.
> 
> Hmm, I thought about this as well. I opted out since adding `-x` made the logs extremely spammy and I feel that information isn't always relevant to users. I feel like users should be able to trust the rules to run what they asked it to with what they expect. The fact that that's not clear is the thing I'd like to address. It could be just that we really do need to add `-x` but I'd rather take some time to think about it more and do that as a separate PR.

One option could be to use `features` to enable turning this on - we could have a `feature` that can be enabled either globally or on a per target basis to turn this on.

UebelAndre:
> One option could be to use `features` to enable turning this on - we could have a `feature` that can be enabled either globally or on a per target basis to turn this on.

That sounds pretty reasonable, but I'd still rather do that in a separate PR.

jsharpe (approved):

slsyy:
> > One option could be to use `features` to enable turning this on - we could have a `feature` that can be enabled either globally or on a per target basis to turn this on.
> 
> That sounds pretty reasonable, but I'd still rather do that in a separate PR.

If you surround `cmake && make && make install` invocation with `set -x` and then `set +x`, then you will get only relevant calls. 
It may still be spam for some, but in my opinion this type of information is quite useful.

UebelAndre:
> If you surround `cmake && make && make install` invocation with `set -x` and then `set +x`, then you will get only relevant calls.
> It may still be spam for some, but in my opinion this type of information is quite useful.

Can you open a new issue for this so I don't forget about it 😅 🙏

Issue #524, opened 2021-02-25
FR: Rules for building pure Ninja projects


UebelAndre:
Maybe there's something I'm overlooking but it seems like the only use of [Ninja](https://ninja-build.org/manual.html) in this project is through CMake. While it's great to have that functionality, it'd be great to be able to build first class Ninja projects in these rules as well.

Issue #490, opened 2021-02-03
ctx.attr.tools_deps and ctx.attr.additional_tools is redundant - rename to a single ctx.attr.tools?


jsharpe:
There is no real distinction between `ctx.attr.tools_deps` and `ctx.attr.additional_tools`. Most other rules canonically use `ctx.attr.tools` so I am proposing that we should deprecate the former two and move to the latter for tools used within the build.

whs-dot-hk:
Please, don't... first you will break api... second they are really different look at the
https://github.com/bazelbuild/rules_foreign_cc/blob/7c4490a2ee289fae3ce78e647772fd0311006cdc/tools/build_defs/framework.bzl#L531

jsharpe:
My proposal is to provide a transition path by marking the existing attributes as deprecated and providing a path to transition to using `tools` only. We could provide a buildozer command to aid the migration but these attributes don't seem widely used at least in the open source uses from what I can see from searching GitHub and the fact that until recently (PR #423 ) the code for handling the `additional_tools` attribute was broken on recent versions of bazel.

I don't follow what is different from the link you provided? Could you expand on this a bit so I can understand where you think the difference is between these two attributes.

Note that currently `ctx.attr.additional_tools` is configured for `target` rather than `exec`/`host` which I think is a bug as anything passed into `additional_tools` can't be run portably in the context of platform support. Also anything that could be passed here intended for `target` could equally be passed to `ctx.attr.additional_inputs`. I think the original intention may have been for one of the two attributes to be the public API and the other an internal 'framework' api for specific rules to add tools that they need.

attilaolah:
+1 for renaming this to `tools` and redirecting both existing attrs with a deprecation warning (but also fix wrt `cfg = "exec"`, or `cfg = "host" until #491 is fixed).

People have to pin Bazel workspaces to commits anyway, and providing support with the deprecation warning for ~6m seems reasonable, esp. given that there hasn't even been any official release tag for this repo. (Well, since the 0.0.8 which seems to have been around forever.)

whs-dot-hk:
`additional_tools` do not go into `PATH`, while `tools_deps` does

jsharpe:
> `additional_tools` do not go into `PATH`, while `tools_deps` does

Do you have a use case where this behaviour is required?

whs-dot-hk:
It agrees with the docs "Not used by the shell script part..."

UebelAndre:
[...]

PR #527, opened 2021-02-27
Added `ninja` rule for building Ninja projects

UebelAndre:
closes https://github.com/bazelbuild/rules_foreign_cc/issues/524

This PR adds a `ninja` rule for building ninja [ninja](https://ninja-build.org/) projects. This rule is virtually identical to the `make` rule.


examples/ninja_simple/code/build.ninja
```diff
@@ -0,0 +1,43 @@
+# build.ninja
+
+ninja_required_version = 1.3
+
+# Variables
+
+builddir = rules_foreign_cc_build
+rm       = rm -fr
+cxx      = clang
```
jsharpe, line 9: I don't know ninja build syntax but can this be read from the environment to pick up `CC` and `CPP_FLAGS` so that it honours the configured cc_toolchain?

tools/build_defs/ninja.bzl
```diff
@@ -0,0 +1,124 @@
...
+def _attrs():
+    """Modifies the common set of attributes used by rules_foreign_cc and sets Ninja specific attrs
+
+    Returns:
+        dict: Attributes of the `ninja` rule
+    """
+    attrs = dict(CC_EXTERNAL_RULE_ATTRIBUTES)
+
+    # Drop old vars
+    attrs.pop("make_commands")
```
jsharpe, line 90: Also drop `additional_tools` here?

examples/ninja_simple/code/build.ninja
UebelAndre: It could be but are those flags actually being set?
```
{"XCODE_VERSION_OVERRIDE": "12.4.0.12D4e", "APPLE_SDK_VERSION_OVERRIDE": "11.1", "APPLE_SDK_PLATFORM": "MacOSX"}
```
this is my `cc_env` for that action and MacOS is the only platform that doesn't use `use_default_shell_env`.

In general though, you can use an environment variable within a command by prefixing it with `$$`. I'm happy to clean up the cc_toolchain in another PR if possible and as an isolated change. It's something that should be done for the repo to ensure the rules work exactly as they're expected (and no longer be limited to what's installed on the BuildKite hosts).

tools/build_defs/ninja.bzl
UebelAndre: Dropping that would require a few more changes to make that attribute optional. I think this should be done as a separate change as part of https://github.com/bazelbuild/rules_foreign_cc/issues/490

examples/ninja_simple/code/build.ninja
jsharpe: Ok, lets fix that up in another PR - I think its common for build tools to expect the c compiler to be specified in `CC`.

tools/build_defs/ninja.bzl
jsharpe: ok

jsharpe (approved):

PR #529, opened 2021-03-01
Pass cc_toolchain files as tools instead of inputs

UebelAndre:
This is a slight cleanup and does not appear as though it'd have a negative impact on users since all the same data will be made available to the relevant [run_shell](https://docs.bazel.build/versions/master/skylark/lib/actions.html#run_shell) command


jsharpe (approved):
I'd started to implement this in #437 but its a good idea to expedite putting in this particular part of that change.

PR #532, opened 2021-03-01
Add `--sysroot` to CPPFLAGS.

rickystewart:
`jemalloc`'s `Makefile` contains this:

```
	@$(CC) -MM $(CPPFLAGS) -MT $@ -o $(@:%.$(O)=%.d) $<
```

`CFLAGS` is *not* passed in, which should usually be fine, except the
logic here only propagates a few of the arguments in `CFLAGS` over to
`CPPFLAGS`, so the `--sysroot` option isn't captured. Add a special
case just for `--sysroot`.


rickystewart:
(This fixed a problem for us at Cockroach Labs, I thought the change would be generally useful although it's very targeted. I haven't contributed code to this repo before, feel free to suggest a different way to fix the same issue :) )

jsharpe:
Just so I can understand this a bit more - where is `--sysroot` coming from? Is this set as a specific env var on a target or is it defined in the toolchain?
Also could you rebase this against origin/main rather than the cockroachdb fork so that CI will pass?

rickystewart:
> Just so I can understand this a bit more - where is --sysroot coming from? Is this set as a specific env var on a target or is it defined in the toolchain?

`--sysroot` ends up here often if you're using a custom toolchain. In our case, if you have one set with a `builtin_sysroot` via [`cc_common.create_cc_toolchain_config_info`](https://docs.bazel.build/versions/master/skylark/lib/cc_common.html#create_cc_toolchain_config_info). Example of how we do that [here](https://github.com/cockroachdb/cockroach/blob/339911c10dd1b62ec36296df2f59d2aacbebec9c/build/toolchains/dev/darwin-x86_64/cc_toolchain_config.bzl.tmpl#L133).

Also, rebased :)

foreign_cc/private/configure_script.bzl
```diff
@@ -111,6 +111,10 @@ def _get_env_vars(
     # -I flags should be put into preprocessor flags, CPPFLAGS
     # https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Preset-Output-Variables.html
     vars["CPPFLAGS"] = deps_flags.flags
+    # CPPFLAGS needs to have access to the custom sysroot, if one is set.
+    for flag in vars["CFLAGS"]:
+        if flag.startswith('--sysroot='):
+            vars["CPPFLAGS"].append(flag)
```
UebelAndre, line 117: What happens if `--sysroot` is already defined in `CPPFLAGS`?

UebelAndre (changes requested):

PR #543, opened 2021-03-08
Add missing environment to cc external rule actions

thekyz:
The environment used to execute actions is too strict, as it is actually restricting access to the toolchain during the call to configure scripts. This fixes it by updating the current env with the default one for shell actions.


thekyz:
@googlebot I signed it!

PR #548, opened 2021-03-09
RBE make

jsharpe:


foreign_cc/cmake.bzl
```diff
@@ -197,6 +195,13 @@ def _get_generator_target(ctx):
         else:
             fail("No generator set and no default is known. Please set the cmake `generator` attribute")
 
+    if "Unix Makefiles" == generator:
```
UebelAndre, line 198: I feel like this is unnecessary and you just append the generator args in the `_cmake_impl` macro. Either that or move allow this macro to also return `tool_deps` or something so there's only one where we check generator info. I think I prefer the latter

examples/third_party/libpng/BUILD.libpng.bazel
```diff
@@ -10,6 +10,7 @@ filegroup(
 cmake(
     name = "libpng",
     cache_entries = {
+        "CMAKE_POLICY_DEFAULT_CMP0074": "NEW",
```
UebelAndre, line 13: What's this for?

foreign_cc/built_tools/make_build.bzl
```diff
@@ -10,6 +10,7 @@ load("//foreign_cc/private:shell_script_helper.bzl", "os_name")
 
 def _make_tool_impl(ctx):
     script = [
+        "##script_prelude##",
```
UebelAndre, line 13: Doesn't the prelude get added in the built tools framework?

UebelAndre (changes requested):
NICE!!! The build passed! This is super awesome.

Just a few comments/questions.

UebelAndre:
Also, I think this would warrant a patch release for the CMake stuff.

foreign_cc/cmake.bzl
jsharpe: Good shout that's a better idea than this, it didn't seem right when I added it.

examples/third_party/libpng/BUILD.libpng.bazel
jsharpe: This enables use of the _ROOT variables in CMake. Unfortunately it was still failing to find zlib correctly on RBE. I suspect the non-RBE builds are finding a system install of zlib.
ZLIB_ROOT probably needs to be `$(location @zlib)` to get this build working on RBE

foreign_cc/built_tools/make_build.bzl
jsharpe: Yes you're right, this was my original change but you've added the framework bits since then so I should remove this.

UebelAndre (approved):
Awesome work!

PR #549, opened 2021-03-11
Set the MAKE environment variable for ./configure

rkolchmeyer:
Automake implements recursive make by calling the program referenced by
the $(MAKE) variable. The value of this variable is set by Autoconf,
using the value of $MAKE at configure time. To support Makefiles that
implement recursive make in this way and run in environments without
make natively installed, let's set the MAKE environment variable at
configure time.

Tested by compiling examples/third_party/bison in an environment without
'make' natively installed.

### DIFF
--- a/tools/build_defs/configure.bzl
+++ b/tools/build_defs/configure.bzl
@@ -47,6 +47,9 @@ def _create_configure_script(configureParameters):
     flags = get_flags_info(ctx)
 
     define_install_prefix = "export INSTALL_PREFIX=\"" + _get_install_prefix(ctx) + "\"\n"
+    # Automake implements recursive make by calling the $(MAKE) variable.
+    # Autoconf defines this variable from the environment.
+    define_make = "export MAKE=\"" + configureParameters.attrs.make_path + "\"\n"
 
     configure = create_configure_script(
         workspace_name = ctx.workspace_name,
@@ -73,7 +76,7 @@ def _create_configure_script(configureParameters):
         autogen_options = ctx.attr.autogen_options,
         autogen_env_vars = ctx.attr.autogen_env_vars,
     )
-    return "\n".join([define_install_prefix, configure])
+    return "\n".join([define_install_prefix, define_make, configure])
 
 def _get_install_prefix(ctx):
     if ctx.attr.install_prefix:


UebelAndre:
I believe there is already a place where environment variables are set for calls to `configure`:
https://github.com/bazelbuild/rules_foreign_cc/blob/c309ec13192f69a46aaaba39587c3d7ff684eb35/tools/build_defs/configure_script.bzl#L78

Any reason for opting out of expanding on that?

jsharpe:
> I believe there is already a place where environment variables are set for calls to `configure`:
> 
> https://github.com/bazelbuild/rules_foreign_cc/blob/c309ec13192f69a46aaaba39587c3d7ff684eb35/tools/build_defs/configure_script.bzl#L78
> 
> Any reason for opting out of expanding on that?

I guess the issue is that the value comes from the toolchain. Maybe the make toolchain should be providing a `$MAKE` Make variable?

UebelAndre:
> I guess the issue is that the value comes from the toolchain. Maybe the make toolchain should be providing a `$MAKE` Make variable?

Sounds easy enough. Is the `make` toolchain the only toolchain which should provide this environment variable? What happens in the case of `cmake` or `ninja`? Also, one other concern is whether or not this is a common enough practice to have it always set.

rkolchmeyer:
Thanks for the feedback.

> Any reason for opting out of expanding on that?

I was a bit uncertain on the best way to make the 'make' path available to that function. When I saw how the INSTALL_PREFIX is handled, it seemed simpler to handle MAKE in the same way.

> I guess the issue is that the value comes from the toolchain. Maybe the make toolchain should be providing a $MAKE Make variable?

This seems like a good idea to me, I'll pursue this a bit.

> Is the make toolchain the only toolchain which should provide this environment variable? What happens in the case of cmake or ninja?

This is a good point, I'm not sure how `cmake` implements recursive make when it generates Makefiles. I'll take a closer look into this.

> Also, one other concern is whether or not this is a common enough practice to have it always set.

It seems standard in Makefiles to expect $(MAKE) to resolve to the currently running `make` program: https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html#MAKE-Variable

Autoconf subverts that expectation a bit though. Autoconf runs its tests against ${MAKE-make} [1], and the generated Makefiles set MAKE to be whatever ${MAKE-make} resolved to in ./configure [2][3]. This all causes problems if MAKE is unset, and the proper 'make' is not in the PATH.

Examples from e2fsprogs:
[1]: https://github.com/tytso/e2fsprogs/blob/master/configure#L6849
[2]: https://github.com/tytso/e2fsprogs/blob/master/configure#L6877
[3]: https://github.com/tytso/e2fsprogs/blob/master/configure#L719

jsharpe:
Thanks for the contribution #638 has now been merged which fixes this.

rkolchmeyer:
Thanks for taking this over. I'm sorry that I never finished it.

PR #552, opened 2021-03-11
Add action_env support for MacOS

UebelAndre:
It's not clear to me why this constraint was added in https://github.com/bazelbuild/rules_foreign_cc/commit/bcb7a345f7e2f9cbb80786c80a15eef4a539ca35 but no longer appears to be needed.

I've updated how environment variables are gathered to always include the default shell env on all platforms


tools/build_defs/cc_toolchain_util.bzl
```diff
@@ -239,6 +244,15 @@ def get_env_vars(ctx):
                 user_compile_flags = copts,
             ),
         ))
+
+    if cc_toolchain:
+        cc_executable = cc_toolchain.compiler_executable
```
jsharpe, line 249: compiler_executable is deprecated and I've definitely had this return the wrong value in the case that there are different tools configured in the toolchain dependent upon the features set.
I think the correct thing to use here is `cc_common.get_tool_for_action`

jsharpe (changes requested):

tools/build_defs/cc_toolchain_util.bzl
UebelAndre: Great to know! Thanks for pointing that out 😄 

Updated

UebelAndre:
There should actually be `--action_env` support in release `0.3.0` I believe.

PR #556, opened 2021-03-11
Added "targets" API

UebelAndre:
## Overview 
This PR introduces the new "targets" API.

Instead of passing commands to the `make_commands` attribute of the core rules, users should now specify what "targets" they want to build. The issue with `make_commands` is it allows users to specify build commands which require tools that re not tracked by bazel, making the artifacts produced non hermetic. 

Every supported build system has the concept of a "target". [cmake](https://cmake.org/cmake/help/latest/manual/cmake.1.html#build-a-project) (with `--target`), [make](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html), and [ninja](https://ninja-build.org/manual.html#_the_literal_phony_literal_rule). The rules should take advantage of this fact and ensure each build is using the correct build tool from the relevant toolchain and not something arbitrary from the build host's system.

This change is most impactful for the `cmake` rule where users were previously expected to author their own build commands but never had reliable access to the cmake binary. By adopting this new API, we're able to ensure users can build with cmake and should have all the appropriate mechanisms to tell cmake where to find it's build tools.

## Deprecations

As part of this change, the `make_commands` attribute is considered deprecated but will stick around for some time to ensure users can comfortably migrate. There are a few immediate deprecations though:
- [keep_going](https://github.com/bazelbuild/rules_foreign_cc/tree/0.1.0/docs#make-keep_going) is no longer supported. Users who require this functionality but do not want to migrate to the new API can set `-k` in their existing `make_commands` settings or simply add `"make -C $$EXT_BUILD_ROOT$$/path/to/makefile/dir -k", "make -C $$EXT_BUILD_ROOT$$/path/to/makefile/dir -k install"` where `/path/to/makefile/dir` would need to be filled in manually depending on the layout of your sourcecode
- `make_commands` no longer has a `PREFIX` argument passed to it. Users can address this by setting the same information via the [env](https://github.com/bazelbuild/rules_foreign_cc/tree/0.1.0/docs#make-env) attribute.
- [cmake_options](https://github.com/bazelbuild/rules_foreign_cc/tree/0.1.0/docs#cmake_external-cmake_options) has been renamed to `generate_args` and requires that arguments be split into key-value pairs. Meaning `["-G \"Visual Studio 15 2017\" -A x64"]` translates to `["-G \"Visual Studio 15 2017\"", "-A x64"]`

## Notes

Users who depend on bash commands that are not calls to make or some build tool in `make_commands` can instead use the existing `postfixup_script` attribute. This will always be rendered into the build script right after the build commands. This is useful for builds that do not have a first class "install" target and need some custom/manual intervention to get artifacts into the correct place.

The `cmake` rule now also always sets the [CMAKE_MAKE_PROGRAM](https://cmake.org/cmake/help/latest/variable/CMAKE_MAKE_PROGRAM.html) when using known generators with toolchains defined in the rules.


UebelAndre:
https://buildkite.com/bazel/rules-foreign-cc/builds/2106 demonstrates that the previous `make_commands` api continues to work.

foreign_cc/cmake.bzl
```diff
@@ -19,24 +19,21 @@ load(
...
+    tools_deps = ctx.attr.tools_deps + cmake_data.deps
 
-    ninja_data = get_ninja_data(ctx)
-    make_commands = ctx.attr.make_commands
-    if _uses_ninja(ctx.attr.make_commands):
-        tools_deps += ninja_data.deps
-        make_commands = [command.replace("ninja", ninja_data.path) for command in make_commands]
+    # For generators using ninja, make sure it's available
+    for option in ctx.attr.cmake_options:
+        if "ninja" in option.lower():
```
jsharpe, line 32: `cmake --help` on macOS gives the following list of generators:
```
* Unix Makefiles               = Generates standard UNIX makefiles.
  Ninja                        = Generates build.ninja files.
  Ninja Multi-Config           = Generates build-<Config>.ninja files.
  Xcode                        = Generate Xcode project files.
  CodeBlocks - Ninja           = Generates CodeBlocks project files.
  CodeBlocks - Unix Makefiles  = Generates CodeBlocks project files.
  CodeLite - Ninja             = Generates CodeLite project files.
  CodeLite - Unix Makefiles    = Generates CodeLite project files.
  Eclipse CDT4 - Ninja         = Generates Eclipse CDT 4.0 project files.
  Eclipse CDT4 - Unix Makefiles= Generates Eclipse CDT 4.0 project files.
  Kate - Ninja                 = Generates Kate project files.
  Kate - Unix Makefiles        = Generates Kate project files.
  Sublime Text 2 - Ninja       = Generates Sublime Text 2 project files.
  Sublime Text 2 - Unix Makefiles
                               = Generates Sublime Text 2 project files.
                               ```
It seems that all ninja generators use the capitalisation `Ninja` so I think it's stronger to search for that as searching all of cmake_options seems like it could hit this due to matching paths or config options. Ultimately though I guess the only impact this may have is a bit of extra disk space utilisation and some extra inputs to the cache key for this action?

File: foreign_cc/cmake.bzl [hunk 2]
```diff
@@ -19,24 +19,21 @@ load(
...
-    "get_ninja_data",
-)
+load("//toolchains/native_tools:tool_access.bzl", "get_cmake_data", "get_ninja_data")
 
 def _cmake_impl(ctx):
     cmake_data = get_cmake_data(ctx)
-    make_data = get_make_data(ctx)
+    ninja_path = None
 
-    tools_deps = ctx.attr.tools_deps + cmake_data.deps + make_data.deps
```
jsharpe, line 33: Where is dependency on the make toolchain now if `Unix Makefiles` generator is picked?

foreign_cc/cmake.bzl
```diff
@@ -73,6 +62,47 @@ def _create_configure_script(configureParameters):
...
+            ctx.expand_location(arg, data)
+            for arg in ctx.attr.build_args
+        ])
+
+        # Generate commands for all the targets, ensuring there's
+        # always at least 1 call to the default target.
+        for target in ctx.attr.targets or [""]:
+            # Note that even though directory is always passed, the
+            # following arguments can take precedence.
+            cmake_commands.append("{cmake} --build {dir} --config {config} {args} {target}".format(
```
jsharpe, line 84: Do we need to explicitly choose CMP0061 here? https://cmake.org/cmake/help/latest/policy/CMP0061.html#policy:CMP0061

jsharpe, line 84: How do we ensure that `cmake --build` uses our toolchain for ninja or make? Do we need to make sure we've linked it in the path here under the canonical name? e.g. using a wrapper script to call emscripten wrapped toolchains.
Do cmake presets (https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html) help with this at all?

foreign_cc/cmake.bzl
```diff
@@ -100,42 +131,60 @@ def _get_install_prefix(ctx):
...
+                "CMake cache entries to initialize (they will be passed with `-Dkey=value`) " +
                 "Values, defined by the toolchain, will be joined with the values, passed here. " +
                 "(Toolchain values come first)"
             ),
             mandatory = False,
             default = {},
         ),
         "cmake_options": attr.string_list(
-            doc = "Other CMake options",
+            doc = "Arugments for CMake's generate command",
```
jsharpe, line 148:
```suggestion
            doc = "Arguments for CMake's generate command",
```

foreign_cc/cmake.bzl
```diff
@@ -100,42 +131,60 @@ def _get_install_prefix(ctx):
...
                 "When True, CMake crosstool file will be generated from the toolchain values, " +
-                "provided cache-entries and env_vars (some values will still be passed as -Dkey=value " +
-                "and environment variables). " +
-                "If CMAKE_TOOLCHAIN_FILE cache entry is passed, specified crosstool file will be used " +
-                "When using this option to cross-compile, it is required to specify CMAKE_SYSTEM_NAME in the " +
-                "cache_entries"
+                "provided cache-entries and env_vars (some values will still be passed as `-Dkey=value` " +
+                "and environment variables). If `CMAKE_TOOLCHAIN_FILE` cache entry is passed, " +
+                "specified crosstool file will be used When using this option to cross-compile, " +
+                "it is required to specify CMAKE_SYSTEM_NAME in the cache_entries"
```
jsharpe, line 166:
```suggestion
                "it is required to specify `CMAKE_SYSTEM_NAME` in the cache_entries"
```

foreign_cc/cmake.bzl
```diff
@@ -100,42 +131,60 @@ def _get_install_prefix(ctx):
...
+            mandatory = False,
+        ),
         "install_prefix": attr.string(
-            doc = "Relative install prefix to be passed to CMake in -DCMAKE_INSTALL_PREFIX",
+            doc = "Relative install prefix to be passed to CMake in `-DCMAKE_INSTALL_PREFIX`",
+            mandatory = False,
+        ),
+        "make_commands": attr.string_list(
+            doc = (
+                "__deprecated__: Optinal hard coded commands to replace the `cmake --build` commands. It's " +
```
jsharpe, line 185:
```suggestion
                "__deprecated__: Optional hard coded commands to replace the `cmake --build` commands. It's " +
```

foreign_cc/configure.bzl
```diff
@@ -161,7 +185,16 @@ def _attrs():
             ),
             mandatory = False,
         ),
+        "targets": attr.string_list(
+            doc = (
+                "A list of targets with in the foreign build system to produce. An empty string (`\"\"`) will result in " +
```
jsharpe, line 190:
```suggestion
                "A list of targets within the foreign build system to produce. An empty string (`\"\"`) will result in " +
```

foreign_cc/make.bzl
```diff
@@ -76,37 +82,41 @@ def _get_install_prefix(ctx):
...
         "prefix": attr.string(
             doc = (
-                "Install prefix, an absolute path. " +
-                "Passed to the GNU make via \"make install PREFIX=<value>\". " +
-                "By default, the install directory created under sandboxed execution root is used. " +
-                "Build results are copied to the Bazel's output directory, so the prefix is only important " +
-                "if it is recorded into any text files by Makefile script. " +
-                "In that case, it is important to note that rules_foreign_cc is overriding the paths under " +
-                "execution root with \"BAZEL_GEN_ROOT\" value."
+                "__deprecated__: To maintain this behavior, pass `PREFIX=<valie>` to the `args` attribute"
```
jsharpe, line 109:
```suggestion
                "__deprecated__: To maintain this behavior, pass `PREFIX=<value>` to the `args` attribute"
```

foreign_cc/make.bzl
```diff
@@ -76,37 +82,41 @@ def _get_install_prefix(ctx):
...
+            ),
+            mandatory = False,
+        ),
+        "targets": attr.string_list(
+            doc = (
+                "A list of targets with in the foreign build system to produce. An empty string (`\"\"`) will result in " +
```
jsharpe, line 115:
```suggestion
                "A list of targets within the foreign build system to produce. An empty string (`\"\"`) will result in " +
```

jsharpe (commented):
A few typos in the docs that need fixing.

Still needs a bit of work in handling how we call out to the toolchain for the selected generator as the toolchain path may not match the canonical name of the tool e.g. `ninja` could be `ninja-wrapper.sh` in the toolchain.

foreign_cc/cmake.bzl
UebelAndre: Good catch, restored the make_data here and added a TODO to expand on controlling the generator target being used. I think this should be it's own attribute so the rules can appropriately identify what toolchains to include in the build.

foreign_cc/cmake.bzl
UebelAndre: This is solved now by setting [CMAKE_MAKE_PROGRAM](https://cmake.org/cmake/help/latest/variable/CMAKE_MAKE_PROGRAM.html). [presets](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html) could totally help but would only do so in versions of cmake greater than 3.19 so I think it should be something added in the future. For now, `CMAKE_MAKE_PROGRAM` offers all the flexibility we need for the current set of toolchains since it works for the `Unix Makefile` and `Ninja` generators. Though, as stated in the documentation, there are other more nuanced generators that we'd have to solve for in another way if we wanted to have first class support for them

UebelAndre: Why would we want to ignore errors from the build commands?

foreign_cc/cmake.bzl
UebelAndre: I've actually gone ahead and added support for detecting the current generator or selecting a default. I think this is the correct thing to do.

foreign_cc/cmake.bzl
UebelAndre: I've since revised how generators are detected. Good call on this one.

foreign_cc/cmake.bzl
jsharpe: My bad, I'd inverted the logic of what CMP0061 is setting.

jsharpe: Yes, I think its fair enough for the moment to assume that for the windows generators the build tools are installed alongside the compilers.

foreign_cc/cmake.bzl
```diff
@@ -73,23 +72,136 @@ def _create_configure_script(configureParameters):
...
         tools = tools,
         flags = flags,
         install_prefix = "$$INSTALL_PREFIX$$",
         root = root,
         no_toolchain_file = no_toolchain_file,
         user_cache = dict(ctx.attr.cache_entries),
-        user_env = dict(ctx.attr.env_vars),
-        options = ctx.attr.cmake_options,
+        user_env = getattr(ctx.attr, "env_vars", {}),
+        options = attrs.generator_args,
```
jsharpe, line 128: I'm not convinced that `generator_args` is the correct name for what is essentially setting the cmake cache variables from the command line. 
I envisaged having a `generator` attr and keeping the existing `cmake_options` and then appending the `generator` value with `-G` to the values in `cmake_options`. We can present a migration warning if `-G` is passed in `cmake_options` and is a generator we explicitly support?

jsharpe: Thinking through it some more I think I'm fine with it as is. I'd confused `cmake_options` and `cache_entries` a bit in my thinking.

UebelAndre:
Anyone looking for the `PREFIX` functionality from `make` can simply pass  `"PREFIX": "$$INSTALLDIR$$",` to the new `env` attribute to maintain the previous behavior.

sam-lunt:
This PR introduced a change of behavior in how the `make_commands` are calculated. Previously, they were calculated from `attrs` but now they are determined using `ctx.attr`. This causes `boost_build` to break. `make_commands` are only set on `attrs`, not on `ctx.attr`, so the call to `./b2 install` is skipped (this is likely the cause of #641). The documentation for this function still mentions that it calls `attrs.make_commands` ([here](https://github.com/bazelbuild/rules_foreign_cc/blob/0.2.0/foreign_cc/private/framework.bzl#L290))

foreign_cc/private/framework.bzl
```diff
@@ -857,3 +866,30 @@ def _collect_libs(cc_linking):
...
+    tools_deps = []
+
+    # Early out if there are no commands set
+    if not make_commands:
+        return make_commands, tools_deps
+
+    if _uses_tool(ctx.attr.make_commands, "make"):
+        make_data = get_make_data(ctx)
+        tools_deps += make_data.deps
+        make_commands = [command.replace("make", make_data.path) for command in make_commands]
```
sam-lunt, line 881: Additionally, this change seems wrong to me. Before, the logic was:
``` python
if line == "make" or line.startswith("make "):
    make_commands.append(line.replace("make", attrs.make_path, 1))
```

This will now change something like `["make", "class_maker"]` to `["/usr/bin/make", "class_/usr/bin/maker"]`.

PR #559, opened 2021-03-14
Added a common framework for built tools

UebelAndre:
To prevent massive text dumps when building "built tool" toolchains, all built tool rules now share a common framework similar to the other rules in `//foreign_cc:defs.bzl`.

Some notable changes:
- each built tool rule has had it's `*_srcs` attribute renamed to `srcs`. So for `ninja_tool`, `ninja_srcs` must now be passed to `srcs`.
- While this isn't a change, built tool rules will not fail if the tool fails to build or install for whatever reason. https://github.com/bazelbuild/rules_foreign_cc/pull/548 is a PR for addressing this issue.

### DIFF
--- a/foreign_cc/built_tools/cmake_build.bzl
+++ b/foreign_cc/built_tools/cmake_build.bzl
@@ -1,46 +1,33 @@
 """ Rule for building CMake from sources. """
 
-load("//foreign_cc/private:detect_root.bzl", "detect_root")
-load("//foreign_cc/private:shell_script_helper.bzl", "convert_shell_script")
+load(
+    "//foreign_cc/built_tools/private:built_tools_framework.bzl",
+    "FOREIGN_CC_BUILT_TOOLS_ATTRS",
+    "FOREIGN_CC_BUILT_TOOLS_HOST_FRAGMENTS",
+    "built_tool_rule_impl",
+)
 
-def _cmake_tool(ctx):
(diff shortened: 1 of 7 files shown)


foreign_cc/built_tools/cmake_build.bzl
```diff
@@ -1,46 +1,33 @@
...
-        "##mkdirs## " + cmake.path,
-        "cd $$BUILD_TMPDIR$$",
-        "./bootstrap --prefix=install",
+        "./bootstrap --prefix=$$INSTALLDIR$$",
+        # TODO: Use make from a toolchain
+        "make",
         "make install",
-        "##copy_dir_contents_to_dir## ./install $BUILD_DIR/" + cmake.path,
-        "cd $$BUILD_DIR$$",
     ]
```
jsharpe, line 16: This could probably be rewritten quite easily as a `configure_make` target?

jsharpe (approved):

foreign_cc/built_tools/cmake_build.bzl
UebelAndre: Yeah, it can and should since it clearly requires a `make` binary. It'd be better to use a rule which set all that up. Though, I don't know if all the provider info is relevant from the configure rule, I'll have to think about that more.

PR #564, opened 2021-03-15
Generate cmake crosstool file as a declared output

jsharpe:
Move cross tool to a declared output to reduce the chances of issues with shell escaping that generating this file via a heredoc creates.


wolfd:
Ah, I was wondering why that file was heredoc'd when making #567. I think your change will make it a lot easier to do what I was doing there.

jsharpe:
> Ah, I was wondering why that file was heredoc'd when making #567. I think your change will make it a lot easier to do what I was doing there.

The reason appears to be so that shell expansion can happen for the values of the cmake cache entries. This is why buildkite is currently failing on this PR.

PR #566, opened 2021-03-15
Expand location in configure_command and autogen_command

jsharpe:


foreign_cc/configure.bzl
```diff
@@ -56,7 +56,7 @@ def _create_configure_script(configureParameters):
         user_options = ctx.attr.configure_options,
         user_vars = dict(ctx.attr.configure_env_vars),
         is_debug = is_debug_mode(ctx),
-        configure_command = ctx.attr.configure_command,
+        configure_command = ctx.expand_location(ctx.attr.configure_command, targets=ctx.attr.tools_deps + ctx.attr.additional_tools),
```
UebelAndre, line 59: Can you also add `data` here? I believe there's a discussion to deprecate these attributes since they're seemingly duplicates https://github.com/bazelbuild/rules_foreign_cc/issues/490 

But I think this should at least encapsulate all sources of data?

UebelAndre (changes requested):

foreign_cc/configure.bzl
jsharpe: Yes; I've realised though that there are a lot more places that we should be calling `ctx.expand_location`. Going to park this for a little bit until I get chance to address it more fully.

Issue #239, opened 2019-03-29
Permit __DATE__, __TIME__, etc in foreign builds


rnburn:
I'm using `configure_make` with apache and get this error
```
<command-line>:0:10: error: 'redacted' undeclared here (not in a function); did you mean 'rename'?
/root/.cache/bazel/_bazel_root/f8087e59fd95af1ae29e8fcb7ff1a3dc/sandbox/linux-sandbox/5/execroot/apache_bazel_example/external/org_apache_httpd/server/buildmark.c:21:36: note: in expansion of macro '__DATE__'
 static const char server_built[] = __DATE__ " " __TIME__;
```
because of the options
```
-Wno-builtin-macro-redefined -D__DATE__="redacted" -D__TIMESTAMP__="redacted" -D__TIME__="redacted"
```
Could those be disabled by default for these external builds?

jwnimmer-tri:
From the peanut gallery (not a maintainer): the problem seems to be that the double-quoting is being dropped.  A literal `"redacted" " " "redacted"` would have been just fine.  Shell escaping must be wrong somewhere?

UlysseM:
I doubt you'd ever want to allow these, as they would change build artifacts all the time, which prevent any kinds of caching that bazel relies on.

Try adding this:
``` configure_options = ["CFLAGS='-Dredacted=\"redacted\"'"],```

**Comment by rnburn** [comment]
Yes, it seems as if the problem is that they're not escaped correctly.

I don't know that it's important for apache - but if they're going to be fixed, I would suggest at least setting them to valid expansions in case there's code that parses them.

**Comment by tonyboxes** [comment]
I'm using `cmake_external` to compile rocksdb and get this error
```
note: in expansion of macro '__DATE__'
```
I don't find `configure_options` attrs from docs.
Could those be disabled by default for these external builds?

**Comment by adam-rocska** [comment]
I ran into the very same error trying to compile php (yeah...) : 
In case someone would like to reproduce it : https://github.com/adam-rocska/rules_php

`/Users/rocskaadam/bin/bazel build --tool_tag=ijwb:IDEA:ultimate --curses=no --color=yes --progress_in_terminal_title=no -- //php:php`

And in my local user `.bazelrc` : 
```
common --javabase @bazel_tools//tools/jdk:absolute_javabase --define=ABSOLUTE_JAVABASE=/Users/rocskaadam/lib/jdk-12.0.2.jdk --sandbox_debug --verbose_failures 

[...]

PR #567, opened 2021-03-16
Fix string escaping for cmake and configure/make

wolfd:
Refresh of #362.
Related to discussion in #239.

To be honest, this isn't the cleanest stuff in the world, but it works in my limited testing. There's a number of layers to this which I think is the cause of past confusion here. There's a lot of instances of strings getting escaped at the wrong stage, and I think it's almost pure luck that this particular combination of escape sequences actually works.

Usage of this PR that I used to "test" it:
- [Using configure_make to build Python 3.8](https://github.com/wolfd/rules_foreign_cc/commit/79a127e23b21bf289f411b4524474a573157977a)
- [CMake usage similar to how Python uses `__TIME__`](https://github.com/wolfd/rules_foreign_cc/blob/72d01448ca0a9704e4298e67424fcae616068308/examples/cmake_defines/lib_a/lib_a.cpp)

It occurs to me that actually compiling stuff in examples might be a good thing to do as part of tests, is there a reason that currently isn't the case?


wolfd:
Things are moving quite quick on `main`! Exciting! (I resolved some conflicts)

UebelAndre:
Hey, thanks for the PR! I think we just got through some of the larger structural changes that would lead to conflicts. From here on I expect more targeted changes. 

Do you think there's a quick change you could also make to one of the examples that would exercise the fix here? The examples seem to do all the heavy lifting in terms of regression testing and would prefer to have something there. 🙏

wolfd:
@googlebot I consent.

jsharpe:
@googlebot I consent.

jsharpe:
PTAL @UebelAndre

UebelAndre (changes requested):
Very nice!!

UebelAndre (approved):

UebelAndre:
(sorry about the misclick) very excited to see this go in, Thank you so much for the PR! 🙏

PR #572, opened 2021-03-17
Keep lines of scripts as a list of commands

jsharpe:
Keep the commands as a list of strings until the point the script is created.


UebelAndre (approved):
Nice! Love it!

PR #573, opened 2021-03-18
Fixed cmake build and target arguments

UebelAndre:
Targets can now be properly passed to the CMake build command and arguments are now appropriately the last thing to be passed to provide the most flexibility.

### DIFF
--- a/.bazelci/config.yaml
+++ b/.bazelci/config.yaml
@@ -48,6 +48,7 @@ tasks:
       - "-//cmake_defines/..."
       - "-//cmake_hello_world_lib/..."
       - "-//cmake_with_data/..."
+      - "-//cmake_with_target/..."
       - "-//cmake_working_dir/..."
       - "-//configure_with_bazel_transitive/..."
       - "-//make_simple/..."
--- a/examples/cmake_with_target/BUILD.bazel
+++ b/examples/cmake_with_target/BUILD.bazel
@@ -1,0 +1,34 @@
+load("@rules_foreign_cc//foreign_cc:defs.bzl", "cmake")
+
+filegroup(
+    name = "srcs",
+    srcs = glob(["src/**"]),
+    visibility = ["//visibility:public"],
+)
+
+# Building all targets with the `srcs` CMake project will fail.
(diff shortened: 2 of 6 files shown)


examples/cmake_with_target/BUILD.bazel
```diff
@@ -0,0 +1,34 @@
...
+)
+
+# Building all targets with the `srcs` CMake project will fail.
+# this example demonstrates how users can avoid unnecessary targets
+cmake(
+    name = "cmake_with_target",
+    build_args = [
+        # Pass arguments to the underlying build tool
+        "--",
+        "-j 1",
```
jsharpe, line 16: Are we better demonstrating `--parallel` here: https://stackoverflow.com/a/50883540/6857772 or https://cmake.org/cmake/help/latest/envvar/CMAKE_BUILD_PARALLEL_LEVEL.html? Feels like a more portable way of doing this.

.bazelci/config.yaml
```diff
@@ -48,6 +48,7 @@ tasks:
       - "-//cmake_defines/..."
       - "-//cmake_hello_world_lib/..."
       - "-//cmake_with_data/..."
+      - "-//cmake_with_target/..."
```
jsharpe, line 51: If we defaulted to ninja then these would work on RBE...

examples/cmake_with_target/BUILD.bazel
UebelAndre: My goal here wasn't strictly to demonstrate parallel builds. I just wanted to show how to pass arguments to the build tool
https://cmake.org/cmake/help/latest/command/build_command.html
note the `-- -i`

.bazelci/config.yaml
UebelAndre: I'd rather spend next release getting RBE to work (build the make binary)

jsharpe (approved):

examples/cmake_with_target/BUILD.bazel
UebelAndre: I've updated the example to hopefully make this clearer

PR #575, opened 2021-03-20
Set CMAKE_INSTALL_PREFIX to the actual install directory to avoid copy

UebelAndre:
There is no reason to install artifacts into some arbitrary temp directory when and copy them into the install dir when we can install them into the install dir. This removes an unnecessary copy.

This PR also removes a redundant example. All functionality is covered by cmake_hello_world_lib


jsharpe (approved):
LGTM

PR #580, opened 2021-03-22
Built ninja tools no longer carry the entire source tree

UebelAndre:
This should be a nice optimization to what was there before but function the same

### DIFF
--- a/foreign_cc/built_tools/ninja_build.bzl
+++ b/foreign_cc/built_tools/ninja_build.bzl
@@ -6,12 +6,15 @@ load(
     "FOREIGN_CC_BUILT_TOOLS_HOST_FRAGMENTS",
     "built_tool_rule_impl",
 )
+load("//foreign_cc/private:shell_script_helper.bzl", "os_name")
 
 def _ninja_tool_impl(ctx):
     script = [
         "./configure.py --bootstrap",
-        # TODO: Reduce unnecessary copys and only keep what's required
-        "##copy_dir_contents_to_dir## $$BUILD_TMPDIR$$ $$INSTALLDIR$$",
+        "mkdir $$INSTALLDIR$$/bin",
+        "cp ./ninja{} $$INSTALLDIR$$/bin/".format(
+            ".exe" if "win" in os_name(ctx) else "",
+        ),
     ]
 
     return built_tool_rule_impl(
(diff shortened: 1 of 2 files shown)


jsharpe (approved):

jsharpe (changes requested):
Looks like this breaks building make on RBE unfortunately :(

jsharpe (approved):
Only observation I have is that by putting the ninja binary into a bin directory this will possibly remove the binary from the path in the same way that make is missing - see #590. This would  impact configure scripts that are expecting to find ninja on the path, assuming that it hits the same bug, but we should probably fix that anyway.

UebelAndre:
> Only observation I have is that by putting the ninja binary into a bin directory this will possibly remove the binary from the path in the same way that make is missing - see #590. This would impact configure scripts that are expecting to find ninja on the path, assuming that it hits the same bug, but we should probably fix that anyway.

so is the path not to use `./bin` or to ensure the PATH variable is appropriately updated? I'd prefer the latter. I dislike how there's a single directory that gets a bunch of stuff copied to it. I'd rather just add the correct paths to the binaries to PATH without moving any files.

jsharpe:
> > Only observation I have is that by putting the ninja binary into a bin directory this will possibly remove the binary from the path in the same way that make is missing - see #590. This would impact configure scripts that are expecting to find ninja on the path, assuming that it hits the same bug, but we should probably fix that anyway.
> 
> so is the path not to use `./bin` or to ensure the PATH variable is appropriately updated? I'd prefer the latter. I dislike how there's a single directory that gets a bunch of stuff copied to it. I'd rather just add the correct paths to the binaries to PATH without moving any files.

AFAIK (but I might be wrong here!) the built ninja would get linked in as `$EXT_BUILD_DEPS/bin/ninja_tool/bin/ninja` and I think that the issue is that the find call to add that to the path is restricted to 1 directory deep. I agree that we should be updating the PATH variable appropriately for this case.

UebelAndre:
> AFAIK (but I might be wrong here!) the built ninja would get linked in as `$EXT_BUILD_DEPS/bin/ninja_tool/bin/ninja` and I think that the issue is that the find call to add that to the path is restricted to 1 directory deep. I agree that we should be updating the PATH variable appropriately for this case.

Yeah, I don't understand why we're doing `$EXT_BUILD_DEPS/bin/ninja_tool/bin/ninja`. I think it's fine if we're gonna muck with the PATH variable but I don't think we should be need to make this `$EXT_BUILD_DEPS/bin` directory. Though I think this goes into a larger discussion about why we don't build from the standard exec root.

PR #583, opened 2021-03-25
Use touch not cp -p to preserve timestamps

illicitonion:
If the file being copied is owned by a different user than the build
action, which may be the case for some remote execution platforms,
`cp -p` fails with `EPERM` whereas `touch -r` won't.

### DIFF
--- a/foreign_cc/private/shell_toolchain/toolchains/impl/default_commands.bzl
+++ b/foreign_cc/private/shell_toolchain/toolchains/impl/default_commands.bzl
@@ -61,7 +61,10 @@ fi
     )
 
 def copy_dir_contents_to_dir(source, target):
-    return """cp -L -p -r --no-target-directory "{}" "{}" """.format(source, target)
+    return """cp -L -r --no-target-directory "{source}" "{target}" && find {target} -type f -exec touch -r "{source}" "{{}}" \\;""".format(
+        source=source,
+        target=target,
+    )
 
 def symlink_contents_to_dir(source, target):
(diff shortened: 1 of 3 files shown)


foreign_cc/private/shell_toolchain/toolchains/impl/default_commands.bzl
```diff
@@ -61,7 +61,7 @@ fi
     )
 
 def copy_dir_contents_to_dir(source, target):
-    return """cp -L -p -r --no-target-directory "{}" "{}" """.format(source, target)
+    return """cp -L -r --no-target-directory "{source}" "{target}" && find {target} -type f -exec touch -r "{source}" "{{}}" \\;""".format(source=source, target=target)
```
UebelAndre, line 64: Can you move the parameters to the next so the line isn't so long?

```python
"".format(
    source = source,
    target = target,
)
```
🙏

UebelAndre (changes requested):
Looks good to me! Just one minor request if you don't mind 😃

foreign_cc/private/shell_toolchain/toolchains/impl/default_commands.bzl
illicitonion: Done! Thanks :)

UebelAndre (approved):
Thanks!

jsharpe:
I wonder if this resolves the issues I was seeing in #548? Just giving it a go to see if this does resolve it.

PR #588, opened 2021-03-28
Use configure_make to build cmake.

### LINKED ISSUE #584: cmake built_tool does not appear to depend on make built_tool
If I have

```
rules_foreign_cc_dependencies( 
   register_default_tools = False,
   register_preinstalled_tools = False,
   register_built_tools = True,
)
```

in my `WORKSPACE` file *without* a preinstalled version of `make` available anywhere in `$PATH`, then

```
% bazel build @rules_foreign_cc//toolchains:cmake_tool
```

fails almost immediately with
```
+ ./bootstrap --prefix=/root/.cache/bazel/_bazel_root/3788110a444fd708b5af5d589e31e877/sandbox/processwrapper-sandbox/3/execroot/core-commit/bazel-out/k8-dbg/bin/external/rules_foreign_cc/toolchains/cmake
---------------------------------------------
CMake 3.19.6, Copyright 2000-2020 Kitware, Inc. and Contributors
Found GNU toolchain
C compiler on this system is: gcc
C++ compiler on this system is: g++
---------------------------------------------
Error when bootstrapping CMake:
Cannot find appropriate Makefile processor on this system.
Please specify one using environment variable MAKE.
---------------------------------------------
Log of errors: /root/.cache/bazel/_bazel_root/3788110a444fd708b5af5d589e31e877/sandbox/processwrapper-sandbox/3/execroot/core-commit/bazel-out/k8-dbg/bin/external/rules_foreign_cc/toolchains/cmake.build_tmpdir/Bootstrap.cmk/cmake_bootstrap.log
```

I presume it should be possible to make the `cmake_tool` depend on the `make_tool`, and require building the `make_tool` first?

jsharpe:
closes #584

### DIFF
--- a/foreign_cc/built_tools/cmake_build.bzl
+++ b/foreign_cc/built_tools/cmake_build.bzl
@@ -1,35 +1,30 @@
 """ Rule for building CMake from sources. """
 
-load(
-    "//foreign_cc/built_tools/private:built_tools_framework.bzl",
-    "FOREIGN_CC_BUILT_TOOLS_ATTRS",
-    "FOREIGN_CC_BUILT_TOOLS_HOST_FRAGMENTS",
-    "built_tool_rule_impl",
-)
+load("//foreign_cc:defs.bzl", "configure_make")
 
-def _cmake_tool_impl(ctx):
-    script = [
-        "./bootstrap --prefix=$$INSTALLDIR$$",
-        # TODO: Use make from a toolchain
-        "make",
-        "make install",
-    ]
+def cmake_tool(name, srcs, **kwargs):
+    tags = ["manual"] + kwargs.pop("tags", [])
 
-    return built_tool_rule_impl(
-        ctx,
-        script,
-        ctx.actions.declare_directory("cmake"),
-        "BootstrapCMake",
+    configure_make(
+        name = "_build_{}".format(name),
+        configure_command = "bootstrap",
+        configure_options = ["--", "-DCMAKE_MAKE_PROGRAM=$$MAKE$$"],
+        # On macOS at least -DDEBUG gets set for a fastbuild
+        copts = ["-UDEBUG"],
+        lib_source = srcs,
+        out_binaries = select({
+            "@platforms//os:windows": ["cmake.exe"],
+            "//conditions:default": ["cmake"],
+        }),
+        out_static_libs = [],
+        out_shared_libs = [],
+        tags = tags,
+        **kwargs
     )
 
-cmake_tool = rule(
-    doc = "Rule for building CMake. Invokes bootstrap script and make install.",
-    attrs = FOREIGN_CC_BUILT_TOOLS_ATTRS,
-    host_fragments = FOREIGN_CC_BUILT_TOOLS_HOST_FRAGMENTS,
-    output_to_genfiles = True,
-    implementation = _cmake_tool_impl,
-    toolchains = [
-        str(Label("//foreign_cc/private/framework:shell_toolchain")),
-        "@bazel_tools//tools/cpp:toolchain_type",
-    ],
-)
+    native.filegroup(
+        name = name,
+        srcs = ["_build_{}".format(name)],
+        output_group = "gen_dir",
+        tags = tags,
+    )
--- a/examples/WORKSPACE.bazel
+++ b/examples/WORKSPACE.bazel
@@ -8,7 +8,7 @@ local_repository(
 load("@rules_foreign_cc//foreign_cc:repositories.bzl", "rules_foreign_cc_dependencies")
 
 # Don't use preinstalled tools to ensure builds are as hermetic as possible
-rules_foreign_cc_dependencies(register_preinstalled_tools = False, register_default_tools = False)
+rules_foreign_cc_dependencies(register_preinstalled_tools = False)
 
 load("@bazel_skylib//:workspace.bzl", "bazel_skylib_workspace")
 
--- a/examples/WORKSPACE.bazel
+++ b/examples/WORKSPACE.bazel
@@ -8,7 +8,7 @@ local_repository(
 load("@rules_foreign_cc//foreign_cc:repositories.bzl", "rules_foreign_cc_dependencies")
 
 # Don't use preinstalled tools to ensure builds are as hermetic as possible
-rules_foreign_cc_dependencies(register_preinstalled_tools = False)
+rules_foreign_cc_dependencies(register_preinstalled_tools = False, register_default_tools = False)
 
 load("@bazel_skylib//:workspace.bzl", "bazel_skylib_workspace")
 
--- a/foreign_cc/private/framework.bzl
+++ b/foreign_cc/private/framework.bzl
@@ -66,6 +66,11 @@ CC_EXTERNAL_RULE_ATTRIBUTES = {
         cfg = "exec",
         default = [],
     ),
+    "copts": attr.string_list(
+        doc = "Optional. Add these options to the compile flags passed to the foreign build system. The flags only take affect for compiling this target, not its dependencies.",
+        mandatory = False,
+        default = [],
+    ),
     "data": attr.label_list(
         doc = "Files needed by this rule at runtime. May list file or rule targets. Generally allows any target.",
         mandatory = False,


foreign_cc/built_tools/cmake_build.bzl
```diff
@@ -6,6 +6,9 @@ def cmake_tool(name, srcs, **kwargs):
     configure_make(
         name="_build_{}".format(name),
         lib_source=srcs,
+        out_binaries = ["cmake"],
```
UebelAndre, line 9: also `cmake.exe` on windows 😅

jsharpe: Done - however CMake really needs openssl as a dependency. There are issues with building openssl using these rules at the moment and so supporting a built from source cmake on windows is not something that will currently work.

jsharpe:
This fixes the build of cmake for linux hosts where openssl is already installed. RBE fails due to the missing openssl dependency and so there is still work to be done to get a properly hermetic cmake build working but unless someone explicitly wants this I think effort is better spent elsewhere at the moment.

foreign_cc/built_tools/cmake_build.bzl
```diff
@@ -1,35 +1,28 @@
...
-    toolchains = [
-        str(Label("//foreign_cc/private/framework:shell_toolchain")),
-        "@bazel_tools//tools/cpp:toolchain_type",
-    ],
-)
+    native.filegroup(
+        name = name,
+        srcs = ["_build_{}".format(name)],
+        output_group = "gen_dir",
+        tags = ["manual"] + kwargs.pop("tags", []),
```
UebelAndre, line 27: Doesn't `pop` remove the item from `kwargs`?

foreign_cc/private/configure_script.bzl
```diff
@@ -38,8 +38,8 @@ def create_configure_script(
         root_path = "$$BUILD_TMPDIR$$"
         configure_path = "{}/{}".format(root_path, configure_command)
 
-    script.append("##export_var## MAKE {}".format(make_path))
     script.append("##enable_tracing##")
+    script.append("##export_var## MAKE {}".format(make_path))
```
UebelAndre, line 42: Why flip these?

UebelAndre (changes requested):
Two questions 😄

foreign_cc/built_tools/cmake_build.bzl
jsharpe: Ah yes - this was a copy and paste from the target above. Will extract into a variable so that it has the correct value.

foreign_cc/private/configure_script.bzl
jsharpe: It was so that I could see what `MAKE` was being set to as the cmake build was complaining that it didn't have a command set for `CMAKE_MAKE_PROGRAM` which I'd assumed would pick up `MAKE` but it doesn't.
I can revert this change if you like?

UebelAndre: Yeah, I think this would show up in the environment dump of a failed build so if something isn't working we'd know what was used where. I know it's a small change but being able to keep the generally massive log size as small as possible is a win to me 😅

jsharpe: Reverted

jsharpe:
@UebelAndre PTAL

UebelAndre:
@jsharpe This might take a while, but could you try setting `register_default_tools = False` for the examples:
https://github.com/bazelbuild/rules_foreign_cc/blob/6d1d16d3bb1c09a5154b483de902755ce1f05746/examples/WORKSPACE.bazel#L11

As mentioned in https://github.com/bazelbuild/rules_foreign_cc/pull/588#issuecomment-874306709, the RBE build is expected to fail but it'd be nice to see other builds succeed.

jsharpe:
> @jsharpe This might take a while, but could you try setting `register_default_tools = False` for the examples:
> 
> https://github.com/bazelbuild/rules_foreign_cc/blob/6d1d16d3bb1c09a5154b483de902755ce1f05746/examples/WORKSPACE.bazel#L11
> 
> As mentioned in [#588 (comment)](https://github.com/bazelbuild/rules_foreign_cc/pull/588#issuecomment-874306709), the RBE build is expected to fail but it'd be nice to see other builds succeed.

Note it will also fail on windows as Make fails to build there and macOS as openssl is not available there either.

UebelAndre:
> > @jsharpe This might take a while, but could you try setting `register_default_tools = False` for the examples:
> 
> > 
> 
> > https://github.com/bazelbuild/rules_foreign_cc/blob/6d1d16d3bb1c09a5154b483de902755ce1f05746/examples/WORKSPACE.bazel#L11
> 
> > 
> 
> > As mentioned in [#588 (comment)](https://github.com/bazelbuild/rules_foreign_cc/pull/588#issuecomment-874306709), the RBE build is expected to fail but it'd be nice to see other builds succeed.
> 
> 
> 
> Note it will also fail on windows as Make fails to build there and macOS as openssl is not available there either.

I was never able (or patient enough) to build cmake as part of a build. Is the fact that it doesn't build on those platforms a regression? I'm assuming not

jsharpe:
> > > @jsharpe This might take a while, but could you try setting `register_default_tools = False` for the examples:
> > 
> > 
> > > 
> > 
> > 
> > > https://github.com/bazelbuild/rules_foreign_cc/blob/6d1d16d3bb1c09a5154b483de902755ce1f05746/examples/WORKSPACE.bazel#L11
> > 
> > 
> > > 
> > 
> > 
> > > As mentioned in [#588 (comment)](https://github.com/bazelbuild/rules_foreign_cc/pull/588#issuecomment-874306709), the RBE build is expected to fail but it'd be nice to see other builds succeed.
> > 
> > 
> > Note it will also fail on windows as Make fails to build there and macOS as openssl is not available there either.
> 
> I was never able (or patient enough) to build cmake as part of a build. Is the fact that it doesn't build on those platforms a regression? I'm assuming not

No its not a regression, I don't think the cmake build has ever worked properly unless the host has all the dependencies already installed.

UebelAndre:
So just for awareness. RBE fails with
```
_____ BEGIN BUILD LOGS _____
+ ./configure.py --bootstrap
/bin/sh: 1: c++: not found
bootstrapping ninja...
when running:  c++ -MMD -MT build/browse.o -MF build/browse.o.d -g -Wall -Wextra -Wno-deprecated -Wno-missing-field-initializers -Wno-unused-parameter -fno-rtti -fno-exceptions -fvisibility=hidden -pipe '-DNINJA_PYTHON="python"' -O2 -DNDEBUG -DUSE_PPOLL -DNINJA_HAVE_BROWSE -I. -c ./src/browse.cc -o build/browse.o
  File "./configure.py", line 470, in <module>
    objs += cxx('browse', order_only=built('browse_py.h'))
  File "./configure.py", line 287, in cxx
    return n.build(built(name + objext), 'cxx', src(name + '.cc'), **kwargs)
  File "./configure.py", line 169, in build
    self._run_command(self._expand(cmd, local_vars))
  File "./configure.py", line 194, in _run_command
    subprocess.check_call(cmdline, shell=True)
  File "/usr/lib/python2.7/subprocess.py", line 541, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command 'c++ -MMD -MT build/browse.o -MF build/browse.o.d -g -Wall -Wextra -Wno-deprecated -Wno-missing-field-initializers -Wno-unused-parameter -fno-rtti -fno-exceptions -fvisibility=hidden -pipe '-DNINJA_PYTHON="python"' -O2 -DNDEBUG -DUSE_PPOLL -DNINJA_HAVE_BROWSE -I. -c ./src/browse.cc -o build/browse.o' returned non-zero exit status 127
_____ END BUILD LOGS _____
```

and windows failed with
```
_____ BEGIN BUILD LOGS _____
+ ./configure.py --bootstrap
bootstrapping ninja...
  File "./configure.py", line 329, in <module>
    if platform.msvc_needs_fs():
  File "./configure.py", line 89, in msvc_needs_fs
    stderr=subprocess.PIPE)
  File "C:\python3\lib\subprocess.py", line 729, in __init__
    restore_signals, start_new_session)
  File "C:\python3\lib\subprocess.py", line 1017, in _execute_child
    startupinfo)
FileNotFoundError: [WinError 2] The system cannot find the file specified
_____ END BUILD LOGS _____
```

Which just seems that https://github.com/bazelbuild/rules_foreign_cc/pull/640 is needed before we can get to the root cause here. But given that we have similar experience in bootstrapping cmake, I'm happy to support an effort to move this forward.

UebelAndre (approved):

PR #591, opened 2021-03-30
Optimize methods in detect_root.bzl

djmarcin:
detect_root and filter_containing_dirs_from_inputs showed up on profiles of our builds as taking a long time (300+ ms). This change uses faster algorithms for detecting the root from a set of source files and eliminating duplicates.

The root is the shortest common dirname among all files. There is an implicit assumption that there is at least one file at the root, otherwise in a tree like the following, the wrong root will be detected, but the existing code also makes this assumption and it seems like it's required.
```
foo/bar/bar.c
foo/baz/baz.c
(Root should be foo, but will be detected as foo/bar)
```

filter_containing_dirs_from_inputs is made faster by avoiding the sort of a potentially large number of files by doing two passes, first to collect a set of directories that contain files, and second to create a list filtering out any directories in the set. (i.e. changes from n log n -> 2n)

This reduces the total time in analysis from 432ms to 134ms for our repository.

### DIFF
--- a/foreign_cc/private/detect_root.bzl
+++ b/foreign_cc/private/detect_root.bzl
@@ -15,54 +15,16 @@ def detect_root(source):
     if len(sources) == 0:
         return ""
 
-    root = None
-    level = -1
-
     # find topmost directory
+    root = None
     for file in sources:
-        file_level = _get_level(file.path)
-
-        # If there is no level set or the current file's level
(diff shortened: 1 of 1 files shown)


UebelAndre (approved):
Awesome change! Thanks for opening this PR 😄 🙏

PR #596, opened 2021-03-31
configure_in_place make a copy of the source code

jsharpe:
When using configure_in_place make a copy of the source code rather than symlinking as the configure script may modify the input source tree which shouldn't be modified.


jsharpe:
> Is there an example (test) we could add for this? Seems like some behavior that could regress.

Yes I'll come up with an example that modifies the inputs in its configure script.

examples/configure_modify_input_source/simple_lib/configure
```diff
@@ -0,0 +1,15 @@
...
+
+# Slightly contrived example of a tool
+# that resolves the symlinks to the real file
+# and attempts to modify it
+# (Using perl here for replacement to avoid difference in in-place
+# arguments on macOS sed)
+# Note that perl -i would just replace the symlink
+# if used directly
+
+perl -i -pe 's/42/0/' $(readlink src/simple.c)
```
UebelAndre, line 11: `perl`?! 😅 Does this work in RBE?

jsharpe: It appears to! I'd default to sed but then I'd need to branch on os in the configure script to pass the correct arguments (or live with the extra `src/simple.c-e` file that would be created on macOS).
Unfortunately this example is still not breaking (it does with `--spawn_strategy=standalone` though)

PR #597, opened 2021-03-31
CMake build in execroot

jsharpe:


UebelAndre (changes requested):
Nice change! I'm wondering though if we can update an example to build on this information? Maybe something that gets used from `$(execpath :foo)` that would break if the build wasn't running from the exec root?

UebelAndre (approved):
After discussing with @jsharpe the tests I was looking for would be something that uses relative paths to define the location of some tool or data but would not work in this case because all that's really happening is the `cd` is being delegated to the build tools.

The change to build in the build root is something that should be done in a separate PR.

If this change is rebased and the conflict resolved I think it's good to go.

PR #598, opened 2021-03-31
Pass empty suffix to sed on macOS

jsharpe:
sed on macOS requires a suffix when using -i. Current usage will create temporary files with -e suffixes so pass an empty suffix to avoid creating the backup files on macOS.


UebelAndre:
Interesting, how'd you come across this one? Curious why this doesn't currently fail in any existing tests.

jsharpe:
> Interesting, how'd you come across this one? Curious why this doesn't currently fail in any existing tests.

It won't fail to run it just creates extra temporary files that aren't required. Came across it because I was looking how the shell toolchain handled the difference in sed arguments for in-place to create an example for #596

PR #599, opened 2021-04-02
Addressed buildifier defects

UebelAndre:


jsharpe:
Could you fix up the python script that generates the bzl file so that it adds the trailing commas this PR adds?

UebelAndre:
Yeah, easier said then done. I could update it to run buildifier after generating but that assumes the user has it installed. I'll see if there's an easier way to correctly write to the file.

PR #600, opened 2021-04-02
Removed all deprecated attributes

UebelAndre:
In doing this, I noticed that there were a bunch of attributes that were actually unused:
https://github.com/bazelbuild/rules_foreign_cc/blob/0.0.9/tools/build_defs/configure_script.bzl#L63-L82

Making their removal safer. This should greatly cleanup the docs and make the usability of the rules clearer


foreign_cc/make.bzl
```diff
@@ -52,64 +51,27 @@ def _create_make_script(configureParameters):
...
         ),
-        "keep_going": attr.bool(
-            doc = (
-                "__deprecated__: To maintain this behavior, pass `-k` to the `args` attribute " +
-                "when not using the `make_commands` attribute."
-            ),
-            mandatory = False,
-            default = True,
-        ),
-        "make_commands": attr.string_list(
```
jsharpe, line 96: How would we support needing different arguments per target? I can imagine some Makefiles existing that need different args passed to install than build but we're removing the escape hatch that would support that here. I don't have a concrete example of this though. Thoughts?

jsharpe (commented):
Just a question about whether we're ready to remove make_commands for the make rule. Otherwise this looks good.

foreign_cc/make.bzl
UebelAndre: I was thinking this would not be supported. I've yet to come across a case where two libraries are built by the same target but require different args (at least looking back at examples within these rules and other projects I'm familiar with). If this is not the case then the current API would definitely need a refactor. Do you do this in any of your projects?

jsharpe: No, I don't do this; was more of a hypothetical use case question. Just wanted to highlight that we are removing an extension point but I guess if someone wants this sort of behaviour they can create a starlark rule that makes use of the framework themselves to support this.

jsharpe (approved):
LGTM once the conflict is resolved.

foreign_cc/make.bzl
UebelAndre: I'm happy to revisit this and add support if there's a large desire for it. But I feel in the common case that users can split their targets up with very little effort or negative impact if they want to use different arguments.

Issue #593, opened 2021-03-30
Absolutize tool paths for configure_make for hermetic toolchains


jheaff1:
Hi,

I’m using a custom cc toolchain with wrappers around the tools checked into my repo, eg //toolchains/gcc_x86_64_linux_musl/wrappers/gcc.

It appears that the `configure_make` rule sets the CXX variable as “ toolchains/gcc_x86_64_linux_musl/wrappers/gcc”. This path is not accessible by the directory in which the build actions take place, therefore the build action fails.

`cmake` rules, however, are successful with my cc toolchain.

After digging around the rules_foreign_cc repo, [cmake_script.bzl](https://github.com/bazelbuild/rules_foreign_cc/blob/bae11c9a508953c1ec3a564102630878737de99d/foreign_cc/private/cmake_script.bzl#L244) sets the `force` flag of `absolutize_path_in_str` to `True`.  However, [`get_env_vars`](https://github.com/bazelbuild/rules_foreign_cc/blob/bae11c9a508953c1ec3a564102630878737de99d/foreign_cc/private/make_script.bzl#L44) used by [configure_script.bzl](https://github.com/bazelbuild/rules_foreign_cc/blob/main/foreign_cc/private/configure_script.bzl) does not set the `force` flag of `absolutize_path_in_str`.

Would I be correct in stating that a solution to the issue is to just set the `force` flag to True in the `_absolutize` function of [make_script.bzl](https://github.com/bazelbuild/rules_foreign_cc/blob/bae11c9a508953c1ec3a564102630878737de99d/foreign_cc/private/make_script.bzl#L163).

UebelAndre:
> Would I be correct in stating that a solution to the issue is to just set the `force` flag to True in the `_absolutize` function of [make_script.bzl](https://github.com/bazelbuild/rules_foreign_cc/blob/bae11c9a508953c1ec3a564102630878737de99d/foreign_cc/private/make_script.bzl#L163).

That's a good question, I think that'd be the correct thing to do. Given that we don't use the default exec root (instead we make a new directory and `cd` into it), all traditional `$(execroot foo)` paths will be wrong so we'd always want to prepend the path to the original exec root. I can't think of why we wouldn't always do this. Perhaps this creates an issue with caching? I don't believe so though... Give it a shot and open a PR? We'll see if the build reveal anything.

jheaff1:
[...]

PR #601, opened 2021-04-02
absolutize tool paths for (configure_)make (#593)

jheaff1:
This change is necessary if a hermetic cc toolchain is being used
(e.g. gcc checked into the workspace repository)

Testing: Successfully built zlib using configure_make and a custom
hermetic cc toolchain.

### DIFF
--- a/foreign_cc/private/make_script.bzl
+++ b/foreign_cc/private/make_script.bzl
@@ -48,7 +48,7 @@ def get_env_vars(
         user_vars,
         deps,
         inputs):
-    vars = _get_configure_variables(tools, flags, user_vars)
+    vars = _get_configure_variables(workspace_name, tools, flags, user_vars)
     deps_flags = _define_deps_flags(deps, inputs)
 
     if "LDFLAGS" in vars.keys():
@@ -124,7 +124,7 @@ _CONFIGURE_TOOLS = {
     # missing: cxx_linker_executable
 }
 
-def _get_configure_variables(tools, flags, user_env_vars):
+def _get_configure_variables(workspace_name, tools, flags, user_env_vars):
     vars = {}
 
     for flag in _CONFIGURE_FLAGS:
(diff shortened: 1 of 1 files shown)


UebelAndre:
Seems good to me, can you think of any way we could prevent this from regressing in future changes? Would hate to break your use of the rules again in the future 😅

UebelAndre (approved):

jheaff1:
> Seems good to me, can you think of any way we could prevent this from regressing in future changes? Would hate to break your use of the rules again in the future 😅

That's a good question. Does rules_foreign_cc have any automated tests as part of its CI pipeline? If so, could there be some tests that use a hermetic cc toolchain?

UebelAndre:
> That's a good question. Does rules_foreign_cc have any automated tests as part of its CI pipeline? If so, could there be some tests that use a hermetic cc toolchain?

The examples are all run as part of CI: https://github.com/bazelbuild/rules_foreign_cc/tree/main/examples

Perhaps something could be added there? Or if there's a easy way to go setup a hermetic toolchain under some condition we could enable that in a dedicated job for building these targets.

jheaff1:
@UebelAndre That could definitely be achieved. Here are a few projects that define cc toolchains:

* https://github.com/grailbio/bazel-toolchain
* https://github.com/silvergasp/bazel-embedded
* https://github.com/d-asnaghi/bazel-arm-none-eabi
* 

One of these projects could be used by rules_foreign_cc. Or perhaps it’s better that rules_foreign_cc define its own toolchain, to prevent a cyclic dependency of e.g bazel-embedded depends on rules_foreign_cc which depends on bazel-embedded

foreign_cc/private/make_script.bzl
```diff
@@ -142,7 +142,8 @@ def _get_configure_variables(tools, flags, user_env_vars):
     for tool in _CONFIGURE_TOOLS:
         tool_value = getattr(tools, _CONFIGURE_TOOLS[tool])
         if tool_value:
-            tools_dict[tool] = [tool_value]
+            # Force absolutize of tool paths, which may relative to the workspace dir (hermetic toolchains) be provided in project repositories (i.e hermetic toolchains)
```
jheaff1, line 145: Need to fix the wording of this comment

PR #611, opened 2021-04-18
Windows builds now use powershell wrappers

UebelAndre:
This enables windows builds to always run using a native shell and allows for `use_default_shell_env` to be set to `False` to make builds more hermetic.


UebelAndre:
https://github.com/bazelbuild/rules_foreign_cc/issues/655 is the better thing to do

PR #612, opened 2021-04-19
Refactored framework to avoid polymorphism complexity

UebelAndre:
The only public facing change here is that [rules_foreign_cc_dependencies](https://github.com/bazelbuild/rules_foreign_cc/tree/98c88de5e0f57dd2a896dd595469ccd6bf97ceab/docs#rules_foreign_cc_dependencies) no longer accepts the [additional_shell_toolchain_mappings](https://github.com/bazelbuild/rules_foreign_cc/tree/98c88de5e0f57dd2a896dd595469ccd6bf97ceab/docs#rules_foreign_cc_dependencies-additional_shell_toolchain_mappings) and [additional_shell_toolchain_package](https://github.com/bazelbuild/rules_foreign_cc/tree/98c88de5e0f57dd2a896dd595469ccd6bf97ceab/docs#rules_foreign_cc_dependencies-additional_shell_toolchain_package) parameters. The toolchains and their implementation are private to the `foreign_cc` package and they're not currently in a state where users should rely on them. Instead, we should focus on improving them and allowing them to remain invisible to users of these rules.

Aside from reorganizing some code, the changes here simplify the repository rule (previously `generate_overloads`, now `framework_toolchain_repository`) to define the toolchains we register. These toolchains have a provider which contains all they commands for the associated `command_srcs` target given to the repository rule. There's now no longer any polymorphism logic to deal with as a result.


foreign_cc/private/framework/toolchains/default_commands.bzl
```diff
@@ -93,15 +94,15 @@ local target="$2"
 mkdir -p "$target"
 if [[ -f "$1" ]]; then
   ln -s -f -t "$target" "$1"
-elif [[ -L "$1" ]]; then
-  local actual=$(readlink "$1")
-  ##symlink_to_dir## "$actual" "$target"
+elif [[ -L "$1" && ! -d "$1" ]]; then
+  cp "$1" "$2"
```
jsharpe, line 98: Should this be `cp -a`?

UebelAndre: This is exactly what the linux commands do.
https://github.com/bazelbuild/rules_foreign_cc/blob/98c88de5e0f57dd2a896dd595469ccd6bf97ceab/foreign_cc/private/shell_toolchain/toolchains/impl/linux_commands.bzl#L97

should that be `cp -a`?

UebelAndre: the default (stale linux) toolchain was instead deleted.

foreign_cc/private/framework/BUILD.bazel
```diff
@@ -1,7 +1,13 @@
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
 
+toolchain_type(
+    name = "toolchain_type",
+    visibility = ["//visibility:public"],
+)
+
```
jsharpe, line 7: I feel there is some value in having the word 'shell' in the Label for this as this Label will be visible to users even though its in the private directory e.g. with toolchain resolution debugging turned on.

UebelAndre: I feel like the package name would be representative of the toolchain type. But I can rename the actual type

UebelAndre:
@jsharpe ready for another review

jsharpe (approved):

Issue #607, opened 2021-04-13
FR: Add powershell toolchain


UebelAndre:
Currently, the rules do not work on windows without bash and even with system with bash, we must include [use_default_shell_env](https://docs.bazel.build/versions/4.0.0/skylark/lib/actions.html#run) to make sure the wrapper scripts have all the MSYS2 tools in the path at all times. This is a bit undesirable since I'd like to build targets without this to make my build more hermetic and reduce the risk of non-deterministic outputs from these other build systems.

Instead of trying to do some `PATH` manipulation, we could introduce a [Powershell](https://docs.microsoft.com/en-us/powershell/) (`pwsh_toolchain`) similar to the existing [shell_toolchain](https://github.com/bazelbuild/rules_foreign_cc/tree/2d98051813a49f5deb6f469ac12e198a0a2f11c5/foreign_cc/private/shell_toolchain) for native windows support.

This should alleviate the need for `use_default_shell_env` and allow us to introduce new attributes to lock down the build actions.

UebelAndre:
https://github.com/bazelbuild/rules_foreign_cc/issues/655 is the more correct thing to do here.

PR #614, opened 2021-04-19
Replaced uses of `$var ` with `$$var$$`

UebelAndre:
This moves more functionality into the framework so things like https://github.com/bazelbuild/rules_foreign_cc/issues/607 are more doable. Instead of relying on bash, we should use the foreign_cc framework to define functionality.

### DIFF
--- a/foreign_cc/private/cmake_script.bzl
+++ b/foreign_cc/private/cmake_script.bzl
@@ -86,7 +86,7 @@ def create_cmake_script(
     # Add definitions for all environment variables
     script = set_env_vars
 
-    directory = "$EXT_BUILD_ROOT/" + root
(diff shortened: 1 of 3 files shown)


jsharpe (commented):
Seems like this PR only changes env vars to use the `$$var$$` syntax rather than using `##export_var##` as described in the title of the PR?

UebelAndre:
> Seems like this PR only changes env vars to use the `$$var$$` syntax rather than using `##export_var##` as described in the title of the PR?

Good catch, I should have reviewed this before opening it. I definitely have a branch with a bunch of those changes though 😅

jsharpe (approved):

Issue #582, opened 2021-03-25
Can't `bazel run` things generated as `out_binaries`


illicitonion:
I put together a `rules_foreign_cc` rule for building a binary which builds with configure+make, and wanted to be able to `bazel run` it.

This was tricky - a bunch of the things I wanted to do didn't work:

First I tried to just `bazel run` the target, but the target isn't labelled as `executable`. Rules can't be conditionally marked `executable` based on e.g. whether they have `out_binaries` (or exactly one `out_binary`). We could do something like "always run the first `out_binary` if there are any, and generate a dummy executable if not", but that seems kind of sketchy, and ideally the `run`-ability could be decided at analysis time.

Next I tried to copy the generated file in a `genrule` marked `executable=True`, but the generated file didn't have an associated label (for `//target:my` with an `out_binary` named `binary`, I couldn't write `//target:my/bin/binary`).

So I ended up writing a wrapper rule, `runnable_binary` which depends on the `rules_foreign_cc` target, looks up the file in the provider and does the copy, and is marked as `executable`.

I can upstream `runnable_binary` if folks would be interested, but I'd be interested in whether there's a better approach here...

jsharpe:
This is similar to what https://github.com/bazelbuild/rules_foreign_cc/pull/417 is attempting to address. It's a tricky problem as we don't know what the runtime dependencies are between the files generated in the `gen_dir` output group is as this is not made explicit to bazel. The standard way of using runfiles in bazel to provide any data files needed by a runnable target generated by a `rules_foreign_cc` rule is unlikely to be correct and so a general solution probably needs to wrap the declared `out_binary`'s somehow and ensure the correct relative file layout is setup. 

I have wondered if we could approach this with aspects and generate the set of `executable` targets that way?

UebelAndre:
I like the idea of a wrapper rule for defining runnable targets

KennyS:
Hey @illicitonion, I just ran into the same issue, could you share "runnable_binary"? If the repo isn't shareable a pastebin or gist link would be great

illicitonion:
Absolutely; this is what we're currently using:

```starlark
def _impl(ctx):
    name = ctx.attr.binary_name or ctx.attr.name
    input = ctx.attr.generated_by[OutputGroupInfo][name].to_list()[0]
    output = ctx.actions.declare_file(name)

    # We're not allowed to specify a file generated by another rule as our executable, so copy it.
    #   
[...]

PR #622, opened 2021-04-27
Provide `out_data_dirs` attribute (#419)

### LINKED ISSUE #419: Support "data" outputs
I'm currently attempting to use rules_foreign_cc to build python (😅), but I'm running into the issue that python depends on non-library files that are output as part of the build process into `lib/pythonX.Y`. As far as I can tell there's no way to support this with the current rules.

It seems like supporting something like `data` outputs would make sense, however in this case we'd also need to support data directories, since I'm not sure (or don't want to enumerate) of all the the file names in the target definition. Unfortunately there's also a lot of files that aren't reproducible in the `lib/pythonX.Y` directory, so maybe we'll need to manually specify them anyways.

I'm open to implementing this myself, but would like to know if this seems like a good direction before I go down that path.

There's also a whole other host of things that I'm sure will make building python non-trivial, but at least this data file requirement seems like it could be common.

jheaff1:
When the `python` executable, built using `configure_make`, is invoked it requires the files within the generated `lib` folder to be available on the file system. 

This commit provides an `out_data_dirs` attribute, which declares the given directories as outputs to the bazel build action. By specifying the generated `lib` dir as one of the `output_data_dirs`, the `lib` directory is accessible when the `python` executable is invoked.

This facilitates hermetic python toolchains, built using the awesome `rules_foreign_cc` repo

This resolves #419


examples/third_party/python/BUILD.bazel
```diff
@@ -0,0 +1,37 @@
...
+
+py_runtime(
+    name = "py3_runtime",
+    files = ["@python"],
+    interpreter = "@python//:python3_fg",
+    python_version = "PY3",
+)
+
+py_runtime_pair(
+    name = "my_py_runtime_pair",
```
UebelAndre, line 18:
```suggestion
    name = "py_runtime_pair",
```

examples/third_party/python/BUILD.bazel
```diff
@@ -0,0 +1,37 @@
...
+    py2_runtime = None,
+    py3_runtime = ":py3_runtime",
+)
+
+toolchain(
+    name = "my_toolchain",
```
UebelAndre, line 24:
```suggestion
    name = "python_toolchain",
```

examples/third_party/python/BUILD.bazel
```diff
@@ -0,0 +1,37 @@
...
+    target_compatible_with = [
+        "@platforms//os:linux",
+        "@platforms//cpu:x86_64",
+    ],
+    toolchain = ":my_py_runtime_pair",
```
UebelAndre, line 29:
```suggestion
    toolchain = ":py_runtime_pair",
```

examples/third_party/python/BUILD.bazel
```diff
@@ -0,0 +1,37 @@
...
+    [
+        "BUILD.python.bazel",
+    ],
+    visibility = ["//visibility:public"],
+)
+
+py_runtime(
+    name = "py3_runtime",
+    files = ["@python"],
+    interpreter = "@python//:python3_fg",
```
UebelAndre, line 13:
```suggestion
    interpreter = "@python//:python3_bin",
```

examples/third_party/python/BUILD.python.bazel
```diff
@@ -0,0 +1,35 @@
...
+    ],
+    out_data_dirs = ["lib"],
+    deps = [
+        "@openssl",
+        "@zlib",
+    ]
+)
+
+filegroup(
+    name = "python3_fg",
```
UebelAndre, line 32:
```suggestion
    name = "python3_bin",
```

UebelAndre (changes requested):
I'm not sure how I feel about accepting directories as the output group here. Given that Bazel can't make virtually any determinism guarantees about these build systems, I think it's better for the rules to request specific outputs. Though, I'd imagine there'd be a massive list of files here if you needed to be explicit. Maybe @jsharpe would have some thoughts there

Also this target doesn't seem to be hooked up. The examples are in a less than ideal state and require you to add things to a test_suite to get them tested:
https://github.com/bazelbuild/rules_foreign_cc/blob/98c88de5e0f57dd2a896dd595469ccd6bf97ceab/examples/third_party/BUILD.bazel#L3-L59

Can you add the new target there?

jheaff1:
@UebelAndre I have actioned your change requests.

I agree that this is against bazel best practice, although the same can be said for how includes are handled in this repository as they too are added using `ctx.actions.declare_directory`.

The `out_data_dirs` attribute is optional for those that choose to utilise it can greatly benefit from the feature, accepting the lack of full determinism

examples/third_party/BUILD.bazel
```diff
@@ -31,6 +31,7 @@ test_suite(
         "//libssh2:libssh2_build_test",
         "//openssl:openssl_build_test",
         "//pcre:pcre_build_test",
+        "//python:python_test",
```
UebelAndre, line 34: Could you also add this to other platforms? It'd be ideal if this could be added to all test_suites.

examples/third_party/python/BUILD.python.bazel
```diff
@@ -0,0 +1,35 @@
...
+    ],
+    env = {"CFLAGS": "-Dredacted='\\\"redacted\\\"'"},
+    # rules_foreign_cc defaults the install_prefix to "python". This conflicts with the "python" executable that is generated.
+    install_prefix = "py_install",
+    lib_source = ":all_srcs",
+    out_binaries = [
+        "python3",
+        "pip3",
+    ],
+    out_data_dirs = ["lib"],
```
UebelAndre, line 24: What's the content of `lib`? I do feel this should be an explicit list.

UebelAndre (changes requested):

examples/third_party/BUILD.bazel
jheaff1: I'm happy to add it to macos, although i do not own a mac so cannot resolve any issues. I'll add the test to the macos test suite and see what happens

examples/third_party/python/BUILD.python.bazel
jheaff1: There are 8624 files in the lib folder. `python` needs access to these files at runtime

examples/third_party/BUILD.bazel
UebelAndre: Is there any reason this wouldn't be expected to work on windows as well?

examples/third_party/python/BUILD.python.bazel
UebelAndre: owch, yeah, I guess directories are a necessity.

examples/third_party/BUILD.bazel
jheaff1: I suppose not, although there aren’t any tests at all for Windows at the moment. Happy to add it to the test suite to give it a go!

examples/third_party/python/BUILD.bazel
```diff
@@ -0,0 +1,34 @@
...
+exports_files(
+    [
+        "BUILD.python.bazel",
+    ],
+    visibility = ["//visibility:public"],
+)
+
+py_runtime(
+    name = "py3_runtime",
+    files = ["@python"],
```
jsharpe, line 12: Could you instead use the `gen_dir` output group to create a filegroup with all of these files? If using `gen_dir` includes too much you can always use the `postfix_script` attribute to remove any files you don't need from the install directory?
If that works then this would avoid the need to add an extra attribute.

examples/third_party/BUILD.bazel
jsharpe: I'm guessing on windows you'll run into problems with compiling openssl and also the `.exe` extensions on the binaries will need some select statements adding.

examples/third_party/python/BUILD.bazel
UebelAndre: This also seems related to https://github.com/bazelbuild/rules_foreign_cc/issues/582

examples/WORKSPACE
```diff
@@ -49,3 +49,5 @@ examples_third_party_repositories()
 load("@bazel_toolchains//rules:rbe_repo.bzl", "rbe_autoconfig")
 
 rbe_autoconfig(name = "buildkite_config")
+
+register_toolchains("@rules_foreign_cc_examples_third_party//python:python_toolchain")
```
UebelAndre, line 53: Can you move this into `./examples/third_party/python/python_repositories.bzl`?

UebelAndre (changes requested):

examples/third_party/BUILD.bazel
UebelAndre: Ah yeah, the OpenSSL build doesn't work :(

examples/third_party/python/BUILD.bazel
jheaff1: @jsharpe  I'll look into using `gen_dir`

examples/WORKSPACE
jheaff1: I tried that but `buildifier` complained about the fact that the functions used in `examples/third_party/repositories.bzl` do not have a `name` argument. This is because `buildifier` considers functions as "macros" if `native` is being used, and when `register_toolchains` was moved to `python_respositories.bzl` I had to invoke it as `native.register_toolchains`.

UebelAndre: If you move that into a macro, you need to use the `native` module, and as for the buildifier isssue, you can annotate that away:
https://github.com/bazelbuild/rules_foreign_cc/blob/98c88de5e0f57dd2a896dd595469ccd6bf97ceab/foreign_cc/repositories.bzl#L11-L12

That should fix the issue and keep the functionality for the python tests primarily in one place.

examples/third_party/python/BUILD.bazel
jheaff1: @jsharpe  I tried adding the following `filegroup` to `Build.python3.bazel`

```
filegroup(
    name = "gen_dir",
    srcs = [":python3"],
    output_group = "gen_dir",
)
```

I then set the `files` attribute of the `py_runtime` to `["@python3//:gen_dir]`. Sadly the files inside the `lib` directory generated by the build of python were not included in the `runfiles` for the `python3_test`. As such, the invocation of `python3` failed.

Perhaps I am doing something wrong?

examples/WORKSPACE
jheaff1: done

examples/third_party/repositories.bzl
```diff
@@ -11,8 +11,11 @@ load("//libpng:libpng_repositories.bzl", "libpng_repositories")
 load("//libssh2:libssh2_repositories.bzl", "libssh2_repositories")
 load("//openssl:openssl_repositories.bzl", "openssl_repositories")
 load("//pcre:pcre_repositories.bzl", "pcre_repositories")
+load("//python:python_repositories.bzl", "python_repositories")
 load("//zlib:zlib_repositories.bzl", "zlib_repositories")
 
+# This buildifier suppression is required as "native.register_toolchains" is used in python_repositories
```
UebelAndre, line 17:
```suggestion
```

I don't think it's necessary to comment about this. I think the `unnamed-macro` defect is a bit too open and gets flagged on things that shouldn't be an issue. But maybe that's just my own lack of understanding.

examples/third_party/WORKSPACE.bazel
```diff
@@ -17,3 +17,5 @@ local_repository(
 load("//:repositories.bzl", "repositories")
 
 repositories()
+
+register_toolchains("//python:python_toolchain")
```
UebelAndre, line 21: This should not be needed now that the toolchain is registered in the `.bzl` file.

UebelAndre (changes requested):

examples/third_party/WORKSPACE.bazel
jheaff1: oops, i meant to remove that. I've now removed ot

examples/third_party/python/BUILD.python3.bazel
```diff
@@ -0,0 +1,34 @@
+load("@rules_foreign_cc//foreign_cc:defs.bzl", "configure_make")
+
+package(default_visibility = ["//visibility:public"])
+
+filegroup(
+    name = "all_srcs",
+    srcs = glob(["**"]),
+)
+
+configure_make(
```
UebelAndre, line 10: For the current set of MacOS errors, maybe it's possible to build only a specific target to avoid whatever is requiring svn? I don't want macos to block this PR but I also feel that all our examples should be cross platform.

UebelAndre (changes requested):

examples/third_party/python/BUILD.python3.bazel
jheaff1: I couldn't figure out a way to avoid the need for svnversion. I could do one of two things:

1. Run only the test for python3 on macos and omit the python2 test
2. build subversion from source and provide the binary to the python2 build. Subversion requires apr and apr-util, which will also need building from source.

I'm happy to give option 2 a go if you think it's best

UebelAndre: I'd be fine ignoring the py2 tests. To me the point of building python using these rules isn't to run the tests, it's just to produce a usable binary. I think in nearly all cases you'd ignore the tests and just build the targets you want maybe after running the tests against your execution or target environment once before. It might be cool to build subversion but I'm worried about that going down a rabbit hole where there are many more dependencies to build.

jheaff1: I have added some commits to this PR to build subversion so that the tests can run on macos

examples/third_party/apr/apr_repositories.bzl
```diff
@@ -0,0 +1,18 @@
...
+    maybe(
+        http_archive,
+        name = "apr",
+        build_file = Label("//apr:BUILD.apr.bazel"),
+        # Apply patch from https://bz.apache.org/bugzilla/show_bug.cgi?id=50146
+        patches = [ Label("//apr:macos.patch") ],
+        sha256 = "48e9dbf45ae3fdc7b491259ffb6ccf7d63049ffacbc1c0977cced095e4c2d5a2",
+        strip_prefix = "apr-1.7.0",
+        urls = [
+            "https://www-eu.apache.org/dist//apr/apr-1.7.0.tar.gz",
```
UebelAndre, line 16: Is the double `//` intended?

jheaff1: I copied that from the old httpd example from this repository. I’ll change it to a single `/` and see if it works.

FYI - I am currently pushing temporary commits while trying to get the macOS tests to work. Once that’s all done I’ll tidy up the commit history.

UebelAndre: All good. We squash all changes before merging so not a huge deal. I just saw you made some changes and wanted to help out with the build failures 😄

jheaff1:
@UebelAndre All tests are happy on linux, but not on macos. I have resolved the issues with building apr on macos but sadly apr-util fails to build as the `wrapped_clang`  "C preprocessor" fails sanity checks. Any ideas? I feel like I am tantalisingly close to it all working on macos

UebelAndre:
> @UebelAndre All tests are happy on linux, but not on macos. I have resolved the issues with building apr on macos but sadly apr-util fails to build as the `wrapped_clang` "C preprocessor" fails sanity checks. Any ideas? I feel like I am tantalisingly close to it all working on macos

I'll see if I can get any more logs out of this later tonight.

jheaff1:
@UebelAndre  I tried using older versions of apr-util and the same error occurred. Must be an issue due to using bazel on modern macOS.

UebelAndre:
> @UebelAndre I tried using older versions of apr-util and the same error occurred. Must be an issue due to using bazel on modern macOS.

Hey, sorry but I wasn't able to find the time to check this out yet. Not to step on your efforts but were you able to confirm that python builds if you disable the tests?

jheaff1:
> > @UebelAndre I tried using older versions of apr-util and the same error occurred. Must be an issue due to using bazel on modern macOS.
> 
> Hey, sorry but I wasn't able to find the time to check this out yet. Not to step on your efforts but were you able to confirm that python builds if you disable the tests?

No worries. 

The build/tests for all third party libraries are successful on Linux. 
Python2 cannot currently build on macOS as it requires subversion, which requires apr-util, which is currently failing.
Subversion also requires sqlite, which is failing on RBE due to outdated autotools installed.

In my efforts to get pytho2 building for macOS, I have resolved issues in building apr and apr-util (for Linux at least). This may be beneficial for many people. If you are still ok with CI not running the python tests for macOS, then I could change this pull request to contribute the following to rules_foreign_cc:

1. build apr (on Linux, RBE and MacOS in CI)
2. build apr-util (On Linux and RBE in CI)
3. build python2 and python3 (on Linux and RBE in CI).

Thoughts?

examples/third_party/python/BUILD.python3.bazel
```diff
@@ -0,0 +1,34 @@
...
+    name = "python3",
+    configure_options = [
+        "--with-openssl=$EXT_BUILD_DEPS/openssl",
+        "--with-zlib=$EXT_BUILD_DEPS/zlib",
+    ],
+    env = {"CFLAGS": "-Dredacted='\\\"redacted\\\"'"},
```
jsharpe, line 16: This workaround shouldn't be needed anymore after #567 was merged - can you try removing this from here and the other locations where this is used?

jheaff1: Sadly the issue still occurs, ie _DATE__ and __TIME__ are defined as `redacted` rather than `"redcted"`, therefore this workaround is still required

jheaff1:
Not sure how I managed it but I seem to have broken the CI. It is complaining about dangling symbolic links. Is it possible to do a `bazel clean`?

UebelAndre:
> Not sure how I managed it but I seem to have broken the CI. It is complaining about dangling symbolic links. Is it possible to do a `bazel clean`?

Odd, I thought CI ran in docker containers that got cleaned up at the end of every build....

UebelAndre:
I'll have more time over the weekend to look at this. Sorry about the slow support here 😅

jheaff1:
> I'll have more time over the weekend to look at this. Sorry about the slow support here 😅

No need to apologise!   Any support at all is much appreciated

jheaff1:
> > Not sure how I managed it but I seem to have broken the CI. It is complaining about dangling symbolic links. Is it possible to do a `bazel clean`?
> 
> Odd, I thought CI ran in docker containers that got cleaned up at the end of every build....

Perhaps some files are being persisted in the remote cache?
https://github.com/bazelbuild/rules_foreign_cc/pull/622#issuecomment-834547106

A similar issue here, seems like removing the entire bazel folder from `~/.cache` worked. Not sure how to do that for the remote cache

UebelAndre:
So with the changes in https://github.com/bazelbuild/rules_foreign_cc/pull/635 and the following patch:
```diff
diff --git a/examples/third_party/python/BUILD.python2.bazel b/examples/third_party/python/BUILD.python2.bazel
index e691e2f..a9122a5 100644
--- a/examples/third_party/python/BUILD.python2.bazel
+++ b/examples/third_party/python/BUILD.python2.bazel
@@ -22,12 +22,13 @@ configure_make(
     configure_options = [
         "--with-openssl=$EXT_BUILD_DEPS/openssl",
         "--with-zlib=$EXT_BUILD_DEPS/zlib",
+        "--enable-optimizations",
     ],
     # rules_foreign_cc defaults the install_prefix to "python". This conflicts with the "python" executable that is generated.
     install_prefix = "py_install",
     lib_source = ":all_srcs",
     out_binaries = [
-        "python2",
+        "python2.7",
     ],
     out_data_dirs = ["lib"],
     tools_deps = select({
@@ -47,5 +48,5 @@ configure_make(
 filegroup(
     name = "python2_bin",
     srcs = [":python2"],
-    output_group = "python2",
+    output_group = "python2.7",
 )
diff --git a/examples/third_party/python/BUILD.python3.bazel b/examples/third_party/python/BUILD.python3.bazel
index 8a0c0f4..0f41807 100644
--- a/examples/third_party/python/BUILD.python3.bazel
+++ b/examples/third_party/python/BUILD.python3.bazel
@@ -16,13 +16,14 @@ configure_make(
     configure_options = [
         "--with-openssl=$EXT_BUILD_DEPS/openssl",
         "--with-zlib=$EXT_BUILD_DEPS/zlib",
+        "--enable-optimizations",
     ],
     env = {"CFLAGS": "-Dredacted='\\\"redacted\\\"'"},
     # rules_foreign_cc defaults the install_prefix to "python". This conflicts with the "python" executable that is generated.
     install_prefix = "py_install",
     lib_source = ":all_srcs",
     out_binaries = [
-        "python3",
+        "python3.9",
     ],
     out_data_dirs = ["lib"],
     deps = [
@@ -38,5 +39,5 @@ configure_make(
 filegroup(
     name = "python3_bin",
     srcs = [":python3"],
-    output_group = "python3",
+    output_group = "python3.9",
 )
diff --git a/examples/third_party/subversion/BUILD.subversion.bazel b/examples/third_party/subversion/BUILD.subversion.bazel
index 1916031..3487cfd 100644
--- a/examples/third_party/subversion/BUILD.subversion.bazel
+++ b/examples/third_party/subversion/BUILD.subversion.bazel
@@ -14,7 +14,10 @@ CONFIGURE_ENV_VARS = {
 configure_make(
     name = "subversion",
     configure_env_vars = select({
-        "@platforms//os:macos": dicts.add({"AR": ""}, CONFIGURE_ENV_VARS),
+        "@platforms//os:macos": dicts.add(
+            {"AR": ""},
+            CONFIGURE_ENV_VARS,
+        ),
         "//conditions:default": CONFIGURE_ENV_VARS,
     }),
     configure_options = [
@@ -25,12 +28,17 @@ configure_make(
         "--with-zlib=$EXT_BUILD_DEPS/zlib",
         "--with-lz4=internal",
         "--with-utf8proc=internal",
+        "--enable-experimental-libtool",
     ],
     lib_source = ":all_srcs",
     out_binaries = [
         "svn",
         "svnversion",
     ],
+    targets = [
+        "bin",
+        "install",
+    ],
     visibility = ["//visibility:public"],
     deps = [
         "@apr",

```

Everything built successfully.

Though, I have some reservations about the changes. subversion and python take a __very__ long time to build. I would like to find a way to disable testing by default and have some functional test using [rules_python](https://github.com/bazelbuild/rules_python) that shows the binaries able to successful run python code and import some built-in modules. I think the additional things being built here are still valuable and worth adding though. But I would also like to find the minimal amount of work required to produce commonly consumed artifacts. I'm worried that if we add large builds and test suites that testing changes will take too long for people to reasonably contribute changes to the project (which matters a lot to me since I can only work on this project in my free time 😄).

jheaff1:
@UebelAndre Thank you so much for the changes, I have applied them and the "standalone" macos tests now pass!

Sadly the non-standalone macos build still fails for the same reason as before, as does the RBE build.

You make a good point about long build times may discourage developers from contributing to this repo in the future. On the other hand, having tests that ensure that Python can build using this repo would be a great regression test.

That's a decision for the owner of this repo, really, I'm not sure who that it is.

As the build of Python on Linux does not require subversion (and therefore does not require sqlite, apr and apr-util), perhaps the CI could test Python only under Linux? That said, apr and apr-util are common dependencies for projects and it would be great to ensure this repo can continue to build them successfully. As such, how about all platforms build apr and apr-util in CI, and only Linux builds Python in CI?

jheaff1:
> Can you rebase and squash so I can get a better sense of what the changes are?

Done :+1:

UebelAndre:
Yeah... there still seems to be some kind of cache poisoning :(

```
checking how to run the C preprocessor... /private/var/tmp/_bazel_buildkite/8afef7f0076038b0098d9c7983f77386/sandbox/darwin-sandbox/55/execroot/rules_foreign_cc_examples/external/local_config_cc/wrapped_clang -E
configure: error: in `/private/var/tmp/_bazel_buildkite/b9053bf78f240a4cb7ec4b8f3e375e3f/sandbox/darwin-sandbox/65/execroot/rules_foreign_cc_examples/bazel-out/host/bin/external/apr_util/apr_util.build_tmpdir':
configure: error: C preprocessor "/private/var/tmp/_bazel_buildkite/8afef7f0076038b0098d9c7983f77386/sandbox/darwin-sandbox/55/execroot/rules_foreign_cc_examples/external/local_config_cc/wrapped_clang -E" fails sanity check
```

I'll dig deeper into this.

jheaff1:
@UebelAndre  Your patch seemed to fix the standalone examples macos build but the non-standalone examples macos build still fails for the same reason. The patch also broke the main macos build :(

UebelAndre:
Ok, the core issue is that paths to binaries in your CC toolchain are being hard coded into `libtool` and we don't include it when doing replacements because we only care about paths to things in the build directory which doesn't include `external/local_config_cc`. I'll need to figure out how to best make this info available for performing replaces 😅

examples/third_party/subversion/BUILD.subversion.bazel
```diff
@@ -0,0 +1,49 @@
...
+    }),
+    configure_options = [
+        "--enable-all-static",
+        "--without-boost",
+        "--with-apr=$EXT_BUILD_DEPS/apr",
+        "--with-apr-util=$EXT_BUILD_DEPS/apr_util",
+        "--with-zlib=$EXT_BUILD_DEPS/zlib",
+        "--with-lz4=internal",
+        "--with-utf8proc=internal",
+        "--enable-experimental-libtool",
```
UebelAndre, line 31: Can you try removing this line?

UebelAndre: Applying this just to see what happens 😅

UebelAndre: I believe that solved it.

jheaff1:
@UebelAndre Thanks to all your help, I had all CI builds/tests passing! apr_util successfully built on macos, which lead to a successful build of subversion and python2 for MacOS. 

As apr_util built successfully on MacOS, I then re-enabled the apr_util build test and CI now fails, it appears to be cache poisoning again, i.e. it is complaining about the `wrapped_clang` executable failing sanity checks, even though I previously had apr_util building successfully on MacOS.

UebelAndre:
> @UebelAndre Thanks to all your help, I had all CI builds/tests passing! apr_util successfully built on macos, which lead to a successful build of subversion and python2 for MacOS.
> 
> As apr_util built successfully on MacOS, I then re-enabled the apr_util build test and CI now fails, it appears to be cache poisoning again, i.e. it is complaining about the `wrapped_clang` executable failing sanity checks, even though I previously had apr_util building successfully on MacOS. The only change I made was re-enable the apr_util build test.
> 
> Here is the successful build - https://buildkite.com/bazel/rules-foreign-cc/builds/2624

It seems apr_util is still attempting to use the `libtool` generated by the `apr` build? Can you try building `apr_util` with CMake? I think that'd avoid this issue.

jheaff1:
@UebelAndre  I’m happy to give it a go but the APR [documentation](https://github.com/apache/apr/blob/trunk/README.cmake), building of APR (util) using CMake is intended for Windows only.

It’s strange that apr-util built fine on MacOS then failed after simply re-enabling the build test. Would a clean of the remote cache resolve the issue?

UebelAndre:
> @UebelAndre I’m happy to give it a go but the APR [documentation](https://github.com/apache/apr/blob/trunk/README.cmake), building of APR (util) using CMake is intended for Windows only.

Dang



> It’s strange that apr-util built fine on MacOS then failed after simply re-enabling the build test. Would a clean of the remote cache resolve the issue?

The issue is that `apr` produces a `libtool` script which contains absolute paths to compilers and linkers used. This file is then cached and passed to the next target, `apr_util`. In a single/clean build, everything works fine because the sandbox is the same. However, if `apr` should be cached but `apr_util` isn't, the cached `libtool` script will be used and contain paths to an old sandbox. This is the error we're seeing and how the cache is being poisoned. `apr_util` needs to either not build with the `apr` libtool and instead use one provided by bazel (ideal) or the generated libtool needs to have the sandbox paths replaced. This can probably be done by expanding the section here:
https://github.com/bazelbuild/rules_foreign_cc/blob/71ebe2b3fdd01042e173049cfc716abf20388868/foreign_cc/private/framework.bzl#L357-L361

to include something like
```python
"##replace_absolute_root_paths## $$INSTALLDIR$$,
```

where `replace_absolute_root_paths` is a new command similar to:
https://github.com/bazelbuild/rules_foreign_cc/blob/71ebe2b3fdd01042e173049cfc716abf20388868/foreign_cc/private/framework/toolchains/macos_commands.bzl#L222-L227

except it'd more generally find paths to the sandbox and replace it.

I believe either of those solutions would solve the issues we're seeing here.

jheaff1:
Ah ok - how would I configure the `apr_util` build to use the `apr` libtool provided by Bazel?

UebelAndre:
> Ah ok - how would I configure the `apr_util` build to use the `apr` libtool provided by Bazel?

I'm not sure. It might not be possible without patching the library.

examples/third_party/apr_util/BUILD.apr_util.bazel
```diff
@@ -0,0 +1,22 @@
...
+
+filegroup(
+    name = "all_srcs",
+    srcs = glob(["**"]),
+)
+
+configure_make(
+    name = "apr_util",
+    configure_env_vars = select({
+        "@platforms//os:macos": {"AR": ""},
```
UebelAndre, line 11:
```suggestion
        "@platforms//os:macos": {},
```

UebelAndre:
Ok, I think the following diff might do it
```diff
diff --git a/foreign_cc/private/framework.bzl b/foreign_cc/private/framework.bzl
index 179a721..fed6e39 100644
--- a/foreign_cc/private/framework.bzl
+++ b/foreign_cc/private/framework.bzl
@@ -363,6 +363,7 @@ def cc_external_rule_impl(ctx, attrs):
         # for the results which are in $INSTALLDIR (with placeholder)
         "##replace_absolute_paths## $$INSTALLDIR$$ $$BUILD_TMPDIR$$",
         "##replace_absolute_paths## $$INSTALLDIR$$ $$EXT_BUILD_DEPS$$",
+        "##replace_sandbox_paths## $$INSTALLDIR$$ $$EXT_BUILD_ROOT$$",
         installdir_copy.script,
         "cd $$EXT_BUILD_ROOT$$",
     ]
diff --git a/foreign_cc/private/framework/toolchains/commands.bzl b/foreign_cc/private/framework/toolchains/commands.bzl
index 691a5c6..d3a9d36 100644
--- a/foreign_cc/private/framework/toolchains/commands.bzl
+++ b/foreign_cc/private/framework/toolchains/commands.bzl
@@ -90,6 +90,13 @@ PLATFORM_COMMANDS = {
         ],
         doc = "Defines a function with 'text' as the function body.",
     ),
+    "define_sandbox_paths": _command_info(
+        arguments = [
+            _argument_info(name = "dir_", data_type = type(""), doc = "Directory where to replace"),
+            _argument_info(name = "abs_path", data_type = type(""), doc = "Absolute path value"),
+        ],
+        doc = "Replaces sandbox path placeholder inside 'dir_' with a provided value 'abs_path'",
+    ),
     "disable_tracing": _command_info(
         arguments = [],
         doc = "Disable script tracing. eg: `set +x`",
@@ -177,6 +184,13 @@ PLATFORM_COMMANDS = {
         ],
         doc = "Replaces all occurrences of 'from_' to 'to_' recursively in the directory 'dir'.",
     ),
+    "replace_sandbox_paths": _command_info(
+        arguments = [
+            _argument_info(name = "dir_", data_type = type(""), doc = "Directory where to replace"),
+            _argument_info(name = "abs_path", data_type = type(""), doc = "Absolute path value"),
+        ],
+        doc = "Replaces sandbox path 'abs_path' inside 'dir_' with a placeholder value",
+    ),
     "script_extension": _command_info(
         arguments = [],
         doc = "Return the extension for the current set of commands (`.sh` for bash, `.ps1` for powershell)",
diff --git a/foreign_cc/private/framework/toolchains/linux_commands.bzl b/foreign_cc/private/framework/toolchains/linux_commands.bzl
index e711502..b40357a 100644
--- a/foreign_cc/private/framework/toolchains/linux_commands.bzl
+++ b/foreign_cc/private/framework/toolchains/linux_commands.bzl
@@ -2,8 +2,6 @@
 
 load(":commands.bzl", "FunctionAndCallInfo")
 
-_REPLACE_VALUE = "\\${EXT_BUILD_DEPS}"
-
 def shebang():
     return "#!/usr/bin/env bash"
 
@@ -191,13 +189,27 @@ fi""".format(dir_ = dir_)
 def define_absolute_paths(dir_, abs_path):
     return "##replace_in_files## {dir_} {REPLACE_VALUE} {abs_path}".format(
         dir_ = dir_,
-        REPLACE_VALUE = _REPLACE_VALUE,
+        REPLACE_VALUE = "\\${EXT_BUILD_DEPS}",
         abs_path = abs_path,
     )
 
 def replace_absolute_paths(dir_, abs_path):
     return "##replace_in_files## {dir_} {abs_path} {REPLACE_VALUE}".format(
         dir_ = dir_,
-        REPLACE_VALUE = _REPLACE_VALUE,
+        REPLACE_VALUE = "\\${EXT_BUILD_DEPS}",
+        abs_path = abs_path,
+    )
+
+def define_sandbox_paths(dir_, abs_path):
+    return "##replace_in_files## {dir_} {abs_path} {REPLACE_VALUE}".format(
+        dir_ = dir_,
+        REPLACE_VALUE = "\\${EXT_BUILD_ROOT}",
+        abs_path = abs_path,
+    )
+
+def replace_sandbox_paths(dir_, abs_path):
+    return "##replace_in_files## {dir_} {abs_path} {REPLACE_VALUE}".format(
+        dir_ = dir_,
+        REPLACE_VALUE = "\\${EXT_BUILD_ROOT}",
         abs_path = abs_path,
     )
diff --git a/foreign_cc/private/framework/toolchains/macos_commands.bzl b/foreign_cc/private/framework/toolchains/macos_commands.bzl
index 74aa547..22dbff5 100644
--- a/foreign_cc/private/framework/toolchains/macos_commands.bzl
+++ b/foreign_cc/private/framework/toolchains/macos_commands.bzl
@@ -2,8 +2,6 @@
 
 load(":commands.bzl", "FunctionAndCallInfo")
 
-_REPLACE_VALUE = "\\${EXT_BUILD_DEPS}"
-
 def shebang():
     return "#!/usr/bin/env bash"
 
@@ -215,13 +213,27 @@ fi""".format(dir_ = dir_)
 def define_absolute_paths(dir_, abs_path):
     return "##replace_in_files## {dir_} {REPLACE_VALUE} {abs_path}".format(
         dir_ = dir_,
-        REPLACE_VALUE = _REPLACE_VALUE,
+        REPLACE_VALUE = "\\${EXT_BUILD_DEPS}",
         abs_path = abs_path,
     )
 
 def replace_absolute_paths(dir_, abs_path):
     return "##replace_in_files## {dir_} {abs_path} {REPLACE_VALUE}".format(
         dir_ = dir_,
-        REPLACE_VALUE = _REPLACE_VALUE,
+        REPLACE_VALUE = "\\${EXT_BUILD_DEPS}",
+        abs_path = abs_path,
+    )
+
+def define_sandbox_paths(dir_, abs_path):
+    return "##replace_in_files## {dir_} {abs_path} {REPLACE_VALUE}".format(
+        dir_ = dir_,
+        REPLACE_VALUE = "\\${EXT_BUILD_ROOT}",
+        abs_path = abs_path,
+    )
+
+def replace_sandbox_paths(dir_, abs_path):
+    return "##replace_in_files## {dir_} {abs_path} {REPLACE_VALUE}".format(
+        dir_ = dir_,
+        REPLACE_VALUE = "\\${EXT_BUILD_ROOT}",
         abs_path = abs_path,
     )
diff --git a/foreign_cc/private/framework/toolchains/windows_commands.bzl b/foreign_cc/private/framework/toolchains/windows_commands.bzl
index 9967fd9..3de8d80 100644
--- a/foreign_cc/private/framework/toolchains/windows_commands.bzl
+++ b/foreign_cc/private/framework/toolchains/windows_commands.bzl
@@ -2,8 +2,6 @@
 
 load(":commands.bzl", "FunctionAndCallInfo")
 
-_REPLACE_VALUE = "\\${EXT_BUILD_DEPS}"
-
 def shebang():
     return "#!/usr/bin/env bash"
 
@@ -202,13 +200,27 @@ fi""".format(dir_ = dir_)
 def define_absolute_paths(dir_, abs_path):
     return "##replace_in_files## {dir_} {REPLACE_VALUE} {abs_path}".format(
         dir_ = dir_,
-        REPLACE_VALUE = _REPLACE_VALUE,
+        REPLACE_VALUE = "\\${EXT_BUILD_DEPS}",
         abs_path = abs_path,
     )
 
 def replace_absolute_paths(dir_, abs_path):
     return "##replace_in_files## {dir_} {abs_path} {REPLACE_VALUE}".format(
         dir_ = dir_,
-        REPLACE_VALUE = _REPLACE_VALUE,
+        REPLACE_VALUE = "\\${EXT_BUILD_DEPS}",
+        abs_path = abs_path,
+    )
+
+def define_sandbox_paths(dir_, abs_path):
+    return "##replace_in_files## {dir_} {abs_path} {REPLACE_VALUE}".format(
+        dir_ = dir_,
+        REPLACE_VALUE = "\\${EXT_BUILD_ROOT}",
+        abs_path = abs_path,
+    )
+
+def replace_sandbox_paths(dir_, abs_path):
+    return "##replace_in_files## {dir_} {abs_path} {REPLACE_VALUE}".format(
+        dir_ = dir_,
+        REPLACE_VALUE = "\\${EXT_BUILD_ROOT}",
         abs_path = abs_path,
     )
diff --git a/foreign_cc/private/make_script.bzl b/foreign_cc/private/make_script.bzl
index 0c503de..a8a6744 100644
--- a/foreign_cc/private/make_script.bzl
+++ b/foreign_cc/private/make_script.bzl
@@ -25,6 +25,9 @@ def pkgconfig_script(ext_build_dirs):
             script.append("##increment_pkg_config_path## $$EXT_BUILD_DEPS$$/" + ext_dir.basename)
         script.append("echo \"PKG_CONFIG_PATH=$${PKG_CONFIG_PATH:-}$$\"")
 
-    script.append("##define_absolute_paths## $$EXT_BUILD_DEPS$$ $$EXT_BUILD_DEPS$$")
+    script.extend([
+        "##define_absolute_paths## $$EXT_BUILD_DEPS$$ $$EXT_BUILD_DEPS$$",
+        "##define_sandbox_paths## $$EXT_BUILD_ROOT$$ $$EXT_BUILD_ROOT$$",
+    ])
 
     return script

```

UebelAndre:
If that last diff does the trick then I can get that in in a separate PR.

jheaff1:
> If that last diff does the trick then I can get that in in a separate PR.

I've applied the patch and now there are "permission denied" errors in the CI build :(

UebelAndre:
> I've applied the patch and now there are "permission denied" errors in the CI build :(

Do you know if the caching issues have been resolved though? I can look into the permissions issues later. Just short on time 😅

jheaff1:
Hmm its hard to tell if the caching issues have been fixed due to the "permission denied" errors. I can successfully build apr_util locally (on Linux).

UebelAndre:
> Hmm its hard to tell if the caching issues have been fixed due to the "permission denied" errors. It seems that the build is not progressing as far as it did before, to the point when the caching issues occurred. I can successfully build apr_util locally (on Linux).

My hope is that my last patch fixed the caching issues and I'd be happy to have replaced caching issues with something a bit more obvious like permissions issues. These failures are also limited to the standalone builds which feels like a good sign.

jheaff1:
> > Hmm its hard to tell if the caching issues have been fixed due to the "permission denied" errors. It seems that the build is not progressing as far as it did before, to the point when the caching issues occurred. I can successfully build apr_util locally (on Linux).
> 
> My hope is that my last patch fixed the caching issues and I'd be happy to have replaced caching issues with something a bit more obvious like permissions issues. These failures are also limited to the standalone builds which feels like a good sign.

That is very true indeed, all tests pass on the non-standalone

jheaff1:
@UebelAndre The failure is happening due to the `sed` command in the `<platform>_commands.bzl` files:

https://github.com/bazelbuild/rules_foreign_cc/blob/71ebe2b3fdd01042e173049cfc716abf20388868/foreign_cc/private/framework/toolchains/linux_commands.bzl#L74

If we consider `apr_util`, in the non-standalone case one of the files that is being modified by `sed` is `apr_util.ext_build_deps/apr/build-1/libtool`. The contents of `apr_util.ext_build_deps` have read-write permissions, therefore `sed` has permission to write a temporary file while performing the in-place (`sed -i`) change.

In the standalone case, the offending file is `copy_apr/apr/build-1/libtool`. The contents of `copy_apr` are read-only, I presume because the `copy_apr` directory is a declared output of the `apr` build and bazel sets build outputs as read-only. Due to the read-only permissions, `sed` fails to write a temporary file while performing the in-place change.

I tried adding a `chmod +w` command before the `sed` command to set the required directory as writable, which resolved the issue for me locally but caused other issues in CI. Plus it seems a bit hacky to overrule the file permissions specified by bazel.

jheaff1:
After adding the `chmod +w` command before the `sed` command, the CI build failed as the target of a `sed` command did not exist. After wrapping the `sed` commands in an `if` statement that checks for the existence of the files, most of the CI build is now successful. Not sure I like the solution though, it feels quite hacky.

The only build errors now is that subversion and python aren't building on MacOS standalone. Sadly the CI logs don't indicate the issue

jheaff1:
Interestingly, I re-ran the build, with no code changes, and now everything passes...

jheaff1:
@UebelAndre @jsharpe Would it be acceptable to disable the MacOS tests for this PR, so that the PR could be merged and Linux users can benefit from hermetic Python toolchains built from source?

UebelAndre:
Hey, sorry about the delay, that might not be the worst thing... I need to figure out if the issue is systemic and will never work in standalone or if it's just this specific example. I don't think I'd want to merge something that just didn't work in one of the environments we're claiming to support. I'm sorry this PR has been sitting for so long 😞 I'm trying to find a block of time where I can sit down and try and solve the wonky permissions issues once and for all. I'll sync up with @jsharpe to see what he thinks though.

UebelAndre:
As a side note, I came across a potential solution for python from @attilaolah 
https://github.com/attilaolah/wasm/blob/a41f98b7206a0ed388a400af56038ae5c8a2d2ac/lib/python/BUILD.bazel#L17-L37

But haven't looked to closely into it. While I do think the `out_data_dirs` attribute is valuable, maybe we can get a building version of a usable python interpreter in the examples somewhere to satisfy some of the need here while I go hunt for the time to dig into this some more 😅

attilaolah:
Building Python for `--cpu=k8` works quite nicely actually, the trouble I was having is running it with the correct `PYTHONHOME` env var. The way I do it currently is via a shell script:

https://github.com/attilaolah/wasm/blob/a41f98b7206a0ed388a400af56038ae5c8a2d2ac/tools/python/python.sh

However, it comes with an unfortunate side effect of having to compile Python twice, and I'm suspecting it is because once it gets compiled with `cfg = "exec"`, and once as a data dep of my `sh_binary()` which somehow causes it to have a different compilation environment.

jheaff1:
Thanks @UebelAndre and @attilaolah.

Taking inspiration from @attilaolah's  [wasm](https://github.com/attilaolah/wasm) repo, I can successfully define a working python toolchain within bazel using rules_foreign_cc to build python **without** the need for the out_dirs_attribute. 

All that was required was a wrapper script around the python executable.

I'm happy to submit another PR to include this in the examples, although the primary issue with this PR was around building python for MacOS, specifically building the transitive dependencies of Python for MacOS (subversion, sqlite, apr-util, apr). As such, the same issues will affect this potential new PR.

Let me know your thoughts

UebelAndre:
> Thanks @UebelAndre and @attilaolah.
> 
> Taking inspiration from @attilaolah's [wasm](https://github.com/attilaolah/wasm) repo, I can successfully define a working python toolchain within bazel using rules_foreign_cc to build python **without** the need for the out_dirs_attribute.

That's great to hear!!!!

> I'm happy to submit another PR to include this in the examples, although the primary issue with this PR was around building python for MacOS, specifically building the transitive dependencies of Python for MacOS (subversion, sqlite, apr-util, apr). As such, the same issues will affect this potential new PR.

I would love to have an example of building a usable python toolchain using `rules_foreign_cc`. I think it'd be super useful to may people and would give us a way to make sure the functionality there doesn't regress so yeah, opening another PR for that would be amazing! 😄 

I think the work in this PR is still super valuable and I want to get it in. I got stuck on the weird issue in https://github.com/bazelbuild/rules_foreign_cc/pull/650 and have had absolutely no time to continue digging on it 😞. I'd at least like to know what's going on there before making any merges so we don't accidentally merge something that causes more issues down the line. I have a busy weekend ahead of me but should have the time to continue this next weekend.

Sorry about the delay here but the PR has not been forgotten 😅

jheaff1:
@UebelAndre Please see PR #677 for adding examples on building python2/python3 to provide hermetic python toolchains

foreign_cc/private/framework/toolchains/linux_commands.bzl
```diff
@@ -71,10 +69,14 @@ if [ -d "$1" ]; then
   local files=$(find -P $1 \\( -type f -and \\( -name "*.pc" -or -name "*.la" -or -name "*-config" -or -name "*.mk" -or -name "*.cmake" \\) \\))
   IFS=$SAVEIFS
   for file in ${files[@]}; do
-    sed -i 's@'"$2"'@'"$3"'@g' "${file}"
+    local backup=$(mktemp)
```
jheaff1, line 72: @UebelAndre It appears that these changes have caused apr_util to fail as the "rules.mk" file is not in the sandbox anymore

UebelAndre: Try with the new changes? I was able to build with them.
```
//python:python2_test                                                    PASSED in 0.5s
//python:python3_test                                                    PASSED in 0.7s
```

jheaff1: After rebasing off your branch again, everything passes except for Python on MacOS standalone. Sadly I do not own a Mac so can't really debug it :(

UebelAndre: Concerning... Seems like a new build succeeded? Is there some flakiness going on here?

UebelAndre:
@googlebot I consent.

foreign_cc/private/framework/toolchains/linux_commands.bzl
jheaff1: I have run the following 5 times from the `examples/third_party` directory on my linux machine:

```bash
bazel clean --expunge
bazel test //python:python_tests
```

The tests passed each time. This would indicate the the issue is specifically to do with the standalone MacOS build

UebelAndre:
I'd say, if you can demonstrate the use of `output_data_dirs` in a smaller test within the `examples` package that'd be ideal. This way https://github.com/bazelbuild/rules_foreign_cc/pull/677 could be focused on using the changes here to provide a hermetic python toolchain. But It'd make sense to confirm everything in this PR so you can benefit from CI.

UebelAndre:
@jheaff1 this should be good to go for a rebase and then @jsharpe would you be able to do another review here?

jsharpe:
> @jheaff1 this should be good to go for a rebase and then @jsharpe would you be able to do another review here?

Sure thing

foreign_cc/private/framework.bzl
```diff
@@ -680,6 +685,10 @@ def _define_outputs(ctx, attrs, lib_name):
 
     out_include_dir = ctx.actions.declare_directory(lib_name + "/" + attrs.out_include_dir)
 
+    out_data_dirs = []
+    for dir in attr_out_data_dirs:
+        out_data_dirs.append(ctx.actions.declare_directory(lib_name + "/" + dir))
```
UebelAndre, line 690: Maybe just for sanity's sake?
```suggestion
        out_data_dirs.append(ctx.actions.declare_directory(lib_name + "/" + dir.lstrip("/")))
```

Not sure if that syntax works and not sure if this is actually a problem? Maybe double check but it'd be nice to either reject malformed input here or correct for it.

UebelAndre (changes requested):

UebelAndre:
An easy way to debug what's going on is to rebuild and set `--experimental_ui_max_stdouterr_bytes=2097152` so the error is visible.

.bazelci/config.yaml
```diff
@@ -34,6 +34,7 @@ tasks:
     linux_targets: &linux_targets_standalone
       - "//..."
       - "//:third_party_examples_linux_tests"
+      - "--experimental_ui_max_stdouterr_bytes=2097152"
```
UebelAndre, line 37: Ah, this isn't a target, it should be 
```yaml
build_flags:
  - "--experimental_ui_max_stdouterr_bytes=2097152"
```

jheaff1: oops, fixed

examples/third_party/python/BUILD.python3.bazel
```diff
@@ -15,7 +15,7 @@ configure_make(
     }),
     configure_options = [
         "--with-openssl=$EXT_BUILD_DEPS/openssl",
-        "--with-zlib=$EXT_BUILD_DEPS/zliasdfb",
+        "--with-zlib=$EXT_BUILD_DEPS/zlib",
```
UebelAndre, line 18: I think in this case bazel would just fetch the cached build. Perhaps re-ordering this list would trigger a new build?

examples/third_party/python/BUILD.python2.bazel
```diff
@@ -0,0 +1,55 @@
...
+        "//conditions:default": {},
+    }),
+    # rules_foreign_cc defaults the install_prefix to "python". This conflicts with the "python" executable that is generated.
+    install_prefix = "py_install",
+    lib_source = ":all_srcs",
+    out_binaries = [
+        "python2.7",
+    ],
+    out_data_dirs = ["lib"],
+    tools_deps = select({
```
jheaff1, line 41: `tool_deps` is now deprecated, change to `build_data`

jheaff1:
For some reason, the setting of `--experimental_ui_max_stdouterr_bytes` for the MacOS standalone examples doesn't seem to take effect. - https://buildkite.com/bazel/rules-foreign-cc/builds/2892#7c5682d8-e486-4a17-98d8-4f1f3edf99ee

UebelAndre:
> For some reason, the setting of `--experimental_ui_max_stdouterr_bytes` for the MacOS standalone examples doesn't seem to take effect. - https://buildkite.com/bazel/rules-foreign-cc/builds/2892#7c5682d8-e486-4a17-98d8-4f1f3edf99ee

That's annoying... https://github.com/bazelbuild/continuous-integration/issues/1184

Try adding the same flag to `test_flags`

jheaff1:
Sadly I won’t have time to work further on this PR tonight, I’ll resume tomorrow. 😢

On 16 Jun 2021, at 5:52 pm, UebelAndre ***@***.***> wrote:

﻿

For some reason, the setting of --experimental_ui_max_stdouterr_bytes for the MacOS standalone examples doesn't seem to take effect. - https://buildkite.com/bazel/rules-foreign-cc/builds/2892#7c5682d8-e486-4a17-98d8-4f1f3edf99ee

That's annoying... bazelbuild/continuous-integration#1184<https://github.com/bazelbuild/continuous-integration/issues/1184>

Try adding the same flag to test_flags

UebelAndre:
> Sadly I won’t have time to work further on this PR tonight, I’ll resume tomorrow. 😢

No worries! Have a good night 😄

jheaff1:
Finally got the failure showing in buildkite - https://buildkite.com/bazel/rules-foreign-cc/builds/2905#dacb1444-0390-4e2e-9fcc-20f4ca71d547

```


Looking in links: /tmp/tmpqia6elrs
Requirement already up-to-date: setuptools in ./py_install/lib/python3.9/site-packages (49.2.1)
Requirement already up-to-date: pip in ./py_install/lib/python3.9/site-packages (20.2.3)
make: Leaving directory '/private/var/tmp/_bazel_buildkite/3caab95e48b31e8eefc2ef3fc495033d/execroot/rules_foreign_cc_examples/bazel-out/host/bin/external/python3/python3.build_tmpdir'
+ set +x
cp: symlink: python3.9: File exists
```

examples/third_party/python/BUILD.python3.bazel
```diff
@@ -0,0 +1,43 @@
...
+
+configure_make(
+    name = "python3",
+    configure_env_vars = select({
+        "@platforms//os:macos": {"AR": ""},
+        "//conditions:default": {},
+    }),
+    configure_options = [
+        "--with-openssl=$EXT_BUILD_DEPS/openssl",
+        "--with-zlib=$EXT_BUILD_DEPS/zlibasdffdsaa",
```
UebelAndre, line 18: I wonder if it would have been better to pass random things to `args`. Is the fact that the build cannot find zlib the issue now? `args` are arguments to make which should be a way of introducing a new build variant that has no impact on the actual build itself.

UebelAndre:
> Finally got the failure showing in buildkite - https://buildkite.com/bazel/rules-foreign-cc/builds/2905#dacb1444-0390-4e2e-9fcc-20f4ca71d547
> 
> ```
> Looking in links: /tmp/tmpqia6elrs
> Requirement already up-to-date: setuptools in ./py_install/lib/python3.9/site-packages (49.2.1)
> Requirement already up-to-date: pip in ./py_install/lib/python3.9/site-packages (20.2.3)
> make: Leaving directory '/private/var/tmp/_bazel_buildkite/3caab95e48b31e8eefc2ef3fc495033d/execroot/rules_foreign_cc_examples/bazel-out/host/bin/external/python3/python3.build_tmpdir'
> + set +x
> cp: symlink: python3.9: File exists
> ```

I'm very confused now. Especially since if I build locally, the build succeeds. Is zlib not actually used for py3? Pretty confused about this failure but feels specific to standalone builds?

jheaff1:
It seems that `zlib` isn't actually a dependency of Python3. I'm not sure why I thought it was to begin with. I ran `./configure --with-zlib` in the `python3` repo and a warning appeared stating that `--with-zlib` is an unrecognised argument. I have now removed the dependency on `zlib`

jheaff1:
> It seems that `zlib` isn't actually a dependency of Python3. I'm not sure why I thought it was to begin with. I ran `./configure --with-zlib` in the `python3` repo and a warning appeared stating that `--with-zlib` is an unrecognised argument. I have now removed the dependency on `zlib`

Turns out I was correct in the first place. After removing the dependency on `zlib`, there is the following error:

```
ZipImportError: can't decompress data; zlib not available
```

I'll add zlib back as a dependency

UebelAndre:
> > It seems that `zlib` isn't actually a dependency of Python3. I'm not sure why I thought it was to begin with. I ran `./configure --with-zlib` in the `python3` repo and a warning appeared stating that `--with-zlib` is an unrecognised argument. I have now removed the dependency on `zlib`
> 
> Turns out I was correct in the first place. After removing the dependency on `zlib`, there is the following error:
> 
> ```
> ZipImportError: can't decompress data; zlib not available
> ```
> 
> I'll add zlib back as a dependency

Even more confusing then! lol How is it possible to build py3 if the path is mangled? Maybe it's a linux only feature? I pulled your branch earlier and successfully built. I'm otherwise not sure how that's possible.

jheaff1:
> > > It seems that `zlib` isn't actually a dependency of Python3. I'm not sure why I thought it was to begin with. I ran `./configure --with-zlib` in the `python3` repo and a warning appeared stating that `--with-zlib` is an unrecognised argument. I have now removed the dependency on `zlib`
> > 
> > 
> > Turns out I was correct in the first place. After removing the dependency on `zlib`, there is the following error:
> > ```
> > ZipImportError: can't decompress data; zlib not available
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'll add zlib back as a dependency
> 
> Even more confusing then! lol How is it possible to build py3 if the path is mangled? Maybe it's a linux only feature? I pulled your branch earlier and successfully built. I'm otherwise not sure how that's possible.

Yeah it is quite bizarre! Are you testing on Linux or Mac?

jheaff1:
The latest build (https://buildkite.com/bazel/rules-foreign-cc/builds/2913#7a58e48b-85b0-4b71-8e4f-6899742b121d) has the same error as before but for Python2 now:

```
cp: symlink: python2: File exists
```

UebelAndre:
> Yeah it is quite bizarre! Are you testing on Linux or Mac?

I'm only testing on my mac.



> The latest build (https://buildkite.com/bazel/rules-foreign-cc/builds/2913#7a58e48b-85b0-4b71-8e4f-6899742b121d) has the same error as before but for Python2 now:
> 
> ```
> cp: symlink: python2: File exists
> ```

I'm a bit busy today so I don't have too much time to dig into this but can you narrow down what specific call is failing here? Maybe by wrapping some commands in `set -x` + `set +x`?

UebelAndre:
Also worth noting that it's the standalone build which failed again.....

jheaff1:
The error seems to occur during or after the `make install` part of the build. The error may occur in the [`copy_dir_contents_to_dir`](https://github.com/bazelbuild/rules_foreign_cc/blob/fccd4ddaebb32de61da50d418db72aa64b316ab7/foreign_cc/private/framework/toolchains/macos_commands.bzl#L85) method

jheaff1:
I added debug prints to the start and end of `copy_contents_to_dir` (which is a recursive function) and the [build log](https://buildkite.com/bazel/rules-foreign-cc/builds/2916#ad1730e2-85b0-4c2f-8c3b-269a8a6c9ea4) suggests that it never got to the end of the function, so `copy_contents_to_dir` must be the culprit

foreign_cc/private/framework/toolchains/macos_commands.bzl
```diff
@@ -84,6 +84,7 @@ fi
 
 def copy_dir_contents_to_dir(source, target):
     text = """\
+echo "DEBUG! Start of copy_dir_contents_to_dir"
```
UebelAndre, line 87:
```suggestion
echo "DEBUG! Start of copy_dir_contents_to_dir"
set -x
```

UebelAndre:
Ok, finally, I think https://github.com/bazelbuild/rules_foreign_cc/pull/687 will solve the issue....

jheaff1:
> Ok, finally, I think #687 will solve the issue....

Awesome! I have merged your branch into this one to see if the standalone MacOS builds are successful first  time

jheaff1:
> > Ok, finally, I think #687 will solve the issue....
> 
> Awesome! I have merged your branch into this one to see if the standalone MacOS builds are successful first time

It passed!

jheaff1:
@UebelAndre  I have rebased off the `main` branch and resolved the build failures, although the "docs" portion of the build is now failing.

UebelAndre:
> @UebelAndre I have rebased off the `main` branch and resolved the build failures, although the "docs" portion of the build is now failing.

Hey, yeah, I unfortunately had to restore the need for PRs to commit docs. So you'll have to run `cd docs && bazel run :generate_docs` and commit the results. I'm working on a permanent solution for this but that should fix the docs issue.

I'm still waiting on my reviewer for https://github.com/bazelbuild/rules_foreign_cc/pull/687 so once that's in then I think this will immediately follow

jheaff1:
> > @UebelAndre I have rebased off the `main` branch and resolved the build failures, although the "docs" portion of the build is now failing.
> 
> Hey, yeah, I unfortunately had to restore the need for PRs to commit docs. So you'll have to run `cd docs && bazel run :generate_docs` and commit the results. I'm working on a permanent solution for this but that should fix the docs issue.
> 
> I'm still waiting on my reviewer for #687 so once that's in then I think this will immediately follow

Great, thanks. I figured the docs out about 10 mins ago and the build is now all green :)

jsharpe:
> I'm still waiting on my reviewer for #687 so once that's in then I think this will immediately follow

Done and merged.

jheaff1:
@UebelAndre @jsharpe This PR is now ready for re-review :)

UebelAndre (approved):
What an amazing job you've done @jheaff1 !!!!

I'll give @jsharpe a moment to look but I think this is ready to go in

examples/third_party/python/BUILD.python2.bazel
```diff
@@ -0,0 +1,47 @@
...
+)
+
+configure_make(
+    name = "python2",
+    build_data = select({
+        "@platforms//os:macos": ["@subversion"],
+        "//conditions:default": [],
+    }),
+    configure_options = [
+        "CFLAGS='-Dredacted=\"redacted\"'",
```
jsharpe, line 17: We need to revisit the quoting if this option is still required.

jsharpe: but at least this will act as a test to make sure its fixed!

jsharpe (approved):
LGTM

jheaff1:
> What an amazing job you've done @jheaff1 !!!!
> 
> I'll give @jsharpe a moment to look but I think this is ready to go in

Thank you very much indeed! Happy I could contribute to this awesome repo

PR #627, opened 2021-04-30
Removed cleanup functionality now that builds run in the Bazel sandbox

UebelAndre:


jsharpe:
Do we need this cleanup path when building with standalone strategy?

UebelAndre:
I'm not sure. I can do some tests though. Good call.

UebelAndre:
Oh, I accidentally pushed to the wrong branch. I'll close this for now since I don't think it's that harmful and haven't tested the behavior in standalone mode.

PR #629, opened 2021-04-30
Added a new `platform_info` target to the foreign_cc framework

UebelAndre:
I don't think this is something that should live in the foreign_cc toolchain. Ideally this would be replaced by something provided by Bazel but until then I like having it separated

### DIFF
--- a/foreign_cc/private/framework/platform.bzl
+++ b/foreign_cc/private/framework/platform.bzl
@@ -1,0 +1,67 @@
+"""A helper module containing tools for detecting platform information"""
+
+ForeignCcPlatformInfo = provider(
+    doc = "A provider containing information about the current platform",
+    fields = {
+        "os": "The platform os",
+    },
+)
+
+def _framework_platform_info_impl(ctx):
+    """The implementation of `framework_platform_info`
+
+    Args:
+        ctx (ctx): The rule's context object
+
+    Returns:
+        list: A provider containing platform info
+    """
+    return [ForeignCcPlatformInfo(
+        os = ctx.attr.os,
+    )]
+
+_framework_platform_info = rule(
+    doc = "A rule defining platform information used by the foreign_cc framework",
+    implementation = _framework_platform_info_impl,
+    attrs = {
+        "os": attr.string(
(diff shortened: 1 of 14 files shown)


foreign_cc/private/framework/platform.bzl
```diff
@@ -0,0 +1,67 @@
...
+    implementation = _framework_platform_info_impl,
+    attrs = {
+        "os": attr.string(
+            doc = "The platform's operating system",
+        ),
+    },
+)
+
+def framework_platform_info(name = "platform_info"):
+    """Define a target contianing platform information used in the foreign_cc framework"""
```
jsharpe, line 34:
```suggestion
    """Define a target containing platform information used in the foreign_cc framework"""
```

foreign_cc/private/framework/platform.bzl
```diff
@@ -0,0 +1,67 @@
...
+            "@platforms//os:ios": "ios",
+            "@platforms//os:linux": "linux",
+            "@platforms//os:macos": "macos",
+            "@platforms//os:none": "none",
+            "@platforms//os:openbsd": "openbsd",
+            "@platforms//os:qnx": "qnx",
+            "@platforms//os:tvos": "tvos",
+            "@platforms//os:watchos": "watchos",
+            "@platforms//os:windows": "windows",
+            "//conditions:default": "unknown",
```
jsharpe, line 49: Add nixos and wasi? https://github.com/bazelbuild/platforms/blob/d4c9d7f51a7c403814b60f66d20eeb425fbaaacb/os/BUILD#L83-L93

jsharpe (approved):
Just a small typo and a couple of extra conditions in the select but otherwise LGTM

foreign_cc/private/framework/platform.bzl
jsharpe: Ah ok, I'd not seen that those changes were only in 4.1 - given we don't have a toolchain defined for those platforms I guess its fine to leave them out for now.

jsharpe (approved):

PR #519, opened 2021-02-23
support *.mk files in a replace_in_file function (#502)

slsyy:
These files contains paths, which are used by dependencies. New
sandboxing feature has exposed that omission

Also set configure_in_place=True, so apr-1-config show existing APR_SOURCE_DIR

PR #635, opened 2021-05-08
Fixed silent failures replacing sandbox paths in outputs

UebelAndre:
In investigating build failures brought up in https://github.com/bazelbuild/rules_foreign_cc/pull/622 I uncovered some messages in the build logs:
```
sed: /private/var/tmp/_bazel_user/2d7cc63c99a1f0f43d4f45bc36f2be85/sandbox/darwin-sandbox/423/execroot/rules_foreign_cc_examples_third_party/bazel-out/host/bin/external/subversion/subversion.ext_build_deps/apr/lib/libapr-1.la: in-place editing only works for regular files
```

I found out that `sed` is unable to update paths that are symlinks. It requires a path to a real file.
We currently only have one use of `sed` and it's to update paths files which contain absolute paths to the current sandbox.
https://github.com/bazelbuild/rules_foreign_cc/blob/b8b88cd2d16035aa1639434eb808f4d67a34d5ae/foreign_cc/private/framework/toolchains/linux_commands.bzl#L68

This is not something that would likely be caught in CI because it's uncommon that we update an individual example. Updating an individual test which also has a dependency on another test would have created a situation where cached assets from a dependency were used to build the updated example. This build would fail as configuration files like the ones specified in the `find` command above would never have had these paths replaced and we'd attempt to access a file in an old sandbox. Thus leading to a failure as they're typically cleaned up after execution.

The fix here ensures that these configuration files are always copied into the build directory in a writable state and that `sed` is only ever given real files. `sed` errors will also be caught and cause build failures.

### DIFF
--- a/foreign_cc/private/framework/toolchains/linux_commands.bzl
+++ b/foreign_cc/private/framework/toolchains/linux_commands.bzl
@@ -112,7 +112,14 @@ def symlink_to_dir(source, target):
 local target="$2"
 mkdir -p "$target"
 if [[ -f "$1" ]]; then
-  ln -s -f -t "$target" "$1"
+  # In order to be able to use `replace_in_files`, we ensure that we create copies of specfieid
+  # files so updating them is possible.
+  if [[ "$1" == *.pc || "$1" == *.la || "$1" == *-config || "$1" == *.mk || "$1" == *.cmake ]]; then
+    dest="$target/$(basename $1)"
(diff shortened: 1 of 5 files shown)


UebelAndre:
Without realizing it I also included `*.mk` to the list of files to replace values in. This would address https://github.com/bazelbuild/rules_foreign_cc/pull/519

jsharpe (approved):

PR #643, opened 2021-05-12
Build make commands from correct attribute dict

sam-lunt:
This fixes the bug introduced in #556 so that the `make_commands` are built from `attrs`, not `ctx.attr`.


jsharpe (approved):
LGTM thanks @sam-lunt

UebelAndre (approved):
Thank you so much for putting this together and sorry about the bug you encountered 😅

sam-lunt:
@jsharpe thanks for the quick turnaround!
@UebelAndre anytime, and no worries : )

PR #644, opened 2021-05-12
Include `libtool` and `.nice` files when replacing sandbox paths

UebelAndre:


jsharpe:
What's a `.nice` file used for - I'm not familiar with that extension?

UebelAndre:
> What's a `.nice` file used for - I'm not familiar with that extension?

There were some `config.nice` files being generated by [apr](https://downloads.apache.org/apr/). I found a link to [apache httpd](https://httpd.apache.org/docs/trunk/install.html#upgrading) was all I found which to me implied they were some kind of configure script? I was hoping you'd recognize them 😅

jsharpe:
> > What's a `.nice` file used for - I'm not familiar with that extension?
> 
> There were some `config.nice` files being generated by [apr](https://downloads.apache.org/apr/). I found a link to [apache httpd](https://httpd.apache.org/docs/trunk/install.html#upgrading) was all I found which to me implied they were some kind of configure script? I was hoping you'd recognize them 😅

I suspect its something very specific to the apr / httpd build; I guess the only overhead in keeping them in there is the extra disk space usage so I guess I'm ok with this as a step to getting apr-util building but guess we should think about how to generalise that function of listing the build artifacts that need to have contents replaced.

UebelAndre:
> I suspect its something very specific to the apr / httpd build; I guess the only overhead in keeping them in there is the extra disk space usage so I guess I'm ok with this as a step to getting apr-util building but guess we should think about how to generalise that function of listing the build artifacts that need to have contents replaced.

Agreed. I'm really starting to feel like we need to ditch this shell trickery and actually write a process wrapper. This list could have some sane defaults but then ultimately be a user provided input.

jsharpe (approved):

PR #647, opened 2021-05-12
Restrict use of `use_default_shell_env` to windows.

UebelAndre:
The primary intent of this change is to have a more structured way of controlling the environment. The way it's currently implemented adds support for [--action_env](https://docs.bazel.build/versions/master/command-line-reference.html#flag--action_env) to MacOS platforms and adds a centralized place in code to manage environment variables.

### DIFF
--- a/foreign_cc/private/framework.bzl
+++ b/foreign_cc/private/framework.bzl
@@ -241,6 +241,48 @@ dependencies.""",
     ),
 )
 
+def _env_prelude(ctx, lib_name, data_dependencies, target_root):
+    """Generate a bash snippet containing environment variable definitions
+
+    Args:
+        ctx (ctx): The rule's context object
+        lib_name (str): The name of the target being built
+        data_dependencies (list): A list of targets representing dependencies
+        target_root (str): The path from the root target's directory in the build output
+
+    Returns:
+        tuple: A list of environment variables to define in the build script and a dict
+            of environment variables
+    """
+    env_snippet = [
+        "export EXT_BUILD_ROOT=##pwd##",
+        "export INSTALLDIR=$$EXT_BUILD_ROOT$$/" + target_root + "/" + lib_name,
+        "export BUILD_TMPDIR=$$INSTALLDIR$$.build_tmpdir",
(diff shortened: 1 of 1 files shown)


foreign_cc/private/framework.bzl
```diff
@@ -241,6 +241,48 @@ dependencies.""",
...
+    # Because windows will be using `use_default_shell_env`, any environment variables
+    # set on the build action will be ignored. To solve for this, we explicitly set
+    # environment variables to tools for the current cc_toolchain in use.
+    if "win" in os_name(ctx):
+        env_snippet.extend(["export {}=\"{}\"".format(key, cc_env[key]) for key in cc_env])
+
+    # Add all user defined variables
+    attr_env = dict()
+    for key, value in getattr(ctx.attr, "env", {}).items():
+        # Ensure the values of the environemnt variables have absolute paths
```
jsharpe, line 280:
```suggestion
        # Ensure the values of the environment variables have absolute paths
```

foreign_cc/private/framework.bzl
```diff
@@ -241,6 +241,48 @@ dependencies.""",
...
+        "export BUILD_TMPDIR=$$INSTALLDIR$$.build_tmpdir",
+        "export EXT_BUILD_DEPS=$$INSTALLDIR$$.ext_build_deps",
+    ]
+
+    # Start with the default shell env to capture `--action_env` args
+    env = dict(ctx.configuration.default_shell_env)
+
+    # Add all environment variables from the cc_toolchain
+    cc_env = _correct_path_variable(get_env_vars(ctx))
+    env.update(cc_env)
```
jsharpe, line 269: Is this the same way that the environment gets overridden for the native `cc_*` rules? I would guess that the behaviour is the opposite way around i.e. `action_env` overrides variables defined by the toolchain but 🤷

jsharpe (changes requested):

foreign_cc/private/framework.bzl
UebelAndre: Yeah, `action_env` should probably take precedence. I'll update this.

jsharpe (approved):

PR #648, opened 2021-05-12
Revert "Include `libtool` and `.nice` files when replacing sandbox paths (#644)"

UebelAndre:
This reverts commit 08598910609153bb82c7e3a91dbf7ae3d17a0575. (#644)


UebelAndre:
I don't think this change is necessary and feels quite specific to `apr`. I think it should be reverted and we should have a more focused solution for things that generate their own `libtool` binaries.

jsharpe (approved):

PR #652, opened 2021-05-21
Updated documentation and removed the need to commit updated docs

UebelAndre:
I wanted to update the docs to be more granular and move the source of truth to the `.bzl` files for some time now but the thing that inspired me was https://docs.aspect.dev/


UebelAndre:
Also @alexeagle is there a way to ensure the `docs.aspect.dev/rules_foreign_cc/` link will work similarly to 
`docs.aspect.dev/rules_nodejs/`? It seems like there's some special behavior for `rules_nodejs`.

alexeagle:
fixed https://docs.aspect.dev/rules_foreign_cc
Happy to chat with maintainers here about how we can do the docsite work for you, and you can just drop complexity around the API docs pipeline.

jsharpe (approved):

PR #653, opened 2021-05-21
set write permission on boost_build for INSTALL_DIR

thekyz:
Since the inputs are readonly in sandboxfs, when rules_foreign copies them the attributes get copied along and the folders stay read-only.

A lot of things inside `bootstrap.sh` and `b2` write directly on that folder so it's unavoidable if you want to support sandboxfs, although there might be a better way to do it but I could not find it.


thekyz:
The RBE failures really don't look related to my changes:
```

(13:32:56) ERROR: Traceback (most recent call last):
--
&nbsp; | File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/90e2e9b2d72fa9b00c19ff0535bb31ae/external/buildkite_config/java/BUILD", line 19, column 60, in &lt;toplevel&gt;
&nbsp; | load("@bazel_tools//tools/jdk:local_java_repository.bzl", "local_java_runtime")
&nbsp; | Error: file '@bazel_tools//tools/jdk:local_java_repository.bzl' does not contain symbol 'local_java_runtime'


```

jsharpe (approved):

Issue #449, opened 2020-12-27
How shared libs with soname should be handled?


slsyy:
Lets use `examples/apache_httpd` example with new `cc_binary` and `cc_test` targets:

```diff
+cc_binary(
+    name = "cc_binary",
+    srcs = ["main.cpp"],
+    deps = [":apr_util"],
+    # linkstatic=True is default
+)
+
+cc_test(
+    name = "cc_test",
+    srcs = ["main.cpp"],
+    deps = [":apr_util"],
+    # linkstatic=False is default
+)
```

and then build both binaries:
```shell
bazel clean
bazel test //apache_httpd:cc_binary 
ldd bazel-bin/apache_httpd/cc_binary
```

Result: OK, everything is statically linked
```shell
	linux-vdso.so.1 (0x00007ffd53fa4000)
	libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x00007f6b62512000)
	libm.so.6 => /usr/lib/libm.so.6 (0x00007f6b623cc000)
	libc.so.6 => /usr/lib/libc.so.6 (0x00007f6b62203000)
	/lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007f6b62723000)
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x00007f6b621e9000)
```

```shell
bazel clean
bazel test //apache_httpd:cc_test 
ldd bazel-bin/apache_httpd/cc_test
```

Result: system-wide shared libs are used, not acceptable as I want to use libs build by bazel
```shell
	linux-vdso.so.1 (0x00007ffdb3da2000)
	libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x00007ff10c441000)
	libm.so.6 => /usr/lib/libm.so.6 (0x00007ff10c2fb000)
	libaprutil-1.so.0 => /usr/lib/libaprutil-1.so.0 (0x00007ff10c2cc000)
	libapr-1.so.0 => /usr/lib/libapr-1.so.0 (0x00007ff10c28e000)
	libc.so.6 => /usr/lib/libc.so.6 (0x00007ff10c0c5000)
	/lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007ff10c652000)
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x00007ff10c0ab000)
[...]

PR #656, opened 2021-05-26
Fixed dangling symlinks in builds

UebelAndre:
In some cases, the assumed output for a build, `lib*.so`, is actually a symlink which points to `lib*.#.#.#.so`. This results in issues like
```
ERROR: /home/user/.cache/bazel/_bazel_user/3821989c4a000f3a0c8be3e30c90c4f7/external/libjpeg_turbo/BUILD.bazel:23:6: output 'external/libjpeg_turbo/libjpeg_turbo/lib/libjpeg.so' is a dangling symbolic link
```

To fix this, any explicitly requested output should be resolved to it's full file.


jsharpe (approved):
+1 as this fixes the dangling symlinks for the specified outputs but we still have #449 to think about how to handle properly - its not ideal that we end up with copies of shared libraries which should be symlinks. `ldconfig` complains about the fact that the `.so` library isn't a symlink.

I've not checked but do we also have broken symlinks in the `gen_dir` output too?

UebelAndre:
> +1 as this fixes the dangling symlinks for the specified outputs but we still have #449 to think about how to handle properly - its not ideal that we end up with copies of shared libraries which should be symlinks. `ldconfig` complains about the fact that the `.so` library isn't a symlink.
> 
> I've not checked but do we also have broken symlinks in the `gen_dir` output too?

Yeah, this would have been an existing issue from before https://github.com/bazelbuild/rules_foreign_cc/pull/575

This change restores previous behavior but is still incorrect but incorrect for functionality that didn't work or was undefined previously.

PR #657, opened 2021-05-26
Add instructions for pre-commit buildifier hook

alexeagle:
Note that https://github.com/bazelbuild/continuous-integration/issues/1161 means users will still be told to do this manually if they don't notice the CONTRIBUTING.md but can be fixed in the future


alexeagle:
will re-open matching the one that just landed in rules_apple

Issue #670, opened 2021-06-10
configure_make() seems to ignore the targets attribute


attilaolah:
It looks like passing `targets = […]` to `configure_make()` has no effect.

I think the `make_commands` arg should be removed from `configure_make`, similarly how it was removed from `make()` and `cmake()`, and replaced with `targets`. Then, if anyone needs a "special" make command, they could still pass in an empty list and have other commands executed in the `postfix_script`, which is more aligned with other rules in this repo.

UebelAndre:
Thanks for opening the issue! I'll have this fixed soon

Issue #668, opened 2021-06-09
Add some sort of "wrapper" argument


attilaolah:
Short version: it would be convenient to have a "wrapper" tool that gets executed as a prefix before `cmake`, `make` and `…/configure`. This is specifically to target the existing wrappers for Emscripten: `emcmake`, `emmake` and `emconfigure`.

---

Version 0.3.0 removes the `make_commands` argument from `make()`. This is great, as passing `targets` and `args` is much more flexible, and in general works better, however, I was (ab)using `make_commands` to pass my own wrapper, and thus call `emmake make` and `emmake make install` directly.

The same would be very useful for `configure_make()`. Currently I use a workaround where I patch in an `emconfigure.sh` into the source archives, which constructs the right command (`emconfigure configure …`) based on environment variables that I need to pass in at build time.

For CMake, there is no easy way to pass in such a wrapper, but the Emscripten CMake wrapper actually does very little other than passing in some cache entries like `CMAKE_TOOLCHAIN_FILE`, which sets `CMAKE_MODULE_PATH`, `CMAKE_CROSS_COMPILING_EMULATOR`, etc.

I've been trying to resist adding such a wrapper arg myself as I was instead trying to get all the right flags from Bazel, to get the regular `./configure` and `make` produce a `wasm32` build based on the CC toolchain flags alone. But it's getting harder to do that, and with the removal of `make_commands` from `make()` (which, btw, is still there in `configure_make`) it is quite a challenge now to pass in the right incantations to `make()` to get it to cross-compile properly.

UebelAndre:
I think the solution here is to register a `` toolchain using `native_tool_toolchain`:
https://github.com/bazelbuild/rules_foreign_cc/blob/2407938f22aaf09e5705e726037cf61e2003b291/toolchains/native_tools/native_tools_toolchain.bzl#L32-L61

We use that here:
https://github.com/bazelbuild/rules_foreign_cc/blob/2407938f22aaf09e5705e726037cf61e2003b291/toolchains/BUILD.bazel#L33-L37
https://github.com/bazelbuild/rules_foreign_cc/blob/2407938f22aaf09e5705e726037cf61e2003b291/toolchains/BUILD.bazel#L77-L84

If you build with this toolchain, then the environment variable `MAKE` will be set to whatever the executable is from that toolchain.

attilaolah:
Oh that's cool! I'll give that a try, thanks.

attilaolah:
[...]

PR #671, opened 2021-06-10
Removed `make_commands` attribute and fixed `configure_make`

UebelAndre:
Attributes which encourage users to write raw bash are generally non-hermetic and lead to a bunch of difficult edge cases to account for. The `targets` attribute is far more desirable since it allows the rules to provide stable and consistent APIs across all the rules. 

Additionally, this change fixes an inconsistency with `configure_make` where the targets API was not being used.
closes https://github.com/bazelbuild/rules_foreign_cc/issues/670


UebelAndre:
I think as part of this change I should also add an example of how to have a custom make binary to replace, what seems like, a common pattern where users use `make_commands` to prefix or change the tool being used based on the host's configuration. This is definitely not recommended but we should find a recommended approach. 

[This comment](https://github.com/bazelbuild/rules_foreign_cc/issues/668#issuecomment-858037712) describes such an approach but I'm not sure if it encompasses enough of the functionality.

UebelAndre:
> I think as part of this change I should also add an example of how to have a custom make binary to replace, what seems like, a common pattern where users use `make_commands` to prefix or change the tool being used based on the host's configuration. This is definitely not recommended but we should find a recommended approach.
> 
> [This comment](https://github.com/bazelbuild/rules_foreign_cc/issues/668#issuecomment-858037712) describes such an approach but I'm not sure if it encompasses enough of the functionality.

I think https://github.com/bazelbuild/rules_foreign_cc/pull/676 would solve for this in a way that makes transitioning easier since it doesn't require users to setup a toolchain

UebelAndre:
This PR is blocked until https://github.com/bazelbuild/rules_foreign_cc/pull/676 is resolved

attilaolah:
Cool, this is unblocked now!

By the way, once this gets merged, and now that #676 is merged too, it would be really handy to have a new release :)

UebelAndre:
@attilaolah This one is ready for review. Let me know if it solves the last issue you're having.

attilaolah:
I'll probably have some timet o take a look at this later today. Running some build tests now to make sure that I have a clean/passing build before patching in this one.

attilaolah:
OK I patched in this change, and all my builds still pass, and the `targets` and `angs` are both working as expected with `configure_make()`. I think this is good to go!

attilaolah (approved):

Issue #666, opened 2021-06-08
Remove env_vars attribute of cmake()


attilaolah:
For some `cmake()` rule where I am passing both `env` and `env_vars`, and both are provided by a `select()`, I somehow get the error below:

```
ERROR: /build/lib/hdf/BUILD.bazel:54:10: in cmake rule //lib/hdf:hdf: 
Traceback (most recent call last):
        File "/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/external/rules_foreign_cc/foreign_cc/cmake.bzl", line 58, column 33, in _cmake_impl
                return cc_external_rule_impl(ctx, attrs)
        File "/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/external/rules_foreign_cc/foreign_cc/private/framework.bzl", line 371, column 38, in cc_external_rule_impl
                ] + attrs.create_configure_script(ConfigureParameters(ctx = ctx, attrs = attrs, inputs = inputs)) + make_commands + postfix_script + [
        File "/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/external/rules_foreign_cc/foreign_cc/cmake.bzl", line 122, column 43, in _create_configure_script
                configure_script = create_cmake_script(
        File "/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/external/rules_foreign_cc/foreign_cc/private/cmake_script.bzl", line 50, column 48, in create_cmake_script
                params = _create_cache_entries_env_vars(toolchain_dict, user_cache, user_env)
        File "/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/external/rules_foreign_cc/foreign_cc/private/cmake_script.bzl", line 170, column 22, in _create_cache_entries_env_vars
                _move_dict_values(toolchain_dict, user_env, _CMAKE_ENV_VARS_FOR_CROSSTOOL)
        File "/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/external/rules_foreign_cc/foreign_cc/private/cmake_script.bzl", line 221, column 31, in _move_dict_values
                value = source.pop(key)
Error in pop: trying to mutate a frozen dict value
ERROR: Analysis of target '//lib/hdf:hdf' failed; build aborted: Analysis of target '//lib/hdf:hdf' failed
INFO: Elapsed time: 1.853s
[...]

PR #675, opened 2021-06-13
Removed legacy `*env_vars` attributes from `cmake` and `configure_make` rule

UebelAndre:
The  `env` attribute, which is common to all rules, should be used instead.

closes https://github.com/bazelbuild/rules_foreign_cc/issues/666


UebelAndre:
@attilaolah if you wanted to take a look, I think this would simplify the attributes.

attilaolah:
Cool, I'll take a look (but not today, I'm busy now.) Thanks for preparing this.

attilaolah:
Let me know when this is ready for review, I have a few uses of `*env_vars` that need to be migrated to `env`.

UebelAndre:
@attilaolah This is now ready for review 😃

attilaolah:
Cool! Let me do a quick test at 12c3956e5c0689b49edb8a565e8264d23b4a74eb, then I'll patch this in and re-run all the builds.

Thanks for keeping me in mind with all these pull request! I really appreciate it.

attilaolah:
Actually, I hit some other problems I need to solve first: looks like updating the CMake version to 3.20.4 broke several builds, now they fail with this error:

```
+ /root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/sandbox/processwrapper-sandbox/19/execroot/wasm/external/emscripten_bin_linux/emscripten/emcmake /root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/sandbox/processwrapper-sandbox/19/execroot/wasm/external/cmake-3.20.4-linux-x86_64/bin/cmake --build . --config Release
configure: /root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/sandbox/processwrapper-sandbox/19/execroot/wasm/external/cmake-3.20.4-linux-x86_64/bin/cmake --build . --config Release -DCMAKE_TOOLCHAIN_FILE=/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/external/emscripten_bin_linux/emscripten/cmake/Modules/Platform/Emscripten.cmake -DCMAKE_CROSSCOMPILING_EMULATOR=/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/sandbox/processwrapper-sandbox/19/execroot/wasm/external/nodejs_linux_amd64/bin/node
Unknown argument -DCMAKE_TOOLCHAIN_FILE=/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/external/emscripten_bin_linux/emscripten/cmake/Modules/Platform/Emscripten.cmake
Usage: cmake --build [<dir> | --preset <preset>] [options] [-- [native-options]]
Options:
  <dir>          = Project binary directory to be built.
  --preset <preset>, --preset=<preset>
                 = Specify a build preset.
  --list-presets
                 = List available build presets.
  --parallel [<jobs>], -j [<jobs>]
                 = Build in parallel using the given number of jobs. 
                   If <jobs> is omitted the native build tool's 
                   default number is used.
                   The CMAKE_BUILD_PARALLEL_LEVEL environment variable
                   specifies a default parallel level when this option
                   is not given.
  --target <tgt>..., -t <tgt>... 
                 = Build <tgt> instead of default targets.
  --config <cfg> = For multi-configuration tools, choose <cfg>.
  --clean-first  = Build target 'clean' first, then build.
                   (To clean only, use --target 'clean'.)
  --verbose, -v  = Enable verbose output - if supported - including
                   the build commands to be executed. 
  --             = Pass remaining options to the native tool.
emcmake: error: '/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/sandbox/processwrapper-sandbox/19/execroot/wasm/external/cmake-3.20.4-linux-x86_64/bin/cmake --build . --config Release -DCMAKE_TOOLCHAIN_FILE=/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/external/emscripten_bin_linux/emscripten/cmake/Modules/Platform/Emscripten.cmake -DCMAKE_CROSSCOMPILING_EMULATOR=/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/sandbox/processwrapper-sandbox/19/execroot/wasm/external/nodejs_linux_amd64/bin/node' failed (returned 1)
```

I'll investigate that a bit (or open a bug if need be) before I can take a look at this pull request.

attilaolah:
I filed #685, in the meantime I can pin to an older version of CMake to help things move forward.

foreign_cc/cmake.bzl
```diff
@@ -218,6 +218,8 @@ def _create_configure_script(configureParameters):
         for arg in ctx.attr.build_args
     ])
 
+    user_env = {key: ctx.expand_location(value, data) for (key, value) in ctx.attr.env}
```
attilaolah, line 221: I think this line might be causing these errors:

```
ERROR: /build/lib/z/BUILD.bazel:6:19: in configure_make rule //lib/z:z: label '//lib/python:runtime' in $(locations) expression is not a declared prerequisite of this rule
```

Or maybe I'm doing something wrong, but just changing `configure_env_vars` in my `//lib/z` to `//vars` breaks the build with `--cpu=wasm32`. I am passing these env vars:

```
"PATH": "${EXT_BUILD_ROOT}/$(execpaths //lib/python:runtime)/bin:${PATH}",
"PYTHONHOME": "${EXT_BUILD_ROOT}/$(execpaths //lib/python:runtime)",
```

attilaolah: Or rather, the issue I'm facing is coming from `configure.bzl` but the idea is the same. How come the locations were expanded earlier, even though you did not call `ctx.expand_location()` on the `env_var` attribute earlier?

attilaolah (commented):
I'm having some issues with expanding the location (via `$(execpaths …)`) so I couldn't get this to work properly yet, but I'll try to dig deeper.

foreign_cc/cmake.bzl
UebelAndre: Can you share the build script that's generated in the before and after?

UebelAndre: Doesn't `execpaths` return multiple files? What is `${EXT_BUILD_ROOT}/$(execpaths //lib/python:runtime)/bin:${PATH}` expected to render into?

attilaolah: Yes, `execpaths` returns multiple paths, but in my case `//lib/python:runtime` is a filegroup that selects the `gen_dir` of the Python target, so it only contains one file (the directory being built).

```bazel
configure_make(
    name = "python",
    …
)

filegroup(
    name = "runtime",
    srcs = [":python"],
    output_group = "gen_dir",
)
```

So this way `//lib/python:runtime` expands to just one value which contains the binaries, libraries, header files and the `$PYTHONHOME`.

attilaolah: Somehow I'm still getting the same error, even after applying 7ae1ea4. This line is for example now failing:

https://github.com/attilaolah/wasm/blob/6f2bac587be981264e90cf4addbbc9bbb7454006/lib/gmp/BUILD.bazel#L15

With the error:

```
ERROR: /build/lib/gmp/BUILD.bazel:7:19: in configure_make rule //lib/gmp:gmp: label '//lib/m4:runtime' in $(locations) expression is not a declared prerequisite of this rule
```

UebelAndre: Oh, I think this was a dumb oversight on my part 😅 

Try now and see the last commit.

attilaolah: Now it works! 🎉

But you fixed it so quickly, that I switched to using a `local_repository()` in the meantime to try and debug the issue. Then I added your fork as a remote and checked it out locally, but by the time I did that, you already committed the fix, so then I had my build passing with a local repo, but failing when patched in this pull request, and so I was scratching my head trying to figure out how the tow was different. 😀

Anyhow, thanks for fixing, I think this should be good to go but I'll quickly do a final build just to double-check.

UebelAndre: Let me know when you've double checked. Then I'd ask that you review and approve the changes here and we can get this closed out 😄

attilaolah:
Another problem is that environment variables with spaces are causing a problem since they are not quoted in the build script, e.g. they look like this now:

```
export CFLAGS=-Wno-error=incompatible-library-redeclaration -Wno-error=unused-but-set-variable
```

attilaolah:
Thanks! Now all my tests pass, just waiting for the final build to confirm that nothing breaks.

attilaolah:
Everything builds now, thanks for fixing up the issues! This should be good to go now.

:shipit:

attilaolah (approved):

attilaolah (approved):

PR #676, opened 2021-06-13
Added `tool_prefix` attribute

UebelAndre:
This adds the `tool_prefix` and `build_data` attribute and deprecates `additional_tools`, `additional_inputs`, and `tool_deps`.

`tool_prefix` is expected to solve for issues brought up in https://github.com/bazelbuild/rules_foreign_cc/pull/671#issuecomment-858932271

Basically, some uses of build tools must be prefixed by things like `emcmake` and `emmake` for [emscripten](https://emscripten.org/). These toolchains should be provided as `build_data` and consumed in `tool_prefix` using [make variable expansion](https://docs.bazel.build/versions/main/be/make-variables.html#predefined_label_variables)

### DIFF
--- a/foreign_cc/private/framework.bzl
+++ b/foreign_cc/private/framework.bzl
@@ -100,7 +100,7 @@ CC_EXTERNAL_RULE_ATTRIBUTES = {
     "env": attr.string_dict(
         doc = (
             "Environment variables to set during the build. " +
-            "`$(execpath)` macros may be used to point at files which are listed as `data`, `deps`, or `build_deps`, " +
+            "`$(execpath)` macros may be used to point at files which are listed as `data`, `deps`, or `build_data`, " +
             "but unlike with other rules, these will be replaced with absolute paths to those files, " +
             "because the build does not run in the exec root. " +
             "No other macros are supported."
--- a/foreign_cc/private/framework.bzl
+++ b/foreign_cc/private/framework.bzl
@@ -184,6 +184,10 @@ CC_EXTERNAL_RULE_ATTRIBUTES = {
         ),
         mandatory = False,
     ),
+    "tool_prefix": attr.string(
+        doc = "A prefix for build commands",
+        mandatory = False,
+    ),
     "tools_deps": attr.label_list(
         doc = "__deprecated__: Please use the `build_data` attribute.",
         mandatory = False,
@@ -457,7 +461,7 @@ def cc_external_rule_impl(ctx, attrs):
         runfiles = runfiles.merge(target[DefaultInfo].default_runfiles)
 
     # TODO: `additional_inputs` is deprecated, remove.
-    for legacy in [ctx.attr.additional_inputs]:
+    for legacy in ctx.attr.additional_inputs:
         runfiles = runfiles.merge(legacy[DefaultInfo].default_runfiles)
 
     externally_built = ForeignCcArtifactInfo(
--- a/foreign_cc/private/framework.bzl
+++ b/foreign_cc/private/framework.bzl
@@ -41,8 +41,7 @@ load(
 CC_EXTERNAL_RULE_ATTRIBUTES = {
     "additional_inputs": attr.label_list(
         doc = (
-            "Optional additional inputs to be declared as needed for the shell script action." +
-            "Not used by the shell script part in cc_external_rule_impl."
+            "__deprecated__: Please use the `build_data` attribute."
         ),
         mandatory = False,
         allow_files = True,
@@ -50,8 +49,7 @@ CC_EXTERNAL_RULE_ATTRIBUTES = {
     ),
     "additional_tools": attr.label_list(
         doc = (
-            "Optional additional tools needed for the building. " +
-            "Not used by the shell script part in cc_external_rule_impl."
+            "__deprecated__: Please use the `build_data` attribute."
         ),
         mandatory = False,
         allow_files = True,
@@ -66,10 +64,18 @@ CC_EXTERNAL_RULE_ATTRIBUTES = {
         mandatory = False,
         default = False,
     ),
+    "build_data": attr.label_list(
+        doc = "Files needed by this rule only during build/compile time. May list file or rule targets. Generally allows any target.",
+        mandatory = False,
+        allow_files = True,
+        cfg = "exec",
+        default = [],
+    ),
     "data": attr.label_list(
         doc = "Files needed by this rule at runtime. May list file or rule targets. Generally allows any target.",
         mandatory = False,
         allow_files = True,
+        cfg = "target",
         default = [],
     ),
     "defines": attr.string_list(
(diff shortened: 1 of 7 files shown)


UebelAndre:
@attilaolah does this PR work for your use case?

attilaolah:
I'll patch it in today and give it a try, can you please hold off for a day? Thanks so much for putting this together!

attilaolah:
There are a few things I'm having troubles with:

- Passing `tool_prefix` to `configure_make` doesn't seem to add the prefix before `configure`.
-  More importantly, the targets I add to `build_data` seem to be not there? Or I'm having troubles finding them. When I add a repo, that seems to be there, but when I add an `sh_binary` or another target build with `foreign_cc` itself, like Python, I can't find where the contents are put.

attilaolah:
Ah OK, so I think I've found out the answer for the second issue: the deps are there but there is no env var pointing at them, and since they are built for a different target (`exec` in my case), it is a bit hard to find them. I think `$(execpath …)` should work though, let me try.

UebelAndre:
>     * Passing `tool_prefix` to `configure_make` doesn't seem to add the prefix before `configure`.

In the configure case, does there need to be a separate prefix for the call to `configure` vs `make`?

attilaolah:
According to https://emscripten.org/docs/compiling/Building-Projects.html#integrating-with-a-build-system, it should be both: `emconfigure ./configure`, then `emmake make`.

attilaolah:
That complicates things a bit though, since the `tool_prefix` in the case of `configure_make()` is ambiguous now, since we have more than one tool. Or, well, the tool is `make`, the configure script is `configure`, so maybe `configure_prefix` + `tool_prefix` would work in the case of `configure_make()`?

attilaolah:
I get an error now:

```
        File "/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/external/rules_foreign_cc/foreign_cc/configure.bzl", line 53, column 33, in _configure_make
                return cc_external_rule_impl(ctx, attrs)
        File "/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/external/rules_foreign_cc/foreign_cc/private/framework.bzl", line 402, column 38, in cc_external_rule_impl
                ] + attrs.create_configure_script(ConfigureParameters(ctx = ctx, attrs = attrs, inputs = inputs)) + make_commands + postfix_script + [
        File "/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/external/rules_foreign_cc/foreign_cc/configure.bzl", line 78, column 96, in _create_configure_script
                configure_prefix = "{} ".format(ctx.expand_location(attrs.configure_prefix, data)) if attrs.configure_prefix else ""
Error: 'struct' value has no field or method 'configure_prefix'
```

UebelAndre:
> That complicates things a bit though, since the `tool_prefix` in the case of `configure_make()` is ambiguous now, since we have more than one tool. Or, well, the tool is `make`, the configure script is `configure`, so maybe `configure_prefix` + `tool_prefix` would work in the case of `configure_make()`?

Without having read this, I added `configure_prefix` 😄 since we'd already had a `configure_command` attribute and the tool actually building the code here is `make` (to my understanding). So `tool_prefix` is still associated with `make`.

UebelAndre:
> I get an error now:
> 
> ```
> Traceback (most recent call last):
>         File "/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/external/rules_foreign_cc/foreign_cc/configure.bzl", line 53, column 33, in _configure_make
>                 return cc_external_rule_impl(ctx, attrs)
>         File "/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/external/rules_foreign_cc/foreign_cc/private/framework.bzl", line 402, column 38, in cc_external_rule_impl
>                 ] + attrs.create_configure_script(ConfigureParameters(ctx = ctx, attrs = attrs, inputs = inputs)) + make_commands + postfix_script + [
>         File "/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/external/rules_foreign_cc/foreign_cc/configure.bzl", line 78, column 96, in _create_configure_script
>                 configure_prefix = "{} ".format(ctx.expand_location(attrs.configure_prefix, data)) if attrs.configure_prefix else ""
> Error: 'struct' value has no field or method 'configure_prefix'
> ```

Oops 😓 Try now!

attilaolah:
Yay, it works!

But, now with `configure_make()`, `configure_prefix` is added, but `tool_prefix` is not added before `make`. (My project still builds properly though, so maybe it's not so important?)

UebelAndre:
> Yay, it works!
> 
> But, now with `configure_make()`, `configure_prefix` is added, but `tool_prefix` is not added before `make`. (My project still builds properly though, so maybe it's not so important?)

I think that's because of the same issue from https://github.com/bazelbuild/rules_foreign_cc/pull/671

UebelAndre:
@attilaolah if you wanted to be my reviewer here and https://github.com/bazelbuild/rules_foreign_cc/pull/671, I could merge and get you on your way 😄 (assuming everything actually works for you).

foreign_cc/private/framework.bzl
```diff
@@ -94,7 +100,7 @@ CC_EXTERNAL_RULE_ATTRIBUTES = {
     "env": attr.string_dict(
         doc = (
             "Environment variables to set during the build. " +
-            "`$(execpath)` macros may be used to point at files which are listed as data deps, tools_deps, or additional_tools, " +
+            "`$(execpath)` macros may be used to point at files which are listed as `data`, `deps`, or `build_deps`, " +
```
attilaolah, line 103: Do you mean `build_data`?

attilaolah (approved):
Thanks! This looks good, and as it is it makes all my current tests pass.

Only point is what looks like a typo in the `env` attribute documentation.

attilaolah:
The only remaining issue now is #671, but we can deal with that separately.

foreign_cc/private/framework.bzl
UebelAndre: Yes! Good catch, thanks!

jheaff1:
Should the documentation in the `docs/` directory be updated to refer to `build_data` rather than `tools_deps`?

UebelAndre:
We no longer update things in `docs/`. Instead, we rely on https://docs.aspect.dev/ to auto-generate the documentation for us using the [stardoc](https://github.com/bazelbuild/stardoc) rules we write. Once we cut a new release, you should see the updated message:
https://github.com/bazelbuild/rules_foreign_cc/blob/fccd4ddaebb32de61da50d418db72aa64b316ab7/foreign_cc/private/framework.bzl#L181-L187

jheaff1:
Ah I see, thanks for the clarification :)

jheaff1:
Are you guys aware that the docs seem to point to dead lines?

i.e. if you press "configure_make" at https://docs.aspect.dev/bazelbuild/rules_foreign_cc/0.3.0/docs/ you get a "Page Not Found" error

UebelAndre:
> Are you guys aware that the docs seem to point to dead lines?
> 
> i.e. if you press "configure_make" at https://docs.aspect.dev/bazelbuild/rules_foreign_cc/0.3.0/docs/ you get a "Page Not Found" error

Yeah, the links within the docs doesn't work right now.... You can still navigate using the side bar but any links we have to internal documentation is currently broken. I'd say raise the issue in the [slack channel](https://slack.bazel.build/) to try and get some eyes on it. Asking in `#general` would be a good place to start.

allsey87:
@attilaolah @UebelAndre if I may quickly ask, how are you currently extracting Emscripten's output? `wasm_cc_binary` is not designed to work with `rules_foreign_cc`, right?

PR #677, opened 2021-06-14
Add regression tests for python dependencies

jheaff1:
This pull request adds regression tests for building dependencies of Python2 and Python3. This PR is a precursor to a PR which will provide the ability to build a hermetic Python2/3 toolchain using rules_foreign_cc

### DIFF
--- a/examples/third_party/BUILD.bazel
+++ b/examples/third_party/BUILD.bazel
@@ -17,6 +17,7 @@ test_suite(
         "//libssh2:libssh2_build_test",
         "//openssl:openssl_build_test",
         "//pcre:pcre_build_test",
+        "//sqlite:sqlite_build_test",
     ],
 )
 
@@ -39,6 +40,8 @@ test_suite(
         "//libssh2:libssh2_build_test",
         "//openssl:openssl_build_test",
         "//pcre:pcre_build_test",
+        # Missing a new enough automake to build
+        #"//sqlite:sqlite_build_test",
     ],
 )
 
@@ -59,6 +62,7 @@ test_suite(
         "//libssh2:libssh2_build_test",
         "//openssl:openssl_build_test",
         "//pcre:pcre_build_test",
+        "//sqlite:sqlite_build_test",
     ],
 )
 
--- a/examples/third_party/BUILD.bazel
+++ b/examples/third_party/BUILD.bazel
@@ -4,6 +4,8 @@ test_suite(
     name = "linux_tests",
     tags = ["manual"],
     tests = [
+        "//apr:apr_build_test",
+        "//apr_util:apr_util_build_test",
         "//bison:bison_build_test",
         "//cares:test_c_ares",
         "//curl:curl_build_test",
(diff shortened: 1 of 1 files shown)


UebelAndre:
Can this PR be repurposed into just adding all the python dependencies if https://github.com/bazelbuild/rules_foreign_cc/pull/650 makes it in (and is indeed the key to getting them working)? It'd be nice to reduce the footprint of https://github.com/bazelbuild/rules_foreign_cc/pull/622

jheaff1:
> Can this PR be repurposed into just adding all the python dependencies if #650 makes it in (and is indeed the key to getting them working)? It'd be nice to reduce the footprint of #622

Sure I can do that. Perhaps instead I should open a separate PR for that and then this PR can continue to be focused on building a hermetic Python toolchain?

UebelAndre:
I thought https://github.com/bazelbuild/rules_foreign_cc/pull/622 was providing that but yeah, having the examples separate could also work. I might even absorb the dependencies in https://github.com/bazelbuild/rules_foreign_cc/pull/650 which I think is appropriate since it would be providing the change and a regression test. That's probably best?

jheaff1:
> I thought #622 was providing that but yeah, having the examples separate could also work. I might even absorb the dependencies in #650 which I think is appropriate since it would be providing the change and a regression test. That's probably best?

Sorry I got this PR mixed up with #622. Im happy to repurpose this PR to add the python dependencies to the examples (apr, apr_util, SQLite, subversion)

jheaff1:
@UebelAndre I have now repurposed this PR to provide regression tests for Python dependencies :)

UebelAndre (approved):
Oops, I should have merged this in the earlier PR. Thanks so much for catching this!!

jheaff1:
@UebelAndre Sadly the build for subversion fails. I reverted commit 9932c7dbc2ef0dd15d60685172e71c350f757069 and it passes for me locally

UebelAndre:
For expedience sake, I opened https://github.com/bazelbuild/rules_foreign_cc/pull/682

UebelAndre:
> @UebelAndre Sadly the build for subversion fails. I reverted commit [9932c7d](https://github.com/bazelbuild/rules_foreign_cc/commit/9932c7dbc2ef0dd15d60685172e71c350f757069) and it passes for me locally (on linux)

Ah, that actually makes sense. That change fixed the fact that the `targets` attribute was being ignored. So the build commands that were actually in use were actually being used were simply
```bash
make 
make install
```
instead of the configured `bin` and `install` targets. This means when we ran `install` only having produced `bin`, we ran into an error for missing files.

UebelAndre:
With https://github.com/bazelbuild/rules_foreign_cc/pull/677#issuecomment-862436317 said and https://github.com/bazelbuild/rules_foreign_cc/pull/682 merged, I'll close this PR but again, thanks @jheaff1 for putting this together. I would have made changes here if github made it easier to suggest arbitrary changes.

jheaff1:
> With [#677 (comment)](https://github.com/bazelbuild/rules_foreign_cc/pull/677#issuecomment-862436317) said and #682 merged, I'll close this PR but again, thanks @jheaff1 for putting this together. I would have made changes here if github made it easier to suggest arbitrary changes.

No problem at all, thanks for your help!

PR #686, opened 2021-06-16
Fix tool_prefix for CMake

### LINKED ISSUE #685: CMake failure after upgrading to CMake 2.20.4
See the details in https://github.com/bazelbuild/rules_foreign_cc/pull/675#issuecomment-862654922.

It looks like recent CMake versions don't expect `-Dvar=value` flags when building with `--build`. However, what the `tool_prefix` does now, is it adds the prefix before the `cmake --build` and `cmake --install`, but not before the first cmake installation.

Note that this might be an issue with the Emscripten `emcmake` wrapper. What we do now is:

```
/path/to/cmake
/path/to/emcmake /path/to/cmake --build .
/path/to/emcmake /path/to/cmake --install .
```

So ideally, instead I would want to invoke CMake like so:

```
/path/to/emcmake /path/to/cmake
/path/to/cmake --build .
/path/to/cmake --install .
```

attilaolah:
Set the prefix before the first CMake invocation (where all the extra flags are passed), and don't set the prefix for the `--build` and `--install` invocations of CMake.

Fixes #685.

### DIFF
--- a/test/cmake_text_tests.bzl
+++ b/test/cmake_text_tests.bzl
@@ -253,7 +253,7 @@ def _merge_flag_values_no_toolchain_file_test(ctx):
         user_env,
         [],
         cmake_commands = [],
-        cmake_prefix = "emcmake"
+        cmake_prefix = "emcmake",
     )
     expected = r"""export CC="/usr/bin/gcc"
 export CXX="/usr/bin/gcc"
--- a/test/cmake_text_tests.bzl
+++ b/test/cmake_text_tests.bzl
@@ -253,12 +253,13 @@ def _merge_flag_values_no_toolchain_file_test(ctx):
         user_env,
         [],
         cmake_commands = [],
+        cmake_prefix = "emcmake"
     )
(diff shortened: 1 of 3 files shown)


attilaolah:
Hold off, this is not good yet.

UebelAndre (approved):
This seems good to me, thanks for updating one of the unittests!

Just need to fix buildifier and seems good to go!

test/cmake_text_tests.bzl
```diff
@@ -253,12 +253,13 @@ def _merge_flag_values_no_toolchain_file_test(ctx):
         user_env,
         [],
         cmake_commands = [],
+        cmake_prefix = "emcmake"
```
UebelAndre, line 256:
```suggestion
        cmake_prefix = "emcmake",
```

attilaolah: Oops, forgot to run buildifier :) Thanks for fixing up!

PR #687, opened 2021-06-18
Updated macos `copy_dir_contents_to_dir` to be like other platforms

UebelAndre:
Unfortunately, the implementation of `cp` for macos does not have `--no-target-directory`
```
-T, --no-target-directory
              treat DEST as a normal file
```

This is likely why the macos implementation of `copy_dir_contents_to_dir` always differered (looking back to https://github.com/bazelbuild/rules_foreign_cc/pull/196 and https://github.com/bazelbuild/rules_foreign_cc/pull/377).

This change aims to bring the macos implementation more into alignment and avoid the recursive function that was there before.

https://unix.stackexchange.com/a/109704 seems to be a good description of the behavior of `--no-target-directory` and I believe I've encompassed that in this change.


UebelAndre:
Sounds like this resolved issues for https://github.com/bazelbuild/rules_foreign_cc/pull/622#issuecomment-863869988

jsharpe (approved):
I wonder if we can achieve the same with `ditto` (https://ss64.com/osx/ditto.html) but lets go with this as its been tested and works

UebelAndre:
> I wonder if we can achieve the same with `ditto` (https://ss64.com/osx/ditto.html) but lets go with this as its been tested and works

OOH! I didn't know that existed, very interesting! If there are any issues from this, that'll definitely be the first thing to try.

PR #689, opened 2021-06-19
Added foreign_cc_binary/library rule

UebelAndre:


zachgrayio:
this looks great, was just thinking about trying something similar

foreign_cc/foreign_cc.bzl
```diff
@@ -0,0 +1,89 @@
...
+    ctx.actions.symlink(
+        output = output,
+        target_file = executable,
+        is_executable = True,
+    )
+
+    return [
+        DefaultInfo(
+            executable = output,
+            runfiles = ctx.attr.target[DefaultInfo].default_runfiles,
```
attilaolah, line 70: I'm guessing this is the main benefit of the foreign_cc_binary, so that we get the runfiles by default?

attilaolah:
I'd be interested in reviving these rules. Even if they go in some sort of `contrib` or `examples` directory, I think there is value in them.

However, I do not feel very strongly about them since I've already sort-of implemented my own version. But I'd be happy to replace my home-grown rules with more "official" ones.

attilaolah:
Please don't close it just yet :)

pjjw:
this seems to work a lot better than the `runnable_files` macro currently in main. do you think we could just replace that macro with this `foreign_cc_binary` impl?

jsharpe:
> this seems to work a lot better than the `runnable_files` macro currently in main. do you think we could just replace that macro with this `foreign_cc_binary` impl?

I'd be happy to merge a PR that did that but I have no time to actually implement said PR; feel free to give it a go.

UebelAndre:
I'm not working this. but someone anyone is free to revive it if they thought it'd be useful!

Issue #690, opened 2021-06-20
Put the docs back!


zachgrayio:
I recently discovered that the documentation has been removed in favor of a third-party site in this [commit](https://github.com/bazelbuild/rules_foreign_cc/commit/f471bbfeb21081c63e00611ae88e3e95d4cbc639#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5).

While I applaud this as a clever SEO strategy on the part of the consultant, this is just a net negative for everyone else, and I'm frankly surprised to see this kind of change coming through to repos hosted under `bazelbuild`. I understand that this is no longer maintained by Google, but even still... 🤦 

I'm all for improving the docs for `bazelbuild` projects but this is clearly the wrong direction. 

Revert #652

UebelAndre:
Sorry the change to the docs upset you! But I thought all the docs and previously links were still valid and active. Can you elaborate more on how this commit impacted you?

zachgrayio:
I think the diff really speaks for itself here honestly, but sure.

As you are aware, it's a common practice to keep the `WORKSPACE` rule boilerplate necessary to add the ruleset to a project, and in many cases even a minimal `BUILD` example **inline** in the README, typically right up at the top. Of course, this is what the ruleset did [previously](https://github.com/bazelbuild/rules_foreign_cc/commit/f471bbfeb21081c63e00611ae88e3e95d4cbc639#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L43). I think out of all of the `bazelbuild` rulesets there's maybe 1-2 which do not follow this pattern, and even these typically link to the stardoc generated docs hosted directly in the repo's `github.io` static site and not externally; something I would imagine would be the first and best choice in this situation.

What's even more curious is that even the [high-level description](https://github.com/bazelbuild/rules_foreign_cc/commit/f471bbfeb21081c63e00611ae88e3e95d4cbc639#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L26) of the ruleset's features has been removed from the README in favor of the generated content hosted externally and buried under a link which is both easy to miss and hosted on an unfamiliar site with unknown uptime/availability, not to mention operated by a for-profit third-party. Not a good look.

[...]

PR #691, opened 2021-06-20
Restored generated docs

UebelAndre:
Outdated docs will render an error message which looks like 
```text
==================== Test output for //:cmake_test:
13a14,16
>
>
>
FAIL: files "stardoc_cmake.md" and "cmake.md" differ. Please run 'cd docs && bazel run :generate_docs' to update documentation
```

closes https://github.com/bazelbuild/rules_foreign_cc/issues/690


UebelAndre:
A patch for `diff_test` used in this PR. It can be removed once https://github.com/bazelbuild/bazel-skylib/pull/307 is closed

alexeagle:
LGTM I agree with @zachgrayio that the repo should stay self-contained like all the other rulesets.

zachgrayio:
Cool. I wonder how difficult it would be to embed some of the generated docs directly into readme so it's front & center but still dynamic and always up to date 🤔   Don't know if anyone has figured out a good workflow there yet?

PR #692, opened 2021-06-21
Added github action to generate docs using mdbook

UebelAndre:
This continues to allow for PRs to avoid committing updated documentation and only requires that the stardoc targets are buildable. Once a change is merged to `main`, a pipeline github action will be run to generate docs using [mdbook](https://github.com/rust-lang/mdBook) and will be published to https://bazelbuild.github.io/rules_foreign_cc/


UebelAndre:
For reviewers, the [github pages](https://bazelbuild.github.io/rules_foreign_cc/) site is already using a rendered version of these docs. Feel free to check it out and let me know what you think.

UebelAndre:
The only thing I don't have a solution for yet is versioned docs. But this doesn't seem like it's common in any other github pages site for other rules.

jsharpe (approved):

UebelAndre:
There are still some more changes I'd like to play around with here but I'm happy you seem to like the general approach. My hope is to generally link to https://bazelbuild.github.io/rules_foreign_cc/main and have https://bazelbuild.github.io/rules_foreign_cc/ be a landing page that just references the different revisions of the docs. Which would contain a link to https://bazelbuild.github.io/rules_foreign_cc/0.2.0 and https://bazelbuild.github.io/rules_foreign_cc/0.3.0

(these links are not active now but I'm hoping they will be one day 😄)

UebelAndre:
It works! We now have something like versioned docs. They're kinda buried but it feels like a start? @jsharpe does this seem like something worth going with?

The root url is basically the navigation page: https://bazelbuild.github.io/rules_foreign_cc
from there you can get to the other docs. And I'll update various github things to point to https://bazelbuild.github.io/rules_foreign_cc/main

jsharpe (approved):
Looks great!

PR #697, opened 2021-06-23
Set progress message when building targets

jheaff1:
This PR modifies the progress message that is printed to console when building targets using `rules_foreign_cc`. Previously, the progress message would be of the (somewhat confusing) form:

```
CcConfigureMakeRule external/openssl/openssl/include
```

This PR modifies the message to be of the form:

```
Building openssl (Foreign Cc - Configure)
```

### DIFF
--- a/foreign_cc/private/framework.bzl
+++ b/foreign_cc/private/framework.bzl
@@ -452,6 +452,10 @@ def cc_external_rule_impl(ctx, attrs):
         use_default_shell_env = "win" in os_name(ctx),
         # this is ignored if use_default_shell_env = True
         env = env,
+        progress_message = "Foreign Cc - {configure_name}: Building {target_name}".format(
+            configure_name = attrs.configure_name,
+            target_name = ctx.attr.name,
+        ),
     )
 
     # Gather runfiles transitively as per the documentation in:


foreign_cc/private/framework.bzl
```diff
@@ -452,6 +452,10 @@ def cc_external_rule_impl(ctx, attrs):
         use_default_shell_env = "win" in os_name(ctx),
         # this is ignored if use_default_shell_env = True
         env = env,
+        progress_message = "Building {target_name} (Foreign Cc - {configure_name})".format(
```
UebelAndre, line 455: Can we actually prefix the progress message with the Foreign Cc tag?
```suggestion
        progress_message = "Foreign Cc - {configure_name}: Building {target_name}".format(
```
I find it easier to see what's going on at a glance this way.

UebelAndre (changes requested):

foreign_cc/private/framework.bzl
jheaff1: > Can we actually prefix the progress message with the Foreign Cc tag?
> 
> I find it easier to see what's going on at a glance this way.

Sure, done

UebelAndre:
Thanks! 🙏

PR #698, opened 2021-06-26
Added cmake 3.20.5 and all other final releases of 3.X

UebelAndre:


jsharpe (changes requested):
These versions also need to be added to the `built_toolchains.bzl` file as it doesn't work to specify a prebuilt version unless the source tar is also added. Ideally the python script will be updated to auto-generate that file too.

UebelAndre:
Good call, I'll add that later today

jsharpe (approved):

PR #701, opened 2021-06-28
Always set use_default_shell_env=True so that action_env is propagated.

### LINKED ISSUE #700: `action_env` fails to propagate to build targets
Unfortunately the code here: https://github.com/bazelbuild/rules_foreign_cc/blob/99ea7e75c2a48cc233ff5e7682c1a31516faa84b/foreign_cc/private/framework.bzl#L300 that attempts to propagate `action_env` fails to do so due to https://github.com/bazelbuild/bazel/issues/4912.

jsharpe:
Fixes #700.


jsharpe:
Note that this PR subtly modifies the precedent of environment variables - those set by the cc_toolchain and the `env` parameter will take precedent over those set via `--action_env`

UebelAndre (approved):

PR #703, opened 2021-06-28
Fix quoting of toolchain flags passed to cmake

jsharpe:
Previously the cmake rule was incorrectly attempting to quote the toolchain flags as they are passed into a HEREDOC in the generated bash script.
This updates the generated bash script to use local variables to pass the flags through to the CMake crosstool file generated by the HEREDOC to avoid having to escape certain characters in those strings. This approach was chosen over disabling expansion in the HEREDOC because it is still desirable to expand some environment variables within the HEREDOC to e.g. access paths for dependencies.

### DIFF
--- a/foreign_cc/private/cmake_script.bzl
+++ b/foreign_cc/private/cmake_script.bzl
@@ -2,6 +2,21 @@
 
 load(":cc_toolchain_util.bzl", "absolutize_path_in_str")
 
+def _escape_dquote_bash(text):
+    """ Escape double quotes in flag lists for use in bash strings that set environment variables """
+
+    # We use a starlark raw string to prevent the need to escape backslashes for starlark as well.
+    return text.replace('"', r'\\\"')
+
+def _escape_dquote_bash_crosstool(text):
+    """ Escape double quotes in flag lists for use in bash strings to be passed to a HEREDOC.
+
+        This requires an additional level of indirection as CMake requires the variables to be escaped inside the crosstool file
+    """
+
+    # We use a starlark raw string to prevent the need to escape backslashes for starlark as well.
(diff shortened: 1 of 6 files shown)


foreign_cc/private/cmake_script.bzl
```diff
@@ -2,6 +2,18 @@
...
 load(":cc_toolchain_util.bzl", "absolutize_path_in_str")
 
+def _escape_dquote_bash(text):
+    """ Escape double quotes in flag lists for use in bash strings to be passed to a HEREDOC. """
+
+    # We use a starlark raw string to prevent the need to escape backslashes for starlark as well.
+    return text.replace('"', r'\\\"')
+
+def _escape_dquote_bash_crosstool(text):
+    """ Escape double quotes in flag lists for use in bash strings to be passed to a HEREDOC. """
```
UebelAndre, line 12: Can you update this docstring to comment on why this is different than `_escape_dquote_bash`?

foreign_cc/private/cmake_script.bzl
```diff
@@ -81,7 +93,7 @@ def create_cmake_script(
         for key in params.env
     ]
     str_cmake_cache_entries = " ".join([
-        "-D{}=\"{}\"".format(key, _escape_dquote_bash(params.cache[key]))
+        "-D{}=\"{}\"".format(key, _escape_dquote_bash_crosstool(params.cache[key]))
```
UebelAndre, line 96: Can you add a comment about what's happening with these `__var_{}` variables here?

UebelAndre (changes requested):
Some small comments but otherwise excited to see this fixed!

UebelAndre (approved):
Awesome! Thank you!

PR #710, opened 2021-07-06
Set CMAKE_INSTALL_LIBDIR

### LINKED ISSUE #518: CMake lib dir selection is not portable
Depending on system the CMake select different output lib directory:

- Centos 8: `/lib64`
- Manjaro 20: `/lib`

I would like to add some bool flag to the `cmake_external` rule, which add an entry: `-DCMAKE_INSTALL_LIBDIR=${out_lib_dir}` to the `cmake_options`, so it should work in the same way on all systems. IMO it should be enabled by default, but I guess such a change break backward compatibility.

What do you think?

jsharpe:
As per the cmake docs: https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html the value of `CMAKE_INSTALL_LIBDIR` can vary depending upon the host OS that cmake is run on. 
Set `CMAKE_INSTALL_LIBDIR` to a deterministic value on all platforms unless the user overrides.

Closes #518

### DIFF
--- a/foreign_cc/private/cmake_script.bzl
+++ b/foreign_cc/private/cmake_script.bzl
@@ -72,8 +72,17 @@ def create_cmake_script(
         "CMAKE_BUILD_TYPE",
         "Debug" if is_debug_mode else "Release",
     )
+
+    # The default for this varies based upon the host OS so explicitly set it
+    # to be a known value unless the user overrides it
+    gnuinstall_libdir = params.cache.get(
+        "CMAKE_INSTALL_LIBDIR",
+        "lib",
+    )
+
     params.cache.update({
         "CMAKE_BUILD_TYPE": build_type,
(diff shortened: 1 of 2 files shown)


UebelAndre (approved):
Thanks for adding this!

jsharpe:
This fails to build libjpeg_turbo which makes use of (a vendored version of) GNUInstallDirs. It seems that behaviour of GNUInstallDirs is a bit strange with respect to `CMAKE_INSTALL_LIBDIR`; it appears to store in its cache the value of prepending `CMAKE_INSTALL_PREFIX` despite the docs suggesting that this should be stored in `CMAKE_INSTALL_FULL_LIBDIR`. 

My feeling at the moment is that we can't automatically set this in the framework and we'll need to leave this to the user to specify because the behaviour of a given build with respect to this variable seems to be inconsistent.

Issue #705, opened 2021-06-30
Potential issue with configure_make vs ffmpeg


HappyCerberus:
I'm struggling to narrow down if this is a problem of `rules_foreign_cc`, but I have trouble integrating `ffmpeg` using the `configure_make` rule.

The configure chokes on this:

`gcc -D_FORTIFY_SOURCE=1 -fstack-protector -fcolor-diagnostics -Wall -Wthread-safety -Wself-assign -fno-omit-frame-pointer -O0 -DDEBUG DEBUG_PREFIX_MAP_PWD=. -isysroot __BAZEL_XCODE_SDKROOT__ -F__BAZEL_XCODE_SDKROOT__/System/Library/Frameworks -F__BAZEL_XCODE_DEVELOPER_DIR__/Platforms/MacOSX.platform/Developer/Library/Frameworks -mmacosx-version-min=11.3 -no-canonical-prefixes -Wno-builtin-macro-redefined -D__DATE__=redacted -D__TIMESTAMP__=redacted -D__TIME__=redacted -target arm64-apple-macosx -c -o /var/folders/rx/1z6cg57d549ffksq5tr8fnvc0000gn/T//ffconf.o7sKUDSO/test.o /var/folders/rx/1z6cg57d549ffksq5tr8fnvc0000gn/T//ffconf.o7sKUDSO/test.c
clang: error: no such file or directory: 'DEBUG_PREFIX_MAP_PWD=.'`

This comes from `ASFLAGS='... -DDEBUG DEBUG_PREFIX_MAP_PWD=. -isysroot...'``

So, not sure if this is Bazel or `rules_foreign_cc`.

I'm on an M1 Mac.

jsharpe:
The issue here looks to be the fact that gcc is being called rather than `wrapped_clang` from the toolchain. The wrapped clang script on macOS has some logic to replace the `DEBUG_PREFIX_MAP_PWD=.` with the correct compiler argument.

jsharpe:
When I've built ffmpeg via `rules_foreign_cc` in the past I've needed to pass the correct compiler name via the `--cc`, `--cxx`, `--dep-cc` and `--host-cc` flags.

HappyCerberus:
Thanks, this is indeed the correct lead. The `ffmpeg` configure seems to be ignoring the `CC`and `CXX`environment variables which is exactly how the sandbox injects the correct path to the wrapper.

[...]

PR #711, opened 2021-07-06
Allow variables set for configure to be referenced as arguments to th…

jsharpe:
…e configure command.

Fixes build issue with #705


jsharpe:
On further thought although this appears to fix the issue, I don't think its quite the right solution.
I think we should be exporting the compiler vars `CC` and `CXX` earlier in the environment and leaving them set for the whole build rather than just setting them for the configure call.

PR #716, opened 2021-07-09
Build make windows

jheaff1:
This pull request fixes the issue of GNU Make failing to build on Windows

### DIFF
--- a/foreign_cc/built_tools/private/built_tools_framework.bzl
+++ b/foreign_cc/built_tools/private/built_tools_framework.bzl
@@ -1,12 +1,18 @@
 """A module defining a common framework for "built_tools" rules"""
 
+load("@bazel_skylib//lib:paths.bzl", "paths")
 load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
 load("//foreign_cc/private:detect_root.bzl", "detect_root")
-load("//foreign_cc/private:framework.bzl", "wrap_outputs")
+load("//foreign_cc/private:framework.bzl", "get_env_prelude", "wrap_outputs")
 load("//foreign_cc/private/framework:helpers.bzl", "convert_shell_script", "shebang")
+load("//foreign_cc/private/framework:platform.bzl", "os_name")
 
 # Common attributes for all built_tool rules
 FOREIGN_CC_BUILT_TOOLS_ATTRS = {
+    "env": attr.string_dict(
+        doc = "Environment variables to set during the build.",
+        default = {},
+    ),
     "srcs": attr.label(
         doc = "The target containing the build tool's sources",
         mandatory = True,
@@ -20,6 +26,11 @@ FOREIGN_CC_BUILT_TOOLS_ATTRS = {
         default = Label("@rules_foreign_cc//foreign_cc/private/framework:platform_info"),
     ),
 }
+
+# Common fragments for all built_tool rules
+FOREIGN_CC_BUILT_TOOLS_FRAGMENTS = [
+    "cpp",
+]
 
 # Common host fragments for all built_tool rules
 FOREIGN_CC_BUILT_TOOLS_HOST_FRAGMENTS = [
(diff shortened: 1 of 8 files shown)


UebelAndre:
Just FYI, I believe we don't have `make` building on windows because we don't have a target running in CI that uses `make`. I'd recommend deleting the [`- "-//make_simple/..."`](https://github.com/bazelbuild/rules_foreign_cc/blob/bb2f0ab0aa7bf1cbcd4836dbfca353f06fd38c41/.bazelci/config.yaml#L110) line from the job definition to enable a target that would start building it:
https://github.com/bazelbuild/rules_foreign_cc/blob/bb2f0ab0aa7bf1cbcd4836dbfca353f06fd38c41/.bazelci/config.yaml#L95-L111

jheaff1:
@UebelAndre  I have re-enabled the `make_simple` test but it seems to be failing. Any ideas?

UebelAndre:
Ugh, these are so annoying:
```
liba.a(liba.o) : error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MT_StaticRelease' doesn't match value 'MD_DynamicRelease' in test_libb.obj
```

I think you need to pass something like:
```
CFLAGS=-MT CXXFLAGS=-MT
```

I've only ever dealt with this in CMake so not quite sure how best to wire this up...

UebelAndre:
Another potential solution may be:
https://github.com/bazelbuild/rules_foreign_cc/blob/bb2f0ab0aa7bf1cbcd4836dbfca353f06fd38c41/examples/cmake_with_data/BUILD.bazel#L7-L12

But you can go either way. Statically liking might be better.

jheaff1:
@UebelAndre  It's passing now, thanks!

UebelAndre:
I wonder how many more tests you can re-enable now that make is buildable 😄

foreign_cc/built_tools/private/built_tools_framework.bzl
```diff
@@ -1,12 +1,17 @@
...
 load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
 load("//foreign_cc/private:detect_root.bzl", "detect_root")
-load("//foreign_cc/private:framework.bzl", "wrap_outputs")
+load("//foreign_cc/private:framework.bzl", "get_env_prelude", "wrap_outputs")
 load("//foreign_cc/private/framework:helpers.bzl", "convert_shell_script", "shebang")
 
 # Common attributes for all built_tool rules
 FOREIGN_CC_BUILT_TOOLS_ATTRS = {
+    "env": attr.string_dict(
+        doc = "Environemnt variables to set during the build.",
```
jsharpe, line 12: Nit: Typo in environment

foreign_cc/built_tools/private/built_tools_framework.bzl
```diff
@@ -44,12 +54,18 @@ def built_tool_rule_impl(ctx, script_lines, out_dir, mnemonic):
     """
 
     root = detect_root(ctx.attr.srcs)
+    lib_name = ctx.attr.name
+    env_prelude = get_env_prelude(ctx, lib_name, [], "")
 
-    script = [
+    cc_toolchain = find_cpp_toolchain(ctx)
+    linker_path_str = "\"" + paths.dirname(cc_toolchain.ld_executable).replace("C:", "/c") + "\""
```
jsharpe, line 61: What if the build is executed from a drive other than C?

jsharpe (commented):
LGTM just a couple of small comments

foreign_cc/built_tools/private/built_tools_framework.bzl
UebelAndre:
```suggestion
        doc = "Environment variables to set during the build.",
```

foreign_cc/built_tools/private/built_tools_framework.bzl
UebelAndre: Would it be possible to use a relative path? If the env prelude from the common framework is being used, you could probably prefix with `$EXT_BUILD_ROOT` or something. But I do think we can't assume `C` drive here.

UebelAndre (changes requested):

jheaff1:
> I wonder how many more tests can be re-enable now that make is buildable 😄
> 
> @jheaff1 Would you mind doing a quick check?

I added some temp commits to check which examples (including the third party examples) the CI could build. I have modified the most recent commit to enable the tests that passed (namely, `cmake_working_dir` and the `curl_build_test`)

foreign_cc/built_tools/private/built_tools_framework.bzl
```diff
@@ -44,12 +55,25 @@ def built_tool_rule_impl(ctx, script_lines, out_dir, mnemonic):
...
     root = detect_root(ctx.attr.srcs)
+    lib_name = ctx.attr.name
+    env_prelude = get_env_prelude(ctx, lib_name, [], "")
+
+    cc_toolchain = find_cpp_toolchain(ctx)
+
+    path_prepend_cmd = ""
+    if "win" in os_name(ctx):
+        # Prepend PATH environment variable with the path to the toolchain linker, which prevents MSYS using its linker (/usr/bin/link.exe) rather than the MSVC linker (both are named "link.exe")
+        linker_path = paths.dirname(cc_toolchain.ld_executable)
```
UebelAndre, line 66: Will it always be the case that this path starts with a drive? I feel like it'd be worth checking for that case but then otherwise doing nothing. Your thoughts?
```suggestion
        linker_path = paths.dirname(cc_toolchain.ld_executable)
        if linker_path[0].is_alpha() and linker_path[1] == ":":
            linker_path = linker_path.replace(linker_path[0:2], "/" + linker_path[0].lower())
```

UebelAndre (changes requested):

foreign_cc/built_tools/private/built_tools_framework.bzl
jheaff1: Sure, done :)

UebelAndre (approved):
This is a huge improvement! This (to me) just unblocked support for an entire platform. Thanks for submitting this!!

jheaff1:
No problem at all, glad I could help :)

PR #718, opened 2021-07-12
Enable more examples tests on windows

UebelAndre:
In enabling additional examples, I noticed two bugs:
1. The `sed` command in [replace_in_files](https://github.com/bazelbuild/rules_foreign_cc/blob/821d1efd24f4605fc96dd7f3f8db31c44d017368/foreign_cc/private/framework/toolchains/windows_commands.bzl#L62-L84) fails with ``sed: -e expression #1, char 162: invalid reference \6 on `s' command's RHS`` or something similar
2. Absolute paths were incorrectly being prefixed with `$EXT_BUILD_ROOT` on Windows.

This PR fixes these and re-enables the examples as regression tests.


jsharpe (approved):
LGTM

PR #721, opened 2021-07-15
Enable openssl builds on windows

UebelAndre:


UebelAndre:
Not quite sure how to go about solving this. I opened https://github.com/openssl/openssl/issues/16087 to try and get some help but if anyone has any ideas I'd love the input!

examples/third_party/openssl/BUILD.openssl.bazel
```diff
@@ -1,56 +1,178 @@
...
+export EXT_BUILD_ROOT="\\$$(normalize "\\$${EXT_BUILD_ROOT}")"
+
+# The configure script needs to be executed with perl. If this is the first
+# argument, pass all commands to perl.
+if [[ \\$$1 == *"Configure" ]]; then
+    "\\$${EXT_BUILD_ROOT}/$(execpath @strawberry_perl//:perl)" \\$$@
+
+    # For some reason, the linker arg '/nologo' gets written out as a
+    # path to some file on disk. This corrects for that
+    sed -i "s@C:/tools/msys64/nologo@/nologo@g" \\$$BUILD_TMPDIR/makefile
```
jheaff1, line 62: I faced this annoying issue also. It’s MSYS2 rewriting `/` to `C:\msys64`. This shouldn’t be happening given that the build script produced by rules_foreign_cc set the `MSYS_NO_PATHCONV` variable - must be a bug with MSYS2. 

To overcome this, I added some code to rules_foreign_cc to rewrite MSVC compiler/linker args so slashes are replaced with dashes, eg `/nologo` with `-nologo`

examples/third_party/openssl/BUILD.openssl.bazel
```diff
@@ -1,56 +1,178 @@
...
+
+    # For some reason no '*.pdb' is created which (may be correct given the
+    # configuration) causes the install target to fail. To fix this the line
+    # is modified
+    # https://github.com/openssl/openssl/issues/947
+    sed -i 's/if "\\$$(SHLIBS)"==""/IF EXIST "ossl_static.pdb"/g' \\$$BUILD_TMPDIR/makefile
```
jheaff1, line 68: I faced this issue also. The root cause is that the MSVC toolchain generated by Bazel adds the `/Z7` compilation flag. However, the for .pdb files to be generated, the compilation flag must be `/Zi`

See https://github.com/jheaff1/rules_foreign_cc/blob/36e298b4f325d447baaf2e692391ef322afe0007/examples/third_party/openssl/BUILD.openssl.bazel#L36

jheaff1:
Please see #729

examples/third_party/openssl/BUILD.openssl.bazel
UebelAndre: One thing I prefer about this approach is that it doesn't override all `CFLAGS`. I think it's better to allow the rules to forward those as-is so the rules build as expected. Perhaps that approach though only leads to more confusion? Particularly if someone builds with a different configuration and it breaks 😓 but I think it should be fine.

UebelAndre:
The implementation here is being deferred to https://github.com/bazelbuild/rules_foreign_cc/pull/729

PR #722, opened 2021-07-18
Added consistent use of `expand_locations` to all rules.

UebelAndre:
There's an [internal macro](https://github.com/bazelbuild/rules_foreign_cc/blob/32e222aeff1220605b80b9c36377db27e7a76555/foreign_cc/private/framework.bzl#L896-L917) for invoking [ctx.expand_location](https://docs.bazel.build/versions/main/skylark/lib/ctx.html#expand_location) and prefixing `execpath` usages with an identifier making it usable as an absolute path in the `ForeignCc` action. This behavior was not consistently used in places where variables are expanded. This PR fixes that to avoid situations where a variable is expanded and then used after the rule has changed directories from the execroot.


foreign_cc/private/framework.bzl
```diff
@@ -893,25 +893,43 @@ def _expand_command_path(binary, path, command):
...
-    expanded_env = dict()
-    for key, value in environ.items():
+    if type(expandable) == type(dict()):
+        expanded_env = dict()
+        for key, value in expandable.items():
+            # If `EXT_BUILD_ROOT` exists in the string, we assume the user has added it themselves
+            if "EXT_BUILD_ROOT" in value:
+                expanded_env.update({key: ctx.expand_location(value, data)})
+            else:
+                expanded_env.update({key: ctx.expand_location(value.replace("$(execpath ", "$EXT_BUILD_ROOT/$(execpath "), data)})
```
jsharpe, line 917: Should we use the `$$VAR$$` form of env variables here? I can't remember the order in which these things get expanded.
```suggestion
                expanded_env.update({key: ctx.expand_location(value.replace("$(execpath ", "$$EXT_BUILD_ROOT$$/$(execpath "), data)})
```

jsharpe (commented):
Just one question - otherwise LGTM

jsharpe (approved):

PR #723, opened 2021-07-18
Updated common `deps` attribute to require `CcInfo` in targets

UebelAndre:
If a target is passed to `deps` that does not contain `CcInfo`, the rules break because of:
https://github.com/bazelbuild/rules_foreign_cc/blob/32e222aeff1220605b80b9c36377db27e7a76555/foreign_cc/private/framework.bzl#L763

I cannot think of a what it would look like to pass something that didn't provide `CcInfo` so my solution is to add this requirement to the attribute and make it an explicit effort to add support for more things. For anyone running into this issue, this error is much clearer.


jsharpe (approved):
LGTM - however this highlights another issue - if you need a non- `CCInfo` dep in `cfg=target` as an input to the build then you could have used `additional_inputs` to pass this dep but this is now marked as deprecated and points towards `build_data` which is in `exec` configuration and so is not equivalent. It appears that `deps` intended to allow additional build dependencies to be made available in `$EXT_BUILD_DEPS` but that obviously doesn't work currently and so no-one is actually using this behaviour via `deps`. 

So should `build_data` really have been in configuration `target` and a `tools` attribute provided in configuration `exec`?

UebelAndre:
> LGTM - however this highlights another issue - if you need a non- `CCInfo` dep in `cfg=target` as an input to the build then you could have used `additional_inputs` to pass this dep but this is now marked as deprecated and points towards `build_data` which is in `exec` configuration and so is not equivalent. It appears that `deps` intended to allow additional build dependencies to be made available in `$EXT_BUILD_DEPS` but that obviously doesn't work currently and so no-one is actually using this behaviour via `deps`.
> 
> So should `build_data` really have been in configuration `target` and a `tools` attribute provided in configuration `exec`?

Is `data` not the `target` cfg attribute? What other things are needed at build time that are in the `target` configuration where `data` and `deps` might not suffice?

jsharpe:
> > LGTM - however this highlights another issue - if you need a non- `CCInfo` dep in `cfg=target` as an input to the build then you could have used `additional_inputs` to pass this dep but this is now marked as deprecated and points towards `build_data` which is in `exec` configuration and so is not equivalent. It appears that `deps` intended to allow additional build dependencies to be made available in `$EXT_BUILD_DEPS` but that obviously doesn't work currently and so no-one is actually using this behaviour via `deps`.
> 
> > 
> 
> > So should `build_data` really have been in configuration `target` and a `tools` attribute provided in configuration `exec`?
> 
> 
> 
> Is `data` not the `target` cfg attribute? What other things are needed at build time that are in the `target` configuration where `data` and `deps` might not suffice?

Yes `data` is but that is also included in `runfiles` by default. You may be compiling a 'fatbin' and so need a dep in `target` configuration but not be a `CcInfo` provider.

UebelAndre:
It sounds like there needs to be some `tools` attribute then? Which would be `build_data` but in exec cfg? Then `data` and `build_data` are in target cfg. I can open a PR for that but think it's unrelated to this change since all this does is change the error message users get from being accidental to intentional.

PR #724, opened 2021-07-18
Added a `make_toolchain`, `nmake` for `msvc` windows platforms

UebelAndre:
The introduction of the new `nmake` toolchain should not impact anyone unless they're using a custom platform which defines the `@bazel_tools//tools/cpp:msvc` constraint.

### DIFF
--- a/foreign_cc/repositories.bzl
+++ b/foreign_cc/repositories.bzl
@@ -50,6 +50,12 @@ def rules_foreign_cc_dependencies(
     if register_default_tools:
         prebuilt_toolchains(cmake_version, ninja_version)
 
+        # `nmake` is a unique toolchin in that it's non-hermetic since it relies on the
+        # host environment but there does not appear to be any way to get `nmake` outside
+        # of installing Visual Studio. It is considered a default toolchain to add better
+        # support for MSVC Windows platforms.
+        native.register_toolchains(str(Label("//toolchains:preinstalled_nmake_toolchain")))
+
     if register_built_tools:
         built_toolchains(
             cmake_version = cmake_version,
--- a/toolchains/BUILD.bazel
+++ b/toolchains/BUILD.bazel
@@ -68,6 +68,21 @@ native_tool_toolchain(
     }),
 )
 
+toolchain(
+    name = "preinstalled_nmake_toolchain",
+    exec_compatible_with = [
+        "@platforms//os:windows",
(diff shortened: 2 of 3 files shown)


jsharpe (approved):

jsharpe (approved):

jheaff1:
Hi @UebelAndre 

It seems the new nmake toolchain requires that the nmake.exe file must be on the PATH? 

Is there away to implement the toolchain so that the MSVC tools are not on the PATH?

FYI I am working on a branch which can build OpenSSL for Windows. I’ve got it working, just need to tidy it up. Nmake is required to build OpenSSL on Windows.

https://github.com/jheaff1/rules_foreign_cc/tree/build_openssl_on_windows

UebelAndre:
There could be a repository rule that does more to find the binary. But it seemed more effort than it's worth since the thing I was hoping for was something hermetic and either buildable or downloadable. Is there an issue with the PATH approach for starters?

And funny you should mention building OpenSSL: https://github.com/bazelbuild/rules_foreign_cc/pull/721

I'll take a look at your branch 😀

jheaff1:
By default, the MSVC tools are not added to the Windows PATH. Also, an MSYS2 shell’s PATH will not contain everything from the host Windows PATH by default. 

Therefore, a developer must either manually add the MSVC tools to the MSYS2 PATH or add the tools to the Windows PATH and launch MSYS2 with the `-use-full-path` flag so that the Windows PATH is automatically appended to the default MSYS2 PATH.

Furthermore, I have found that it’s best to use the `incompatible_strict_action_env` flag with Bazel, otherwise rebuilds occur whenever the PATH variable changes, which can reduce cache hits in a remote cache shared between developers. With this flag set, the MSVC tools will not be on the PATH when actions are run (I think), even if the PATH of the developer’s shell has the MSVC tools on the PATH. Apparently this flag will one day be default in Bazel, which would cause the new `nmake` toolchain to break (I think).

Perhaps the `@local_config_cc` repo contains the path to the MSVC tools. I’ll take a look

jheaff1:
Actually I just realised that `rules_foreign_cc` adds the MSVC tools to the PATH [here](https://github.com/bazelbuild/rules_foreign_cc/blob/main/foreign_cc/private/framework.bzl#L298). Therefore all my concerns given above shouldn't be an issue after all :)

PR #725, opened 2021-07-18
Update `configure_options` to be passed before the `--prefix` arg

UebelAndre:
Some projects (like [openssl](https://github.com/openssl/openssl)) have order specific variables. This allows the attribute to enforce that order.

### DIFF
--- a/foreign_cc/private/configure_script.bzl
+++ b/foreign_cc/private/configure_script.bzl
@@ -70,7 +70,7 @@ def create_configure_script(
             options = " ".join(autoreconf_options),
         ).lstrip())
 
-    script.append("{env_vars} {prefix}\"{configure}\" --prefix=$$BUILD_TMPDIR$$/$$INSTALL_PREFIX$$ {user_options}".format(
+    script.append("{env_vars} {prefix}\"{configure}\" {user_options} --prefix=$$BUILD_TMPDIR$$/$$INSTALL_PREFIX$$".format(
         env_vars = _get_env_vars(workspace_name, tools, flags, env_vars, deps, inputs),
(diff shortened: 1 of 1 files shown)


foreign_cc/private/configure_script.bzl
```diff
@@ -70,7 +70,7 @@ def create_configure_script(
             options = " ".join(autoreconf_options),
         ).lstrip())
 
-    script.append("{env_vars} {prefix}\"{configure}\" --prefix=$$BUILD_TMPDIR$$/$$INSTALL_PREFIX$$ {user_options}".format(
+    script.append("{env_vars} {prefix}\"{configure}\" {user_options} --prefix=$$BUILD_TMPDIR$$/$$INSTALL_PREFIX$$".format(
```
jsharpe, line 73: I wonder if we should provide an optional placeholder to allow the user to add `--prefix` into a specific place in the order of the arguments in case there is a configure script that isn't auto tools that expects it the other way around (or indeed in the middle of a list of options?)
Not every configure script is autotools based - which is the case with openssl (which is perl IIRC)

jsharpe (changes requested):
I think this needs to be more configurable than this - I suspect there will be configure scripts that expect the existing behaviour that this change would break.

I'd suggest a placeholder such as `$(prefix)` or `$$PREFIX_ARG$$` or something that can be added to the arguments to control this behaviour?

UebelAndre:
This turned to not be necessary for what I was looking to do (https://github.com/bazelbuild/rules_foreign_cc/pull/721) so I'll close for now since I can't quite think of a good way to express the ordering of the arguments.

PR #728, opened 2021-07-19
Add missing includes for std::runtime_error

jsharpe:
The compilers in CI are reasonably old and aren't strict at requiring the correct headers included for classes defined in std. This adds the missing include headers that are needed to be standards compliant when using the exception `std::runtime_error`


UebelAndre (approved):
Thanks! Seems fine to me but can you add the context for this to the description? 🙏

PR #729, opened 2021-07-19
Build OpenSSL using MSVC toolchain on Windows

jheaff1:
This PR modifies rules_foreign_cc so that OpenSSL successfully builds using the MSVC toolchain on Windows.

Note that I started working on this branch before the recent commit on master whereby the `nmake` toolchain was added. I instead added an `nmake` attribute to the `configure_make()` rule. It would be good to get the reviewers' thoughts on which approach is best

### DIFF
--- a/examples/third_party/openssl/BUILD.openssl.bazel
+++ b/examples/third_party/openssl/BUILD.openssl.bazel
@@ -2,7 +2,7 @@
 https://github.com/bazelbuild/rules_foreign_cc/issues/337
 """
 
-load("@rules_foreign_cc//foreign_cc:defs.bzl", "configure_make")
+load("@rules_foreign_cc//foreign_cc:defs.bzl", "configure_make", "configure_make_variant")
 
 # Read https://wiki.openssl.org/index.php/Compilation_and_Installation
 
@@ -15,37 +15,74 @@ CONFIGURE_OPTIONS = [
     "no-comp",
     "no-idea",
     "no-weak-ssl-ciphers",
+    "no-shared",
 ]
 
-configure_make(
+LIB_NAME = "openssl"
+
+MAKE_TARGETS = [
+    "build_libs",
+    "install_dev",
+]
+
+alias(
     name = "openssl",
+    actual = select({
+        "@rules_foreign_cc_examples//:msvc_compiler": "openssl_msvc",
+        "//conditions:default": "openssl_default",
+    }),
+    visibility = ["//visibility:public"],
+)
+
+configure_make_variant(
+    name = "openssl_msvc",
+    build_data = [
+        "@nasm//:nasm",
+        "@perl//:perl",
+    ],
+    configure_command = "Configure",
+    configure_in_place = True,
+    configure_options = CONFIGURE_OPTIONS + [
+        "VC-WIN64A",
+        # Unset Microsoft Assembler (MASM) flags set by built-in MSVC toolchain,
+        # as NASM is unsed to build OpenSSL rather than MASM
+        "ASFLAGS=\" \"",
+    ],
+    configure_prefix = "$PERL",
+    env = {
+        # The Zi flag must be set otherwise OpenSSL fails to build due to missing .pdb files
+        "CFLAGS": "-Zi",
+        "PATH": "$(dirname $(execpath @nasm//:nasm)):$PATH",
+        "PERL": "$(execpath @perl//:perl)",
+    },
+    lib_name = LIB_NAME,
+    lib_source = ":all_srcs",
+    out_static_libs = [
+        "libssl.lib",
+        "libcrypto.lib",
+    ],
+    targets = MAKE_TARGETS,
+    toolchain = "@rules_foreign_cc//toolchains:preinstalled_nmake_toolchain",
+)
+
+configure_make(
+    name = "openssl_default",
     configure_command = "config",
     configure_in_place = True,
-    configure_options = select({
-        "@platforms//os:macos": [
-            "ARFLAGS=r",
-            "no-afalgeng",
-            "no-asm",
-            "no-shared",
-        ] + CONFIGURE_OPTIONS,
-        "//conditions:default": [
-            "no-shared",
-        ] + CONFIGURE_OPTIONS,
-    }),
+    configure_options = CONFIGURE_OPTIONS,
     env = select({
         "@platforms//os:macos": {"AR": ""},
         "//conditions:default": {},
     }),
+    lib_name = LIB_NAME,
     lib_source = ":all_srcs",
+    # Note that for Linux builds, libssl must come before libcrypto on the linker command-line.
+    # As such, libssl must be listed before libcrypto
     out_static_libs = [
-        "libcrypto.a",
         "libssl.a",
+        "libcrypto.a",
     ],
-    targets = [
-        "build_libs",
-        "install_dev",
-    ],
-    visibility = ["//visibility:public"],
+    targets = MAKE_TARGETS,
 )
 
 filegroup(
--- a/foreign_cc/private/framework.bzl
+++ b/foreign_cc/private/framework.bzl
@@ -454,9 +454,9 @@ def cc_external_rule_impl(ctx, attrs):
         command = wrapped_outputs.wrapper_script_file.path,
         execution_requirements = execution_requirements,
         use_default_shell_env = True,
-        progress_message = "Foreign Cc - {configure_name}: Building {target_name}".format(
+        progress_message = "Foreign Cc - {configure_name}: Building {lib_name}".format(
             configure_name = attrs.configure_name,
-            target_name = ctx.attr.name,
+            lib_name = lib_name,
         ),
     )
 
--- a/examples/BUILD.bazel
+++ b/examples/BUILD.bazel
@@ -16,6 +16,13 @@ config_setting(
     visibility = ["//visibility:public"],
 )
 
+config_setting(
+    name = "msvc_compiler",
+    flag_values = {
+        "@bazel_tools//tools/cpp:compiler": "msvc-cl",
+    },
+)
+
 test_suite(
     name = "third_party_examples_linux_tests",
     tags = ["manual"],
--- a/examples/third_party/BUILD.bazel
+++ b/examples/third_party/BUILD.bazel
@@ -15,7 +15,7 @@ test_suite(
         "//libjpeg_turbo:libjpeg_turbo_build_test",
         "//libpng:test_libpng",
         "//libssh2:libssh2_build_test",
-        "//openssl:openssl_build_test",
+        "//openssl:openssl_test_suite",
         "//pcre:pcre_build_test",
         "//python:python_tests",
         "//sqlite:sqlite_build_test",
@@ -40,7 +40,7 @@ test_suite(
         "//libjpeg_turbo:libjpeg_turbo_build_test",
         "//libpng:test_libpng",
         "//libssh2:libssh2_build_test",
-        "//openssl:openssl_build_test",
+        "//openssl:openssl_test_suite",
         "//pcre:pcre_build_test",
         "//python:python_tests",
         # Missing a new enough automake to build
@@ -64,7 +64,7 @@ test_suite(
         "//libjpeg_turbo:libjpeg_turbo_build_test",
         "//libpng:test_libpng",
         "//libssh2:libssh2_build_test",
-        "//openssl:openssl_build_test",
+        "//openssl:openssl_test_suite",
         "//pcre:pcre_build_test",
         "//python:python_tests",
         "//sqlite:sqlite_build_test",
@@ -77,7 +77,7 @@ test_suite(
     tags = ["manual"],
     tests = [
         "//curl:curl_build_test",
-        "//openssl:openssl_build_test",
+        "//openssl:openssl_test_suite",
         # TODO: Add more windows tests
     ],
 )
--- a/examples/third_party/BUILD.bazel
+++ b/examples/third_party/BUILD.bazel
@@ -8,7 +8,7 @@ test_suite(
         "//apr_util:apr_util_build_test",
         "//bison:bison_build_test",
         "//cares:test_c_ares",
-        "//curl:curl_build_test",
+        "//curl:curl_test_suite",
         "//gn:gn_launch_test",
         "//gperftools:test",
         "//libgit2:libgit2_build_test",
@@ -32,7 +32,7 @@ test_suite(
         # Missing a new enough m4 to build
         # "//bison:bison_build_test",
         "//cares:test_c_ares",
-        "//curl:curl_build_test",
+        "//curl:curl_test_suite",
         # Attempts to access git sha during configure of build so fails
         # "//gn:gn_launch_test",
         "//gperftools:test",
@@ -56,7 +56,7 @@ test_suite(
         "//apr:apr_build_test",
         "//apr_util:apr_util_build_test",
         "//cares:test_c_ares",
-        "//curl:curl_build_test",
+        "//curl:curl_test_suite",
         "//gn:gn_launch_test",
         "//gperftools:test",
         "//iconv:iconv_build_test",
@@ -76,7 +76,7 @@ test_suite(
     name = "windows_tests",
     tags = ["manual"],
     tests = [
-        "//curl:curl_build_test",
+        "//curl:curl_test_suite",
         "//openssl:openssl_test_suite",
         # TODO: Add more windows tests
     ],
--- a/examples/third_party/BUILD.bazel
+++ b/examples/third_party/BUILD.bazel
@@ -77,6 +77,7 @@ test_suite(
     tags = ["manual"],
     tests = [
         "//curl:curl_build_test",
+        "//openssl:openssl_build_test",
         # TODO: Add more windows tests
     ],
 )
--- a/examples/third_party/curl/BUILD.bazel
+++ b/examples/third_party/curl/BUILD.bazel
@@ -1,4 +1,5 @@
 load("@bazel_skylib//rules:build_test.bzl", "build_test")
+load("@rules_cc//cc:defs.bzl", "cc_test")
 
 exports_files(
     [
@@ -7,10 +8,42 @@ exports_files(
     visibility = ["//visibility:public"],
 )
 
+cc_test(
+    name = "curl_test",
+    srcs = ["curl_test.cc"],
+    defines = select({
+        "@rules_foreign_cc_examples//:msvc_compiler": ["CURL_STATICLIB"],
+        "//conditions:default": [],
+    }),
+    linkopts = select({
+        "@platforms//os:linux": ["-ldl"],
+        "@rules_foreign_cc_examples//:msvc_compiler": [
+            "crypt32.lib",
+            "ws2_32.lib",
+            "advapi32.lib",
+            "user32.lib",
+        ],
+        "//conditions:default": [],
+    }),
+    deps = [
+        "@curl",
+        "@openssl",
+    ],
+)
+
 build_test(
-    name = "curl_build_test",
+    name = "build_test",
     targets = [
         "@curl//:curl",
+    ],
+    visibility = ["//:__pkg__"],
+)
+
+test_suite(
+    name = "curl_test_suite",
+    tests = [
+        ":build_test",
+        ":curl_test",
     ],
     visibility = ["//:__pkg__"],
 )
--- a/examples/third_party/curl/BUILD.curl.bazel
+++ b/examples/third_party/curl/BUILD.curl.bazel
@@ -10,8 +10,10 @@ _CACHE_ENTRIES = {
     "BUILD_SHARED_LIBS": "off",
     "CMAKE_BUILD_TYPE": "RELEASE",
     "CMAKE_PREFIX_PATH": "$$EXT_BUILD_DEPS$$/openssl",
+    "CMAKE_USE_OPENSSL": "on",
     # TODO: ldap should likely be enabled
     "CURL_DISABLE_LDAP": "on",
+    "OPENSSL_ROOT_DIR": "$$EXT_BUILD_DEPS$$/openssl",
 }
 
 _MACOS_CACHE_ENTRIES = dict(_CACHE_ENTRIES.items() + {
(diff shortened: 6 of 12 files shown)


foreign_cc/configure.bzl
```diff
@@ -197,6 +200,13 @@ def _attrs():
             ),
             mandatory = False,
         ),
+        "nmake": attr.bool(
```
UebelAndre, line 203: I don't think `nmake` should be a setting like this. I think (obviously biased since I wrote it this way 😅) it's better to use a toolchain to provide the make binary. In the implementation from https://github.com/bazelbuild/rules_foreign_cc/pull/724 it's true that a binary isn't technically being used but this does stay within the bounds of existing behavior. In discovering that we cannot build OpenSSL without it kinda revealed to me the mixed up state of the toolchains. On Windows, the default toolchain contains tools from MSVC, but we build and use a GNU Make binary. This mix is not supported by OpenSSL since if you target `VC-WIN64A` (which we need to because of the `cc_toolchain` using MSVC), it generates a Makefile which is not compatible with a GNUMake binary. The expectation is to use `nmake` which would have been bundled with the MSVC toolchain. While I'm sure you already noticed this in implementing this PR (thanks BTW!), the thing I'm trying to point out is the mix of a Gnu/MSYS2/mingw64 tools with MSVC tools. My preference is to keep this configuration in a toolchain so users with custom toolchain can define everything in one place without needing to copy/paste boiler plate `select` statements on this (and likely other rules).

I'm not sure if this is where this conversation should be had so up to you but we should pick either this PR or the previous PR 😅

foreign_cc/private/configure_script.bzl
```diff
@@ -173,7 +173,7 @@ _CONFIGURE_TOOLS = {
     "AR": "cxx_linker_static",
     "CC": "cc",
     "CXX": "cxx",
-    # missing: cxx_linker_executable
+    "LD": "cxx_linker_executable",
```
UebelAndre, line 176: Do you know why this was "missing" before? If the implementation is this simple 😅?

UebelAndre (changes requested):

foreign_cc/private/configure_script.bzl
jheaff1: I'm not sure either, although I think this change is the culprit behind the Linux build failures in BuildKite :(

foreign_cc/configure.bzl
jheaff1: That all makes sense. I'm unsure how to tell Bazel to use `nmake` only for certain  Windows builds (i.e. OpenSSL) and use `make` for others.

It is worth bearing in mind that the CLI arguments to `nmake` and `make` are different, therefore there will have to be some Starlark code along the lines of "if the `make` toolchain is GNU Make, have the command as `make -C path/to/dir`. If the `make` toolchain is `nmake`, have the command as `nmake -F path/to/dir/makefile`

UebelAndre: I have a solution for that in my PR. I've solved it with transitions but you'd otherwise need to define a platform which fit the constraints. I used an `msvc` constraint from Bazel so I'd expect it to be available in some existing contexts

foreign_cc/private/configure_script.bzl
jheaff1: Please see new commit I have added to this PR regarding the linker executable

UebelAndre: The commit is kinda hard to read, I'm wondering if the line endings are being changed for the file since all lines have mods but it doesn't look like they have changed.

jheaff1: I just realised that I don't actually need this change for OpenSSL to build. I initially made the change as i was initially building OpenSSL with no configure options. Now that OpenSSL is building with "no-shared", the LD executable is no longer required.

This change seemed to open a can of worms, which i would rather not resolve with this PR as it is a wider issue. The upshot is that it appears rules_foreign_cc cannot build shared libraries.

foreign_cc/configure.bzl
jheaff1: I may be misunderstanding how transitions work but I find the transition you created in your PR a tad confusing.

Before the transition, Bazel is targeting the Windows OS and MSVC compiler. After the transition, Bazel is targeting the Windows OS and the MSVC compiler. 

Perhaps the confusion stems from the fact that typically a toolchain is a variant of a tool suite specific to a platform, eg GNU Make for Linux, or CMake for Windows.

As nmake will only ever be used in conjunction with MSVC on Windows, it’s confusing that an nmake toolchain exists at all. The existence of the toolchain also implies that nmake should be used instead of make whenever MSVC is used, which is not the case.

Hope that made sense, I may just be talking nonsense 😂

UebelAndre: The transition is definitely some funny business 😅 

I'll have to dig up some docs to confirm this but looking at the build logs, the Windows builds are by default using a MSVC toolchain. However, I don't think there are any constraints for this toolchain (maybe aside from `@platforms//os:windows`). This means, we build on a `platform(constraints = ["@platforms//os:windows"])` platform. This is also the case with the GNUMake toolchain. It has no constraints and is expected to match on the current (any) platform. So, since we now have a second `make_toolchain`, how do we handle telling a particular rule to use that toolchain instead. This is done by registering the default toolchain as GNU with no constraints but the `nmake` as `msvc` because that's the only environment `nmake` is expected to exist.

If you run with `--toolchain_resolution_debug` You can see that the `nmake` toolchain is being rejected everywhere except the openssl target which is using the `@openssl_lib//:msvc_platform` platform.
```
INFO: ToolchainResolution: Target platform @local_config_platform//:host: Selected execution platform @local_config_platform//:host, type @rules_foreign_cc//foreign_cc/private/framework:shell_toolchain -> toolchain @rules_foreign_cc_framework_toolchain_windows_commands//:commands, type @bazel_tools//tools/cpp:toolchain_type -> toolchain @local_config_cc//:cc-compiler-x64_windows, type @rules_foreign_cc//toolchains:make_toolchain -> toolchain @rules_foreign_cc//toolchains:built_make
INFO: ToolchainResolution: Target platform @local_config_platform//:host: Selected execution platform @local_config_platform//:host,
INFO: ToolchainResolution:     Type @rules_foreign_cc//toolchains:make_toolchain: execution platform @local_config_platform//:host: Rejected toolchain @rules_foreign_cc//toolchains:preinstalled_nmake; mismatching values: msvc
```
```
INFO: ToolchainResolution: Target platform @local_config_platform//:host: Selected execution platform @openssl_lib//:msvc_platform, type @rules_foreign_cc//foreign_cc/private/framework:shell_toolchain -> toolchain @rules_foreign_cc_framework_toolchain_windows_commands//:commands, type @bazel_tools//tools/cpp:toolchain_type -> toolchain @local_config_cc//:cc-compiler-x64_windows, type @rules_foreign_cc//toolchains:make_toolchain -> toolchain @rules_foreign_cc//toolchains:preinstalled_nmake
```

This allows us to continue to have GNUMake be the default make implementation (which may be something open for discussion) and have OpenSSL use the correct toolchain without impacting other targets. It's the most self-contained thing I could think of.

Does that clarify anything?

jheaff1: That does indeed, than you very much for that. I’ll incorporate your “transition” code into this PR

jheaff1: @UebelAndre As Python is built using configure+GNU make and given that OpenSSL is a dependency of Python, does the OpenSSL transition to nmake mean that another transition will need to be implemented to make Python build with GNU make rather than nmake?

UebelAndre: No, I believe the way it's implemented is that you depend on the `openssl_msvc` target which has an attribute which takes `openssl` in a `msvc_platform` configuration. Then forwards that so no other target needs to know that happened.

UebelAndre:
I'm happy to have this PR be the PR for OpenSSL support but if that's the case, then I'd ask you take the `curl_test.cc` and `openssl_test.cc` files and targets from https://github.com/bazelbuild/rules_foreign_cc/pull/721 as well. They're tiny `cc_test` targets that use `curl` and `openssl` which I thought would be a perfect test of the `openssl` build. We previously only tested that the ForeignCc rule ran successfully. But this would be the first meaningful test for each target. I particularly think that curl is a good one because it tests `openssl` as a transitive dependency which would be unique since I think it's one of the few that have multiple `out_static_lib` files.

But that said, I'm running into linker errors for the tests and could use some help working through them 😅. I'd love to collaborate on them. Let me know if you're interested in picking up these targets.

jheaff1:
Sure, I’ll add the curl and OpenSSL tests to this PR 😁

foreign_cc/configure.bzl
jheaff1: I'm concerned that if we go with the transitions approach, that means that any project that wishes to use rules_foreign_cc to build OpenSSL and its dependants (i.e. Python) must create a proxy repository or similar. This isn't ideal as it enforces complexity onto the client projects of rules_foreign_cc. Is there a way that rules_foreign_cc could absorb most of the complexity and to keep usage of rules_foreign_cc simple?

Not sure if this helps, but I found that the `@bazel_tools//tools/cpp:msvc` config setting wouldn't actually be honoured when building with the MSVC compiler. See [this](https://github.com/bazelbuild/bazel/issues/7730) bazel issue. Instead, using `@bazel_tools//tools/cpp:compiler` as shown [here](https://github.com/bazelbuild/rules_foreign_cc/blob/3756af442f84d9ea0d25820a6f8da6b7aa69c3a6/examples/third_party/openssl/BUILD.openssl.bazel#L9) seemed to work. It seems that the [abseil](https://github.com/abseil/abseil-cpp/blob/master/absl/BUILD.bazel#L28) project does the same. Perhaps using `@bazel_tools//tools/cpp:compiler` rather than `@bazel_tools//tools/cpp:msvc` removes the need to define a new platform and could simplify the selection of nmake?

UebelAndre:
Oh, one other nit, can you move the `nasm` and `perl` repositories inside of `openssl_repositories.bzl`? Unless we're compiling these tools, I don't think they should have their own top level directory. Instead, having them hidden as an implementation detail I think would make things a bit cleaner.

foreign_cc/configure.bzl
UebelAndre: Sure! If something can be done to reduce complexity but also not change the default make toolchain on windows, that'd be ideal! That said though, if it makes sense to change it, I think we totally can but that should be a separate PR.

jheaff1:
> Oh, one other nit, can you move the `nasm` and `perl` repositories inside of `openssl_repositories.bzl`? Unless we're compiling these tools, I don't think they should have their own top level directory. Instead, having them hidden as an implementation detail I think would make things a bit cleaner.

Sure, done

foreign_cc/configure.bzl
```diff
@@ -79,11 +79,14 @@ def _create_configure_script(configureParameters):
     prefix = "{} ".format(ctx.expand_location(attrs.tool_prefix, data)) if attrs.tool_prefix else ""
     configure_prefix = "{} ".format(ctx.expand_location(ctx.attr.configure_prefix, data)) if ctx.attr.configure_prefix else ""
 
+    make_path = "nmake.exe" if ctx.attr.nmake else attrs.make_path
```
UebelAndre, line 82: If we're using transitions this can be ripped out.

foreign_cc/configure.bzl
```diff
@@ -79,11 +79,14 @@ def _create_configure_script(configureParameters):
...
+
     for target in ctx.attr.targets:
         # Configure will have generated sources into `$BUILD_TMPDIR` so make sure we `cd` there
-        make_commands.append("{prefix}{make} -C $$BUILD_TMPDIR$$ {target} {args}".format(
+        make_commands.append("{prefix}{make} {flags} {target} {args}".format(
             prefix = prefix,
-            make = attrs.make_path,
+            make = make_path,
+            flags = "-F $$BUILD_TMPDIR$$/makefile" if ctx.attr.nmake else "-C $$BUILD_TMPDIR$$",
```
UebelAndre, line 89: I think `-F` and `-C` are quite different but `-C` should work in both cases (at least I observed this behavior).

foreign_cc/private/cc_toolchain_util.bzl
```diff
@@ -368,14 +368,25 @@ def get_flags_info(ctx, link_output_file = None):
...
-        assemble = _add_if_needed(flags.assemble, copts),
+        cc = _convert_flags(cc_toolchain_.compiler, _add_if_needed(flags.cc, copts)),
+        cxx = _convert_flags(cc_toolchain_.compiler, _add_if_needed(flags.cxx, cxxopts)),
+        cxx_linker_shared = _convert_flags(cc_toolchain_.compiler, _add_if_needed(flags.cxx_linker_shared, linkopts)),
+        cxx_linker_static = _convert_flags(cc_toolchain_.compiler, flags.cxx_linker_static),
+        cxx_linker_executable = _convert_flags(cc_toolchain_.compiler, _add_if_needed(flags.cxx_linker_executable, linkopts)),
+        assemble = _convert_flags(cc_toolchain_.compiler, _add_if_needed(flags.assemble, copts)),
     )
 
+def _convert_flags(compiler, flags):
```
UebelAndre, line 379: Can you make this an isolated PR? Seems like a great change 😄

foreign_cc/private/configure_script.bzl
```diff
@@ -194,10 +194,15 @@ def _get_configure_variables(workspace_name, tools, flags, user_env_vars):
     for tool in _CONFIGURE_TOOLS:
         tool_value = getattr(tools, _CONFIGURE_TOOLS[tool])
         if tool_value:
-            # Force absolutize of tool paths, which may relative to the workspace
-            # dir (hermetic toolchains) be provided in project repositories
-            # (i.e hermetic toolchains).
-            tools_dict[tool] = [_absolutize(workspace_name, tool_value, True)]
+            # Force absolutize of tool paths, which may relative to the exec root (e.g. hermetic toolchains built from source)
```
UebelAndre, line 197: Can this also be it's own PR? 😅

foreign_cc/private/framework.bzl
```diff
@@ -299,7 +300,13 @@ def get_env_prelude(ctx, lib_name, data_dependencies, target_root):
     env.update(cc_env)
 
     # Add all user defined variables
-    env.update(expand_locations(ctx, ctx.attr.env, data_dependencies))
+    user_vars = expand_locations(ctx, ctx.attr.env, data_dependencies)
```
UebelAndre, line 303: Can this also be it's own PR? 😅

UebelAndre (changes requested):
There are some things that I think should be split into their own PRs but that is something that can wait until everything is working if you want. But this is the same approach I took for my draft and feel splitting out the changes makes it easier to review for everyone.

jheaff1:
@UebelAndre The build is now passing! Those linker errors you mentioned were indeed a real pain but thankfully it's all working now. I'll now address your comments about splitting this work up in to several PRs. :)

examples/third_party/openssl/BUILD.openssl.bazel
```diff
@@ -66,8 +66,8 @@ configure_make(
             "libssl.lib",
         ],
         "//conditions:default": [
```
UebelAndre, line 68: So was this the fix? To swap the ordering here?

jheaff1: Yep, I mentioned this in the commit message

foreign_cc/private/cc_toolchain_util.bzl
jheaff1: Please see #731

foreign_cc/private/configure_script.bzl
jheaff1: Please see #732

examples/third_party/openssl/BUILD.openssl.bazel
UebelAndre: Ah, I should have looked at that... Can you also add a comment saying the ordering there is important?

foreign_cc/private/framework.bzl
jheaff1: Please see #733

examples/third_party/openssl/BUILD.openssl.bazel
jheaff1: Sure no probs

jheaff1:
@UebelAndre  Please see my latest WIP commit which implements a simple `configure_nmake` macro to handle the transition to using the nmake toolchain. If you are happy with it, I can create a separate PR for the   changes

foreign_cc/nmake.bzl
```diff
@@ -0,0 +1,59 @@
...
+    incompatible_use_toolchain_transition = True,
+)
+
+def configure_nmake(name, **kwargs):
+    """ Wrapper macro around configure_make to force usage of the preinstalled nmake toolchain """
+    configure_make_target_name = name + "_nmake"
+
+    configure_make(
+        name = configure_make_target_name,
+        **kwargs
```
jsharpe, line 53: Don't you need some special handling of deps to transition them back to the original `make_toolchain`? Otherwise transitive deps will try to use `nmake` in place of `make`?

foreign_cc/nmake.bzl
```diff
@@ -0,0 +1,59 @@
...
+    )
+    extra_toolchains_transitioned_foreign_cc_target(
+        name = name,
+        extra_toolchains = ["@rules_foreign_cc//toolchains:preinstalled_nmake_toolchain"],
+        target = configure_make_target_name,
```
jsharpe, line 58: I might be wrong here but I think this needs to have `tags` set from `kwargs["tags"]`?

foreign_cc/nmake.bzl
jheaff1: You are right, as I had mistakenly implemented an "outgoing edge" transition. I have now modified it to be an "incoming edge" transition, which according to the [docs](https://docs.bazel.build/versions/main/skylark/config.html#attaching-transitions) should apply to the provided target only

foreign_cc/nmake.bzl
jheaff1: Hmm not sure. Happy to add it although I can't find any documentation online to suggest this is necessary. Thoughts?

jheaff1:
> Doesn't `extra_toolchains` need to be prepended in case the user provides a `make_toolchain` override via `--extra_toolchains`?

good spot! Have actioned

jheaff1:
Inspecting the BuildKite logs, it appears I have uncovered a bug for Windows and MacOS when the the target name is not the same as the repo name. I.e. all examples on the the `main` branch have targets like `@openssl//:openssl`, `@zlib//:zlib` etc. Now that I rearranged `BUILD.openssl.bazel` to have an alias, there seem to be issues with copying/symlinking files

jsharpe:
> Inspecting the BuildKite logs, it appears I have uncovered a bug for Windows and MacOS when the the target name is not the same as the repo name. I.e. all examples on the the `main` branch have targets like `@openssl//:openssl`, `@zlib//:zlib` etc. Now that I rearranged `BUILD.openssl.bazel` to have an alias, there seem to be issues with copying/symlinking files

I think I may have run into this too - I think it puts the outputs into a subdirectory named as the target name instead of in the root of the output dir

jheaff1:
> > Inspecting the BuildKite logs, it appears I have uncovered a bug for Windows and MacOS when the the target name is not the same as the repo name. I.e. all examples on the the `main` branch have targets like `@openssl//:openssl`, `@zlib//:zlib` etc. Now that I rearranged `BUILD.openssl.bazel` to have an alias, there seem to be issues with copying/symlinking files
> 
> I think I may have run into this too - I think it puts the outputs into a subdirectory named as the target name instead of in the root of the output dir

I think there were two problems. The one I mentioned and another where dependants cannot find the library, eg curl cannot find openssl artefacts. The solution to the second problem was to simply add the `lib_name` attribute to the openssl targets. The first problem seems to only affect  me locally and not in BuildKite.

jheaff1:
@UebelAndre @jsharpe BuildKite is happy now (apart from Buildifier). Am I OK to submit a separate PR with addition of the `configure_nmake` macro + related changes?

foreign_cc/nmake.bzl
UebelAndre: You'll definitely want to parse `tags` from `kwargs` and likely apply the `manual` tag to the "hidden" `configure_make` target.

UebelAndre:
> @UebelAndre @jsharpe BuildKite is happy now (apart from Buildifier). Am I OK to submit a separate PR with addition of the `configure_nmake` macro + related changes?

I think you're fine to submit a separate PR but the reason I'd like that is to have a place to have a focused discussion on it. I'm having trouble making a decision on the `configure_nmake` rule and maybe seeing it in isolation will help.

jheaff1:
> > @UebelAndre @jsharpe BuildKite is happy now (apart from Buildifier). Am I OK to submit a separate PR with addition of the `configure_nmake` macro + related changes?
> 
> I think you're fine to submit a separate PR but the reason I'd like that is to have a place to have a focused discussion on it. I'm having trouble making a decision on the `configure_nmake` rule and maybe seeing it in isolation will help.

Sure, please see #734

jheaff1:
This PR is now ready for re-review now that PR #734  has been merged :)

examples/third_party/openssl/BUILD.openssl.bazel
```diff
@@ -15,37 +15,74 @@ CONFIGURE_OPTIONS = [
...
+    configure_options = CONFIGURE_OPTIONS + [
+        "VC-WIN64A",
+        # Unset Microsoft Assembler (MASM) flags set by built-in MSVC toolchain,
+        # as NASM is unsed to build OpenSSL rather than MASM
+        "ASFLAGS=\" \"",
+    ],
+    configure_prefix = "$PERL",
+    env = {
+        # The Zi flag must be set otherwise OpenSSL fails to build due to missing .pdb files
+        "CFLAGS": "-Zi",
```
UebelAndre, line 54: Is there a way to avoid overwriting `CFLAGS` here? I think it'd be nice to allow Bazel to continue to define them.

UebelAndre (changes requested):
Awesome work! One last question but otherwise looks good to me!

examples/third_party/openssl/BUILD.openssl.bazel
jheaff1: The `configure_make` rule will append `-Zi` to the `CFLAGS`, rather than overwrite them :)

UebelAndre (approved):
This is a *huge* step forward for the rules, really proving out the windows side of it. Thank you so much for putting all this together!

jheaff1:
> This is a _huge_ step forward for the rules, really proving out the windows side of it. Thank you so much for putting all this together!

You're very welcome! Thank you for your help

PR #732, opened 2021-07-20
Wrap tool paths that contain whitespaces in quotes

jheaff1:
This change is required when using the MSVC toolchain on Windows,
as paths for tools such as the compiler contain whitespaces (e.g.
C:\Program Files\...)


UebelAndre (approved):
Awesome! Thanks for submitting this!

PR #733, opened 2021-07-20
Prepend user-specified PATH to existing PATH

jheaff1:
If a user specifies a PATH value as part of an `env` attribute, the
value will be prepended to the existing PATH.

An example requirement for this change is that the MSVC build of
OpenSSL requires that the Netwide Assembler (NASM) must be on the
PATH.

### DIFF
--- a/foreign_cc/private/framework.bzl
+++ b/foreign_cc/private/framework.bzl
@@ -103,7 +103,8 @@ CC_EXTERNAL_RULE_ATTRIBUTES = {
             "`$(execpath)` macros may be used to point at files which are listed as `data`, `deps`, or `build_data`, " +
             "but unlike with other rules, these will be replaced with absolute paths to those files, " +
             "because the build does not run in the exec root. " +
-            "No other macros are supported."
+            "No other macros are supported." +
+            "Variables containing `PATH` (e.g. `PATH`, `LD_LIBRARY_PATH`, `CPATH`) entries will be prepended to the existing variable."
         ),
     ),
     "lib_name": attr.string(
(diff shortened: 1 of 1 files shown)


UebelAndre (approved):
Thanks! This seems good to me but I'd love to get @jsharpe to take a look too!

foreign_cc/private/framework.bzl
```diff
@@ -103,7 +103,8 @@ CC_EXTERNAL_RULE_ATTRIBUTES = {
             "`$(execpath)` macros may be used to point at files which are listed as `data`, `deps`, or `build_data`, " +
             "but unlike with other rules, these will be replaced with absolute paths to those files, " +
             "because the build does not run in the exec root. " +
-            "No other macros are supported."
+            "No other macros are supported." +
+            "PATH entries will be prepended to the existing PATH."
```
jsharpe, line 107: nit:
```suggestion
            "`PATH` entries will be prepended to the existing `PATH`."
```

jsharpe (commented):
One small nit in the doc string.

Only comment I have is should we actually be prepending to all keys that exist in the `cc_env` dict and not just `PATH` e.g. `LIBRARY_PATH`, `CPATH`, `LD_LIBRARY_PATH`, `DYLD_LIBRARY_PATH` are all possible other env variables that we might want to preserve? I'd assume that anything that gets set in `cc_env` is required for the proper functioning of the configured `cc_toolchain` and so we should be preserving it.

jheaff1:
> One small nit in the doc string.
> 
> Only comment I have is should we actually be prepending to all keys that exist in the `cc_env` dict and not just `PATH` e.g. `LIBRARY_PATH`, `CPATH`, `LD_LIBRARY_PATH`, `DYLD_LIBRARY_PATH` are all possible other env variables that we might want to preserve? I'd assume that anything that gets set in `cc_env` is required for the proper functioning of the configured `cc_toolchain` and so we should be preserving it.

Actioned

foreign_cc/private/framework.bzl
jheaff1: Actioned

jheaff1:
Can this be merged now? 😁

jsharpe (approved):

jsharpe:
> Can this be merged now? 😁

Done. Thanks!

jheaff1:
Thank you!

PR #734, opened 2021-07-21
Add <rule>_variant macros

jheaff1:
The macros utilise bazel "transitions" to set the `make` toolchain to
be the preinstalled `nmake` tool.

Note that the msvc constraint was removed from the
`exec_compatible_with` attribute of `preinstalled_nmake_toolchain` as
the condition is not actually met even when building with msvc. See
https://github.com/bazelbuild/bazel/issues/7730.

This will be tested in PR #729

### DIFF
--- a/foreign_cc/private/transitions.bzl
+++ b/foreign_cc/private/transitions.bzl
@@ -1,0 +1,69 @@
+"""This file contains rules for configuration transitions"""
+
+load("//foreign_cc:providers.bzl", "ForeignCcDepsInfo")
+
+def _extra_toolchains_transition_impl(settings, attrs):
+    return {"//command_line_option:extra_toolchains": attrs.extra_toolchains + settings["//command_line_option:extra_toolchains"]}
+
+_extra_toolchains_transition = transition(
+    implementation = _extra_toolchains_transition_impl,
+    inputs = ["//command_line_option:extra_toolchains"],
+    outputs = ["//command_line_option:extra_toolchains"],
+)
+
+def _extra_toolchains_transitioned_foreign_cc_target_impl(ctx):
+    # Return the providers from the transitioned foreign_cc target
+    return [
+        ctx.attr.target[DefaultInfo],
+        ctx.attr.target[CcInfo],
+        ctx.attr.target[ForeignCcDepsInfo],
+        ctx.attr.target[OutputGroupInfo],
+    ]
+
+extra_toolchains_transitioned_foreign_cc_target = rule(
+    doc = "A rule for adding extra toolchains to consider when building the given target",
+    implementation = _extra_toolchains_transitioned_foreign_cc_target_impl,
+    cfg = _extra_toolchains_transition,
+    attrs = {
+        "extra_toolchains": attr.string_list(
+            doc = "Additional toolchains to consider",
+            mandatory = True,
+        ),
+        "target": attr.label(
+            doc = "The target to build after considering the extra toolchains",
+            providers = [ForeignCcDepsInfo],
+            mandatory = True,
+        ),
+        "_allowlist_function_transition": attr.label(
+            default = "@bazel_tools//tools/allowlists/function_transition_allowlist",
+        ),
+    },
+    incompatible_use_toolchain_transition = True,
+)
+
+def make_variant(name, rule, toolchain, **kwargs):
+    """ Wrapper macro around foreign cc rules to force usage of the given make variant toolchain.
+
+    Args:
+        name: The target name
+        rule: The foreign cc rule to instantiate, e.g. configure_make
+        toolchain: The desired make variant toolchain to use, e.g. @rules_foreign_cc//toolchains:preinstalled_nmake_toolchain
+        **kwargs: Remaining keyword arguments
+    """
+
+    make_variant_target_name = name + "_"
+
+    tags = kwargs.pop("tags", [])
+
+    rule(
+        name = make_variant_target_name,
+        tags = tags + ["manual"],
+        **kwargs
+    )
+
+    extra_toolchains_transitioned_foreign_cc_target(
+        name = name,
+        extra_toolchains = [toolchain],
+        target = make_variant_target_name,
+        tags = tags,
+    )
--- a/toolchains/BUILD.bazel
+++ b/toolchains/BUILD.bazel
@@ -72,7 +72,6 @@ toolchain(
     name = "preinstalled_nmake_toolchain",
     exec_compatible_with = [
         "@platforms//os:windows",
-        "@bazel_tools//tools/cpp:msvc",
     ],
     toolchain = ":preinstalled_nmake",
     toolchain_type = ":make_toolchain",
--- a/foreign_cc/cmake.bzl
+++ b/foreign_cc/cmake.bzl
@@ -144,6 +144,7 @@ load(
     "create_attrs",
     "expand_locations",
 )
+load("//foreign_cc/private:transitions.bzl", "make_variant")
 load(
     "//foreign_cc/private/framework:platform.bzl",
     "os_name",
@@ -416,3 +417,18 @@ cmake = rule(
     # version is updated to a release of Bazel containing the new default for this setting.
     incompatible_use_toolchain_transition = True,
 )
+
+def cmake_variant(name, toolchain, **kwargs):
+    """ Wrapper macro around the cmake() rule to force usage of the given make variant toolchain.
+
+    Args:
+        name: The target name
+        toolchain: The desired make variant toolchain to use, e.g. @rules_foreign_cc//toolchains:preinstalled_nmake_toolchain
+        **kwargs: Remaining keyword arguments
+    """
+    make_variant(
+        name = name,
+        rule = cmake,
+        toolchain = toolchain,
+        **kwargs
+    )
--- a/foreign_cc/configure.bzl
+++ b/foreign_cc/configure.bzl
@@ -18,6 +18,7 @@ load(
     "create_attrs",
     "expand_locations",
 )
+load("//foreign_cc/private:transitions.bzl", "make_variant")
 load("//foreign_cc/private/framework:platform.bzl", "os_name")
 load("//toolchains/native_tools:tool_access.bzl", "get_make_data")
 
@@ -230,3 +231,18 @@ configure_make = rule(
     # version is updated to a release of Bazel containing the new default for this setting.
     incompatible_use_toolchain_transition = True,
 )
+
+def configure_make_variant(name, toolchain, **kwargs):
+    """ Wrapper macro around the configure_make() rule to force usage of the given make variant toolchain.
+
+    Args:
+        name: The target name
+        toolchain: The desired make variant toolchain to use, e.g. @rules_foreign_cc//toolchains:preinstalled_nmake_toolchain
+        **kwargs: Remaining keyword arguments
+    """
+    make_variant(
+        name = name,
+        rule = configure_make,
+        toolchain = toolchain,
+        **kwargs
+    )
--- a/foreign_cc/defs.bzl
+++ b/foreign_cc/defs.bzl
@@ -1,13 +1,16 @@
 """Public entry point to all Foreign CC rules and supported APIs."""
 
 load(":boost_build.bzl", _boost_build = "boost_build")
-load(":cmake.bzl", _cmake = "cmake")
-load(":configure.bzl", _configure_make = "configure_make")
-load(":make.bzl", _make = "make")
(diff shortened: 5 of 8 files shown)


toolchains/BUILD.bazel
```diff
@@ -72,7 +72,6 @@ toolchain(
     name = "preinstalled_nmake_toolchain",
     exec_compatible_with = [
         "@platforms//os:windows",
-        "@bazel_tools//tools/cpp:msvc",
     ],
     toolchain = ":preinstalled_nmake",
     toolchain_type = ":make_toolchain",
```
jsharpe, line 77: It feels like if its going to be a separate rule then its worth migrating this to a separate toolchain type specific to nmake? That would remove the need for the transition?

jsharpe: ah but then you'd need a separate rule implementation instead of a macro but it should be able to reuse most of the existing `configure_make` rule?

jheaff1: Sure, that would work. I assume @UebelAndre added  the `nmake` variant of the `make` toolchain as `nmake` and `make` are very similar, i.e. they have makefiles

UebelAndre: They're similar but this would also not require anyone to register extra toolchains. Thinking about the `cmake` rule, which takes a `make_toolchain`, I didn't think it should be necessary to have separate code paths for this variant of make. @jsharpe raised the question that maybe even `ninja` should be a `make_toolchain`. I disagree with this since I view `make` and `ninja` to be quite different. Though, from the perspective of `cmake`, they're not really. They're the underlying build tool that's used and it doesn't really care about much more than that. However, `rules_foreign_cc` supports more build systems than `cmake` as first class rules. So I feel we should view each build system as a unique rule with it's own toolchain. From what I've gathered though, `nmake` is just a variant of `gnu make`. There are some differences but is it really considered a separate build tool? To me it feels like a windows specific implementation of an existing tool. But if we feel that it's unique enough, then I'd say the correct path is to promote it to have it's own toolchain. Then yes, I'd say we should add rules for it which share the implementation of the existing `make` and `configure_make` rule.

jheaff1: I'd be happy with either approach but of course you two are the `rules_foreign_cc` experts so I'll leave it up to you to decide and I'll implement what you decide.

jheaff1: To help you guys decide, i have submitted PR #735 which implements the `configure_nmake` rule rather than a macro

UebelAndre: So, I think we should not write a separate rule for `nmake`. I think the `make` and `configure_make` rules should be able to work regardless of any make variant and for any variant we find difficult to fit into the existing rule, then perhaps it should be promoted to have it's own toolchain. But I don't think `nmake` is an example of that. So I'd vote against the other PR

foreign_cc/nmake.bzl
```diff
@@ -0,0 +1,57 @@
+"""Helper rules and macros for building projects using the MSVC nmake build tool"""
+
+load("//foreign_cc:providers.bzl", "ForeignCcDepsInfo")
+load(":configure.bzl", "configure_make")
+
+def _extra_toolchains_transition_impl(settings, attrs):
+    return {"//command_line_option:extra_toolchains": attrs.extra_toolchains + settings["//command_line_option:extra_toolchains"]}
```
UebelAndre, line 7: I'm not too familiar with extra toolchains, can you explain how this can use a toolchain that isn't registered? What's the guarantee that the thing that's returned here is going to be used for the target if another toolchain was matched?

foreign_cc/nmake.bzl
```diff
@@ -0,0 +1,57 @@
...
+    """ Wrapper macro around configure_make to force usage of the preinstalled nmake toolchain """
+    configure_make_target_name = name + "_nmake"
+
+    configure_make(
+        name = configure_make_target_name,
+        **kwargs
+    )
+    extra_toolchains_transitioned_foreign_cc_target(
+        name = name,
+        extra_toolchains = ["@rules_foreign_cc//toolchains:preinstalled_nmake_toolchain"],
```
UebelAndre, line 55: I feel like this should be configurable. Maybe instead of having a dedicated `configure_nmake` target, there's a `configure_make_variant` rule? And you can pass in a toolchain you want the build to use?

The functionality for this should then be moved to `/foreign_cc/private` and setup for `make` and `configure_make`.... what do you think?

UebelAndre (changes requested):

foreign_cc/nmake.bzl
jheaff1: Sure

[`extra_toolchains`](https://docs.bazel.build/versions/main/command-line-reference.html#flag--extra_toolchains) are considered before those registered in a WORKSPACE file (by `register_toolchains`). As such, by applying the transition it is like the target is being built like `bazel build <target> --extra_toolchains="@rules_foreign_cc//toolchains:preinstalled_nmake_toolchain"` and the nmake toolchain will be considered before the toolchains registered in the WORKSPACE file

jheaff1: I personally prefer this approach to transitioning the host platform as I think it is less confusing.

When transitioning the host platform, the transition is from a Windows OS platform targeting MSVC to a Windows OS platform targeting MSVC. Essentially the approach leverages [this](https://github.com/bazelbuild/bazel/issues/7730) issue.

foreign_cc/nmake.bzl
jheaff1: Hmm I could make it more generic, although I'm not aware of any `make` variants other than GNU Make and NMake. If that is unlikely to ever change then perhaps having `configure_make` and `configure_nmake` is clearer and easy to use.

Plus `rules_foreign_cc` allow the registering of custom native toolchains via the [`rules_foreign_cc_dependencies`](https://github.com/bazelbuild/rules_foreign_cc/blob/main/foreign_cc/repositories.bzl#L9) macro. 

If a user must use a `configure_make_variant` and provide a toolchain to use, it will be confusing which toolchain will be used, the one passed in to `configure_make_variant` or the one passed into `rules_foreign_cc_dependencies`?

With `configure_make` and `configure_nmake` rules, this confusion is abstracted away from the user.

Hope that made sense :)

toolchains/BUILD.bazel
jheaff1: @jsharpe If you agree with @UebelAndre I will close PR #735 and address comments on this PR :)

foreign_cc/nmake.bzl
UebelAndre: Looking at [wikipedia](https://en.wikipedia.org/wiki/Make_(software)#Derivatives) there seem to be quite a few. I don't think I see how the explicit `configure_nmake` vs `configure_make_variant` makes a difference in how clear it is what toolchain is being used. The core issue I'm worried about is someone wanting to define a custom `nmake` toolchain but is not able to get the rule to use it because we only transition to add the one built into `rules_foreign_cc`. It could be the case that someone uses [jom](https://wiki.qt.io/Jom) in their project (a derivative of `nmake`) but has no way to get the rules to use it.

jheaff1: Ah ok that all makes sense, I didn't realise how many variants there are.

To clarify, `configure_make` should be replaced with `configure_make_variant`, which includes an attribute that allows  the user to specify the toolchain. Something like:

```
configure_make_variant(
 name = "openssl_nmake"
 variant_toolchain = "@rules_foreign_cc//toolchains:preinstalled_nmake_toolchain"
)

configure_make_variant(
 name = "openssl_gnumake"
 variant_toolchain = "@rules_foreign_cc//toolchains:prebuilt_make_toolchain"
)

or should `configure_make` continue to exist?
```

UebelAndre: Oh, I somehow lost my response....

I think `configure_make` should continue to exists. The new `*_variant` macros should only be considered special case helpers for people who have cross platform dependencies that use weird toolchains. I would say that the way the toolchains are currently setup are incorrect. If Bazel is using a `msvc` compiler, it should probably be using `nmake`. But I think this would be a larger breaking change so to transition things over, the `*_variant` rules help the user explicitly define the toolchain they want.

But just to be clear, I think the goal is to have users define the toolchains the traditional way and the rules should continue to try to support that. Replacing rules which do traditional toolchain management vs this custom solution should be avoided if possible. But I think this is a justifiable case until we can revisit how to manage these toolchains.

What do you think?

jheaff1: Sure, sounds good to me! I’ll try to find some time to implement what you suggest 😁👍🏻

UebelAndre: You're a legend!! 🙌

toolchains/BUILD.bazel
jsharpe: Yes I agree with @UebelAndre lets move forward with this approach.

jheaff1:
@UebelAndre  How does this PR look now? To see how the new `make_variant` macro can be used, please see:

https://github.com/bazelbuild/rules_foreign_cc/blob/b46cec49dc1f7e32f61a9e6adee05e17287b1550/examples/third_party/openssl/BUILD.openssl.bazel#L38-L47

foreign_cc/make_variant.bzl
```diff
@@ -0,0 +1,35 @@
+"""Helper rules and macros for building projects using the variants of GNU Make, such as MSVC's NMake"""
```
UebelAndre, line 1: This may be a lame reason, but I was thinking it'd be nicer to have this new rule in each module so docs for it are rendered into existing pages as add-ons vs making this it's own isolated page. Unless you feel strongly, could this be moved to `foreign_cc/private/transitions.bzl` and have a macro in `make.bzl`, `configure_make.bzl` and perhaps `cmake.bzl` where you have `${rule}_variant` that calls: 
```python
make_variant(name = name, rule = ${rule}, toolchain = toolchain, **kwargs)
```
?

UebelAndre (changes requested):
This looks good to me! I had one nit but the real reason I didn't click approve was to ask if you could update the PR title and description 🙏 😄 

Thanks for working so hard on this!

foreign_cc/make_variant.bzl
jheaff1: Sure, done

jheaff1:
> This looks good to me! I had one nit but the real reason I didn't click approve was to ask if you could update the PR title and description 🙏 😄
> 
> Thanks for working so hard on this!

I've updated the PR title and description 👍

foreign_cc/private/transitions.bzl
```diff
@@ -0,0 +1,69 @@
...
+    },
+    incompatible_use_toolchain_transition = True,
+)
+
+def make_variant(name, rule, toolchain, **kwargs):
+    """ Wrapper macro around foreign cc rules to force usage of the given make variant toolchain.
+
+    Args:
+        name: The target name
+        rule: The configure_make() or make() rule to instantiate
```
jheaff1, line 49: Change comment to not specify `configure_make` and `make`

UebelAndre (approved):
Looks good to me! Thanks for working with me on this and all the hard work you've put in!

Buildifier seems unhappy and would like @jsharpe to take a look too but I'm happy here!

jheaff1:
> Looks good to me! Thanks for working with me on this and all the hard work you've put in!
> 
> Buildifier seems unhappy and would like @jsharpe to take a look too but I'm happy here!

My pleasure! :)

jheaff1:
Hi @jsharpe,

If you are happy with this PR, could you please merge it? 😁

foreign_cc/private/transitions.bzl
```diff
@@ -0,0 +1,69 @@
...
+
+    make_variant_target_name = name + "_"
+
+    tags = kwargs["tags"] if "tags" in kwargs else []
+
+    rule(
+        name = make_variant_target_name,
+        tags = tags + ["manual"],
+        **kwargs
+    )
```
jsharpe, line 62: This logic isn't quite right as is as you can end up with duplicate `tags` passed to `rule`

I think this fixes the logic:
```suggestion
    tags = kwargs.pop("tags", [])

    rule(
        name = make_variant_target_name,
        tags = tags + ["manual"],
        **kwargs
    )
```

jsharpe (changes requested):
Apologies for the delay in getting round to reviewing this.

One small bit of logic to fix up around the handling of `tags` but otherwise I think I'm happy with this.

foreign_cc/private/transitions.bzl
jsharpe: BTW don't use the commit suggestion option as this upsets the CLA bot.

jsharpe (approved):

PR #735, opened 2021-07-21
Add a configure_nmake rule

jheaff1:
Note that the msvc constraint was removed from the
exec_compatible_with attribute of preinstalled_nmake_toolchain as
the condition is not actually met even when building with msvc. See
bazelbuild/bazel#7730.

This will be tested in PR #729

### DIFF
--- a/foreign_cc/configure.bzl
+++ b/foreign_cc/configure.bzl
@@ -19,10 +19,10 @@ load(
     "expand_locations",
 )
 load("//foreign_cc/private/framework:platform.bzl", "os_name")
-load("//toolchains/native_tools:tool_access.bzl", "get_make_data")
+load("//toolchains/native_tools:tool_access.bzl", "get_make_data", "get_nmake_data")
 
-def _configure_make(ctx):
-    make_data = get_make_data(ctx)
+def _configure_make(ctx, tool_name):
+    make_data = get_make_data(ctx) if tool_name == "make" else get_nmake_data(ctx)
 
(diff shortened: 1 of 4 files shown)


UebelAndre (changes requested):
One thing that sucks about having a separate target is that consumers of these rules now need to be aware of the difference when defining library paths and such:
https://github.com/bazelbuild/rules_foreign_cc/blob/c7330faee503cc1b4e7280769480b540fe91cc86/examples/third_party/curl/BUILD.curl.bazel#L12

By having two different rules, there will likely be two different targets and users now have to account for this. Or, if we're to go with this solution, how can we avoid users having to select on each time they have a cross-platform target?

jheaff1:
> One thing that sucks about having a separate target is that consumers of these rules now need to be aware of the difference when defining library paths and such:
> https://github.com/bazelbuild/rules_foreign_cc/blob/c7330faee503cc1b4e7280769480b540fe91cc86/examples/third_party/curl/BUILD.curl.bazel#L12
> 
> By having two different rules, there will likely be two different targets and users now have to account for this. Or, if we're to go with this solution, how can we avoid users having to select on each time they have a cross-platform target?

I came across this exact issue in PR #729. Thankfully the solution was simply setting the `lib_name` attribute of the two `openssl` targets to be `openssl`. Dependants such as curl then need only refer to the `openssl` directory

UebelAndre:
https://github.com/bazelbuild/rules_foreign_cc/pull/734#discussion_r674106039

jheaff1:
Superseded by #734

PR #749, opened 2021-08-01
CC env prelude

### LINKED ISSUE #727: gn example assumes clang++ is available
If `clang++` isn't installed then the build of the gn example fails.

jsharpe:
Always set the environment variables for the toolchain e.g. `CC`, `CXX`, `AS` and the associated flags e.g. `CXXFLAGS`  so that third party build systems can pick up on these by default.

Fixes #727 and #705


jsharpe:
The problem with this PR is that propagating the toolchain flags correctly requires setting flags for the linker which necessitates setting `LD`. This then causes issues with some foreign build systems that expect `LD` to be the c++ compiler but bazel doesn't set the toolchain up in this manner and so this breaks the build.

I'm not sure how to fix this as we don't have control over what `cc_toolchain` has been configured and so I'm not sure how we can setup the correct flags for all possible external build systems that could be invoked via these rules. @UebelAndre do you have any views or ideas on this?

jsharpe:
> any update?

Sorry, I don't currently have the time to revive this PR. Feel free to take it and attempt to fix it up. IIRC the main issue with this was setting a useful value for CXX; since the bazel toolchain sets the compiler to e.g. `gcc` but passes `-x c++` via `CXXFLAGS` then you end up using `gcc` as the driver for compiling C++ code. This mostly works until you get to the link stage as then only `gcc` is called which doesn't know to pull in the std lib for c++ code and you end up with unresolved symbols. I couldn't come up with a way to reliably set all the environment variables to correct values to. make this change work in general.

gou4shi1:
Thanks for your reply, I already have a workaround to compile ffmpeg in bazel, i.e. patch ffmpeg's configure file:
``` diff
diff --git a/configure b/configure
index 6409b94..bcf46d8 100755
--- a/configure
+++ b/configure
@@ -3762,10 +3762,10 @@ libdir_default='${prefix}/lib'
 mandir_default='${prefix}/share/man'
 
 # toolchain
-ar_default="ar"
-cc_default="gcc"
-cxx_default="g++"
-host_cc_default="gcc"
+ar_default="$AR"
+cc_default="$CC"
+cxx_default="$CXX"
+host_cc_default="$CC"
 doxygen_default="doxygen"
 install="install"
 ln_s_default="ln -s -f"
```
Just in case anybody else also meet this issue :)

PR #751, opened 2021-08-02
Re-add setting of PREFIX variable in make invocation

### LINKED ISSUE #736: make install prefix: bug or incorrect documentation
Hi,

The documentation of the `make` rule states the following:

https://github.com/bazelbuild/rules_foreign_cc/blob/b51f25ee6289f72031ae203c5c6ad917fbb3d5d1/foreign_cc/make.bzl#L92

Before [these](https://github.com/bazelbuild/rules_foreign_cc/commit/e4399415b806938ef4179dd9c72f76cdfe102478#diff-4a65571d1835d240c82a2c62d7550f8b537ebe729a68f63f6a9598f2a591997aL52-R63) changes in commit e4399415b806938ef4179dd9c72f76cdfe102478, `make` was indeed invoked with `PREFIX=<install dir>`. Commit e4399415b806938ef4179dd9c72f76cdfe102478 removed the `PREFIX=<install dir>` args to the `make` invocation.

Was this change intentional? If so, the documentation for the `make` rule should be updated. If it wasn't intentional, then the change has introduced a bug and should be resolved. I'm happy to submit a PR to update the documentation or fix the issue, if is indeed an issue :)

jheaff1:
Resolves  #736

### DIFF
--- a/foreign_cc/make.bzl
+++ b/foreign_cc/make.bzl
@@ -56,12 +56,13 @@ def _create_make_script(configureParameters):
     make_commands = []
     prefix = "{} ".format(expand_locations(ctx, attrs.tool_prefix, data)) if attrs.tool_prefix else ""
     for target in ctx.attr.targets:
-        make_commands.append("{prefix}{make} -C $$EXT_BUILD_ROOT$$/{root} {target} {args}".format(
+        make_commands.append("{prefix}{make} -C $$EXT_BUILD_ROOT$$/{root} {target} {args} PREFIX={install_prefix}".format(
             prefix = prefix,
             make = attrs.make_path,
             root = root,
             args = args,
             target = target,
+            install_prefix = ctx.attr.install_prefix,
         ))
 
     return create_make_script(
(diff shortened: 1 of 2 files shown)


foreign_cc/make.bzl
```diff
@@ -56,10 +56,11 @@ def _create_make_script(configureParameters):
     make_commands = []
     prefix = "{} ".format(expand_locations(ctx, attrs.tool_prefix, data)) if attrs.tool_prefix else ""
     for target in ctx.attr.targets:
-        make_commands.append("{prefix}{make} -C $$EXT_BUILD_ROOT$$/{root} {target} {args}".format(
+        make_commands.append("{prefix}{make} -C $$EXT_BUILD_ROOT$$/{root} {target} PREFIX={install_prefix} {args}".format(
```
UebelAndre, line 59:
```suggestion
        make_commands.append("{prefix}{make} -C $$EXT_BUILD_ROOT$$/{root} {target} {args} PREFIX={install_prefix}".format(
```

UebelAndre (approved):
Thanks! One nit but otherwise looks good to me 😄

PR #753, opened 2021-08-05
Add expat dep to apr_util build

### LINKED ISSUE #750: apr_util example missing explicit expat dependency
See release notes from 1.6.0: https://downloads.apache.org/apr/CHANGES-APR-UTIL-1.6.
Expat is a dependency of apr_util and so this should be included in the example as it fails to build if expat development. headers aren't installed in the host system.

jheaff1:
The build of apr_util would previously fail if expat libs and headers were not preinstalled in the build environment.

Resolves #750

### DIFF
--- a/examples/third_party/BUILD.bazel
+++ b/examples/third_party/BUILD.bazel
@@ -28,7 +28,8 @@ test_suite(
     tags = ["manual"],
     tests = [
         "//apr:apr_build_test",
-        "//apr_util:apr_util_build_test",
+        # Copying files to RBE does not preserve timestamps, invoking autotools which is not new enough to build apr_util
+        # "//apr_util:apr_util_build_test",
         # Missing a new enough m4 to build
         # "//bison:bison_build_test",
         "//cares:test_c_ares",
--- a/examples/third_party/apr_util/BUILD.apr_util.bazel
+++ b/examples/third_party/apr_util/BUILD.apr_util.bazel
@@ -10,6 +10,7 @@ configure_make(
     configure_in_place = True,
     configure_options = [
         "--with-apr=$EXT_BUILD_DEPS/apr",
+        "--with-expat=$EXT_BUILD_DEPS/expat",
     ],
     env = select({
         "@platforms//os:macos": {"AR": ""},
@@ -18,5 +19,8 @@ configure_make(
     lib_source = ":all_srcs",
     out_static_libs = ["libaprutil-1.a"],
     visibility = ["//visibility:public"],
-    deps = ["@apr"],
+    deps = [
+        "@apr",
+        "@expat",
+    ],
 )
--- a/examples/third_party/apr_util/BUILD.expat.bazel
+++ b/examples/third_party/apr_util/BUILD.expat.bazel
@@ -1,0 +1,47 @@
+load("@rules_foreign_cc//foreign_cc:defs.bzl", "cmake", "configure_make")
+
+filegroup(
+    name = "all_srcs",
+    srcs = glob(["**"]),
+)
+
+LIB_NAME = "expat"
+
+alias(
+    name = "expat",
+    actual = select({
+        "@rules_foreign_cc_examples//:msvc_compiler": "expat_msvc",
+        "//conditions:default": "expat_default",
+    }),
+    visibility = ["//visibility:public"],
+)
+
+cmake(
+    name = "expat_msvc",
+    cache_entries = {
+        "CMAKE_BUILD_TYPE": "Release",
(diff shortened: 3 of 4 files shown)


jsharpe:
The RBE failures you are seeing is something to do with the shell toolchain file not preserving timestamps correctly when copying over the sources. I've got the same issue trying to get the bison example to build and haven't been able to get to the bottom of the cause of the failure yet.

jheaff1:
> The RBE failures you are seeing is something to do with the shell toolchain file not preserving timestamps correctly when copying over the sources. I've got the same issue trying to get the bison example to build and haven't been able to get to the bottom of the cause of the failure yet.

Yeah i thought it would be due to the timestamps of files not being preserved when copied to the remote executor. I will try downgrading the version of libexpat that does not require too new autotools

jheaff1:
> > The RBE failures you are seeing is something to do with the shell toolchain file not preserving timestamps correctly when copying over the sources. I've got the same issue trying to get the bison example to build and haven't been able to get to the bottom of the cause of the failure yet.
> 
> Yeah i thought it would be due to the timestamps of files not being preserved when copied to the remote executor. I will try downgrading the version of libexpat that does not require too new autotools

So downgrading expat just resorted in other issues. Instead I have used the latest version of expat and removed the apr_util build test for RBE.

Perhaps the RBE executor should be updated to Ubuntu 20.04?

jsharpe:
> > > The RBE failures you are seeing is something to do with the shell toolchain file not preserving timestamps correctly when copying over the sources. I've got the same issue trying to get the bison example to build and haven't been able to get to the bottom of the cause of the failure yet.
> > 
> > 
> > Yeah i thought it would be due to the timestamps of files not being preserved when copied to the remote executor. I will try downgrading the version of libexpat that does not require too new autotools
> 
> So downgrading expat just resulted in other issues. Instead I have used the latest version of expat and removed the apr_util build test for RBE.
> 
> Perhaps the RBE executor should be updated to Ubuntu 20.04?

The RBE executor can only be ubuntu 16.04 - see https://github.com/bazelbuild/continuous-integration/blob/c8bb9e9525edbb3944993eb91dc8d54a62e71b9f/buildkite/bazelci.py#L522

jheaff1:
@jsharpe  ah ok. For now, could we disable the apr_util build for RBE so that this can be merged? It is a stepping stone towards being able to build apr_util on Windows, for which I’d like to submit a PR

examples/third_party/BUILD.bazel
```diff
@@ -28,8 +28,8 @@ test_suite(
     tags = ["manual"],
     tests = [
         "//apr:apr_build_test",
-        "//apr_util:apr_util_build_test",
         # Missing a new enough m4 to build
+        # "//apr_util:apr_util_build_test",
```
jsharpe, line 32: Nit: it's failing to build due to not preserving time stamps rather than a dep on m4 - can you add a comment to this effect?

jsharpe (commented):
I'm happy to take this out of RBE for now so long as @UebelAndre is too, I don't think it's currently testing anything uniquely and there are other packages showing the same issues with time stamps.

UebelAndre:
> I'm happy to take this out of RBE for now so long as @UebelAndre is too, I don't think it's currently testing anything uniquely and there are other packages showing the same issues with time stamps.

I think it's fine, if the project needs `expat` to build we should provide it. And if that caused RBE to fail, we should make sure to investigate that and try to solve it.

jheaff1:
My understating of the issue is as follows:

Projects that use autotools typically provide a “source distribution” with a release, which is a typically a tarball containing the source code and a `configure` file generated by autotools. This allows people to build the project without needing autotools installed in the build environment.

If the timestamp of certain files (eg `configure.in`) is changed, then autotools is invoked to regenerate `configure`.

Before commit 
0f8b8c22f1ac23c4b2da3fd25c52620798fb6dfc, even local builds (ie not RBE) using rules_foreign_cc would invoke autotools as the `copy_dir_contents_to_dir` function in the shell toolchain would not preserve timestamps.

If Bazel does indeed not preserve timestamps when copying files to the remote executor, then I think the solution is for Bazel itself to be modified to preserve timestamps.

examples/third_party/BUILD.bazel
jheaff1: Yep for some reason I mistook “m4” with “autotools”. 

I’m on holiday for a week now but will change the comment when I get back :)

jheaff1: done :)

jsharpe (approved):

PR #759, opened 2021-08-12
Fix autoreconf ".sh" typo

fwingerter-Ocient:
This was misleading previously; `autoreconf.sh` does not exist and the options are used with plain old `autoreconf`.


fwingerter-Ocient:
Never mind. I'm not signing any CLAs. I'll file a bug and someone else can fix this.

PR #762, opened 2021-08-16
Build log4cxx

jheaff1:
This PR adds the build of log4cxx to the CI tests.

This should serve as a regression test for the issue described in #344, as log4cxx has transitive dependencies built by rules_foreign_cc, namely `apr`


examples/third_party/log4cxx/compilation.patch
```diff
@@ -0,0 +1,108 @@
+--- src/main/cpp/inputstreamreader.cpp	2008-03-31 23:34:09.000000000 +0100
```
jheaff1, line 1: Get rid of this file

examples/third_party/log4cxx/BUILD.bazel
```diff
@@ -0,0 +1,16 @@
+load("@bazel_skylib//rules:build_test.bzl", "build_test")
+
+exports_files(
```
UebelAndre, line 3: Can you add the patches to the exports list here?

UebelAndre (changes requested):
Thanks!! Two requests so far.

Can you add a `.md` file or something describing the patches? There seems to be quite a few of them and I think it'd be nice to get an overview and maybe some info about when they might not be needed anymore. But it doesn't have to be extensive. 😄 🙏

jheaff1:
> Thanks!! Two requests so far.
> 
> Can you add a `.md` file or something describing the patches? There seems to be quite a few of them and I think it'd be nice to get an overview and maybe some info about when they might not be needed anymore. But it doesn't have to be extensive. 😄 🙏

I have managed to reduce the number of patches required by adding a `-Wno-narrowing` compiler flag.

I have added comments to `log4cxx_repositories.bzl` for the patches. Let me know if you would like it reworked further :)

UebelAndre (approved):
Looks good to me, Thanks for adding this!

PR #766, opened 2021-08-18
Updated examples to use Visual Studio 2019

UebelAndre:
This also tracks the resolution of https://github.com/bazelbuild/continuous-integration/issues/1204. If this PR turns green then that issue is considered resolved.


homuler:
I'm sorry if you already know how to resolve the build error.
I'm not sure because I can't test it in the same environment, but I think you also need to overwrite `BAZEL_VC`, because the path to VS2017 is set to it now.

> BAZEL_VC=(C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\VC)

jsharpe (approved):

PR #767, opened 2021-08-19
Fix building for iOS

keith:
Some projects like envoy-mobile use rules_foreign_cc to build targeting
iOS, so this is useful to validate on CI as part of the macOS test
build. This also fixes a bug where this was broken since 
c41020e4655fc9adb4bad7e9c04a3f2cf2cc9250 because the wrong SDKROOT was used

### DIFF
--- a/examples/WORKSPACE.bazel
+++ b/examples/WORKSPACE.bazel
@@ -26,7 +26,33 @@ load("//deps:deps_jvm_external.bzl", "deps_jvm_external")
 
 deps_jvm_external()
 
+local_repository(
+    name = "rules_foreign_cc_examples_third_party",
+    path = "third_party",
+)
+
+load("@rules_foreign_cc_examples_third_party//:repositories.bzl", examples_third_party_repositories = "repositories")
+
+examples_third_party_repositories()
+
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "bazel_toolchains",
+    sha256 = "179ec02f809e86abf56356d8898c8bd74069f1bd7c56044050c2cd3d79d0e024",
+    strip_prefix = "bazel-toolchains-4.1.0",
+    urls = [
+        "https://mirror.bazel.build/github.com/bazelbuild/bazel-toolchains/releases/download/4.1.0/bazel-toolchains-4.1.0.tar.gz",
+        "https://github.com/bazelbuild/bazel-toolchains/releases/download/4.1.0/bazel-toolchains-4.1.0.tar.gz",
+    ],
+)
+
+load("@bazel_toolchains//rules:rbe_repo.bzl", "rbe_autoconfig")
+
+# Creates a default toolchain config for RBE.
+# Use this as is if you are using the rbe_ubuntu16_04 container,
+# otherwise refer to RBE docs.
+rbe_autoconfig(name = "buildkite_config")
 
 http_archive(
     name = "build_bazel_rules_apple",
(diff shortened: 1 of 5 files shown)


examples/WORKSPACE.bazel
```diff
@@ -26,6 +26,42 @@ load("//deps:deps_jvm_external.bzl", "deps_jvm_external")
 
 deps_jvm_external()
 
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
```
keith, line 31: I started to do the deps.bzl files like the other things do, but I think that would require I do 4, or pin the versions of each of these deps in the main deps file, which seemed unnecessary. happy to do whatever here though

keith:
So I currently expect this to fail, since compatibility with iOS was broken somewhere in this diff https://github.com/bazelbuild/rules_foreign_cc/compare/d54c78ab86b40770ee19f0949db9d74a831ab9f0...b51f25ee6289f72031ae203c5c6ad917fbb3d5d1 which we discovered when updating envoy and envoy-mobile's usage of the rules.

My assumption is that it was caused by c41020e4655fc9adb4bad7e9c04a3f2cf2cc9250 which always assumes `SDKROOT` is macOS specific. But I haven't been able to fix it yet in my tests even after fixing that. The `-isysroot` arg cmake produces is always pointing to a macOS SDK

keith:
cc @jsharpe from the commit above hoping you have some ideas here

keith:
You can see the current failure here https://buildkite.com/bazel/rules-foreign-cc/builds/3452#a07e12a5-3042-484f-98d2-c60cdbe57a95

keith:
Ok confirmed that was the issue and pushed a fix here. I was thrown off while trying to reduce because of https://github.com/bazelbuild/rules_foreign_cc/pull/769

keith:
I don't see how this could have affected the Linux python3 setup, but I will try to debug more

jsharpe:
> I don't see how this could have affected the Linux python3 setup, but I will try to debug more

Yes, that's an odd failure; I can't see how this change would have affected the RBE python3 build either.

examples/WORKSPACE.bazel
jsharpe: @UebelAndre do you have an opinion on this?

examples/third_party/cares/BUILD.bazel
```diff
@@ -1,8 +1,24 @@
...
     srcs = ["c-ares-test.cpp"],
     visibility = ["//:__pkg__"],
     deps = ["@cares"],
 )
+
+cc_library(
+    name = "ios_lib",
+    srcs = ["c-ares-test.cpp"],
+    tags = ["manual"],
+    deps = ["@cares"],
```
jsharpe, line 15: Wow - this is simpler than I expected - I would have expected us to need to set `CMAKE_SYSTEM_NAME` in the CMake target but clearly CMake has enough logic to autodetect the correct config from the environment?

examples/WORKSPACE.bazel
UebelAndre: I think this is ok but we should then come up with a new strategy for selecting what targets should run on which platforms. Maybe it's time to roll out `exec_compatible_with` or `target_compatible_with` instead of having `examples/third_party` be it's own package. But for the time being, can these dependencies be added to [examples/third_party/WORKSPACE.bazel](https://github.com/bazelbuild/rules_foreign_cc/blob/4ee863c1fc88b554b9b185f192eaadd2e44c011e/examples/third_party/WORKSPACE.bazel) as well?

examples/third_party/cares/BUILD.bazel
keith: Yea the key is that CMake respects `SDKROOT` to determine this I think. It works great even in the complex cases with envoy that has some significant rules using this as well

keith:
My current assumption is that by changing the workspaces here I'm triggering a build that was previously cached, and somehow rotted, maybe through a docker image update

keith:
Ok I take it back, it seems to be a workspace ordering issue. I think because of some transitive dependency on `zlib`

jsharpe (approved):
Thanks for getting to the bottom of this! It's very useful to have more diverse use cases in the tests

keith:
❤️ thanks!

PR #769, opened 2021-08-20
Fix building without sandboxing

keith:
Since 466c32c70f6262f43eac06ad5e9dc2cbecbba228 any changes you made
while testing rules_foreign_cc, or changes to those rules, would not
invalidate the CMakeCache.txt and lead to build issues. It wasn't the
case before that because new temp dirs were used each time.


keith:
cc @attilaolah @scele

foreign_cc/private/framework/toolchains/windows_commands.bzl
```diff
@@ -41,6 +41,9 @@ def disable_tracing():
 def mkdirs(path):
     return "mkdir -p " + path
 
+def rm_rf(path):
+    return "rm -rf " + path
```
keith, line 45: Is this valid for windows? I see some places using this without going through this abstraction so maybe it is?

jsharpe: Yes - its running in a msys2 environment so rm -rf is fine.

jsharpe (approved):
Thanks for finding this issue!

PR #777, opened 2021-08-25
Pass toolchain and user env variables to make invocation

### LINKED ISSUE #776: `make` does not populate environment variables from the toolchain
The documentation of the `make` rule says that it passes

> other environment variables for compilation and linking, taken from Bazel C/C++ toolchain and passed dependencies

to `make`, but that does not seem to be the case at least in version 0.5.1. In fact, the `tools` variable at https://github.com/bazelbuild/rules_foreign_cc/blob/da8952e27c46659866b455de58f5eb3a86185507/foreign_cc/make.bzl#L45 is not consumed by `create_make_script`, whereas it is used by e.g. [`create_configure_script`](https://github.com/bazelbuild/rules_foreign_cc/blob/da8952e27c46659866b455de58f5eb3a86185507/foreign_cc/private/configure_script.bzl#L7).

This is problematic when trying to build `Makefile`-only projects such as [AFL++](https://github.com/AFLplusplus/AFLplusplus). 
Since `configure_make` does not set the enviroment variables for the invocation of the `make_commands` (only for the `configure` script), relying on `configure_make` with an empty `configure` script is not a functional workaround.

fmeum:
Fixes #776.

### DIFF
--- a/examples/make_simple/BUILD.bazel
+++ b/examples/make_simple/BUILD.bazel
@@ -3,9 +3,12 @@ load("@rules_foreign_cc//foreign_cc:defs.bzl", "make")
 
 make(
     name = "make_lib",
-    build_data = ["//make_simple/code:clang_wrapper.sh"],
+    build_data = ["//make_simple/code:cxx_wrapper.sh"],
+    copts = [
+        "-DREQUIRED_DEFINE",
+    ],
     env = {
-        "CLANG_WRAPPER": "$(execpath //make_simple/code:clang_wrapper.sh)",
+        "CXX_WRAPPER": "$(execpath //make_simple/code:cxx_wrapper.sh)",
     },
     lib_source = "//make_simple/code:srcs",
     out_static_libs = ["liba.a"],
@@ -15,6 +18,9 @@ cc_test(
     name = "test_lib",
     srcs = [
         "test_libb.cpp",
+    ],
+    copts = [
+        "-std=c++11",
     ],
     tags = ["windows"],
     deps = [
--- a/foreign_cc/private/configure_vars.bzl
+++ b/foreign_cc/private/configure_vars.bzl
@@ -1,0 +1,137 @@
+"""Helper methods to assembler configure env variables from Bazel information."""
+
+load(":cc_toolchain_util.bzl", "absolutize_path_in_str")
+load(":framework.bzl", "get_foreign_cc_dep")
+
+# buildifier: disable=function-docstring
+def get_configure_vars(
+        workspace_name,
+        tools,
+        flags,
+        user_vars,
+        deps,
+        inputs):
+    vars = _get_configure_variables(workspace_name, tools, flags, user_vars)
+    deps_flags = _define_deps_flags(deps, inputs)
+
+    if "LDFLAGS" in vars.keys():
+        vars["LDFLAGS"] = vars["LDFLAGS"] + deps_flags.libs
+    else:
+        vars["LDFLAGS"] = deps_flags.libs
+
+    # -I flags should be put into preprocessor flags, CPPFLAGS
+    # https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Preset-Output-Variables.html
+    vars["CPPFLAGS"] = deps_flags.flags
+
+    return " ".join(["{}=\"{}\""
+        .format(key, _join_flags_list(workspace_name, vars[key])) for key in vars])
+
+def _define_deps_flags(deps, inputs):
+    # It is very important to keep the order for the linker => put them into list
+    lib_dirs = []
+
+    # Here go libraries built with Bazel
+    gen_dirs_set = {}
+    for lib in inputs.libs:
+        dir_ = lib.dirname
+        if not gen_dirs_set.get(dir_):
+            gen_dirs_set[dir_] = 1
+            lib_dirs.append("-L$$EXT_BUILD_ROOT$$/" + dir_)
+
+    include_dirs_set = {}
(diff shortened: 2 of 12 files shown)


fmeum:
@UebelAndre Would you be able to take a look at this PR? Without it, the `make` can't really be used safely as it does not take Bazel toolchain information into account.

foreign_cc/private/configure_vars.bzl
```diff
@@ -0,0 +1,137 @@
+"""Helper methods to assembler configure env variables from Bazel information."""
```
UebelAndre, line 1: Is the issue in `make` or `configure_make`? I think that between the two rules `configure_make` should generally be an extension of `make` and should try to call functionality from that rule as much as possible. Is all the configuration done here at all specific to `configure_make`?

UebelAndre (changes requested):
Sorry for the delay! One thing I want to clarify just to make sure I understand the change before continuing 😄

foreign_cc/private/configure_vars.bzl
fmeum: The issue is in `make`: `configure_make` correctly sets the usual env variables before invoking the `configure` script (which likely uses these variables to generate the Makefile). `make`, without this PR, never sets any variables. I tried to extract and reuse the common functionality.

UebelAndre: Could the semantics of the file and symbols within it be changed from `configure_make` to `make`? I'm most concerned in understanding what is required to get `make` specifically to work and what things are specific to `configure` or any of the tooling there.

fmeum: We could rename everything here to `env_vars` rather than `configure_vars`. Nothing seems specific to configure, it's just that the variables passed to `make` are the same that are typically used to influence what a `configure` script does. 

Would you want me to carry out the renaming?

UebelAndre: But otherwise, yeah. I think the rename would be good

fmeum: I carried out the rename in https://github.com/bazelbuild/rules_foreign_cc/pull/777/commits/ef8d5b13837c0cce36fd93d627dc73cad74394a3. Could you take a look?

UebelAndre: Fantastic! Thank you!

UebelAndre (changes requested):
Thank you so much for putting this together! The last thing before an approval though, would you be able to add some regression test? The rules as they are now are very prone to regressions so I think it'd be very helpful to have either a unit test in `./tests` or an integration test in `./examples`. Do you have a minimal repro case you'd be willing to share?

fmeum:
> Thank you so much for putting this together! The last thing before an approval though, would you be able to add some regression test? The rules as they are now are very prone to regressions so I think it'd be very helpful to have either a unit test in `./tests` or an integration test in `./examples`. Do you have a minimal repro case you'd be willing to share?

Could such an integration test also consist of a `make` project added to `examples/third_party`? I am making these fixes to build open-source dependencies after all, so they would be natural examples.

UebelAndre:
It depends on the size of the project. We have some fairly large 3rd party examples currently but we feel they're worth always building since they have complex build scripts to exercise the rules and are widely used by users of `rules_foreign_cc`. I'd prefer a simplified example so if there were to be a regression. The iteration is fast and it's clear what the specific requirement is. Vs having to understand a larger project first to understand why a change broke support for it

fmeum:
> It depends on the size of the project. We have some fairly large 3rd party examples currently but we feel they're worth always building since they have complex build scripts to exercise the rules and are widely used by users of `rules_foreign_cc`. I'd prefer a simplified example so if there were to be a regression. The iteration is fast and it's clear what the specific requirement is. Vs having to understand a larger project first to understand why a change broke support for it

I rebased this onto #796 and turned `make_simple` into a more complete integration test verifying the presence of toolchain linkopts and rule copts. However, I am hitting a CI failure on Windows that I don't understand: [cl.exe fails to create liba.o](https://buildkite.com/bazel/rules-foreign-cc/builds/3556#26f2979a-09d2-424d-a78e-135c5baf83ef). Do you have an idea what could be wrong here? If this is a known issue with `cl.exe`, I could reintroduce a wrapper script that uses `clang-cl` instead, I just found the test to be more realistic this way.

examples/make_simple/BUILD.bazel
```diff
@@ -3,10 +3,9 @@ load("@rules_foreign_cc//foreign_cc:defs.bzl", "make")
 
 make(
     name = "make_lib",
-    build_data = ["//make_simple/code:clang_wrapper.sh"],
-    env = {
-        "CLANG_WRAPPER": "$(execpath //make_simple/code:clang_wrapper.sh)",
-    },
```
jsharpe, line 9: IIRC the clang_wrapper.sh script is present here to test the functionality of referencing a build tools from the environment so I'd rather this wasn't removed.

fmeum:
@jsharpe I reintroduced the wrapper and along the way fixed the test failure on Windows.

examples/make_simple/BUILD.bazel
fmeum: I reintroduced the wrapper and along the way fixed the test failure on Windows.

fmeum:
@UebelAndre Could you take another look?

UebelAndre (approved):
Ah! Sorry for the delay again! Thank you so much for putting this together and awesome job on identifying the windows issue 😄

PR #781, opened 2021-09-06
Resolve symlinks among output binaries

fmeum:
Certain (C)Make projects (such as [AFL++](https://github.com/AFLplusplus/AFLplusplus)) emit binaries that are symlinks to other emitted binaries. When built with `rules_foreign_cc`, this can lead to non-deterministic dangling symlink errors since Bazel visits the outputs in an unspecified order. This is fixed by resolving symlinks among the emitted binaries, just like it is already done for libraries.


jsharpe (approved):
LGTM, thanks!

Issue #655, opened 2021-05-21
Add process wrapper


UebelAndre:
These rules should ideally use some process wrapper written in a language other than bash that can orchestrate builds are replace the need for system dependencies like `find`, `sed`, `cat`, `touch`, and more.

This should replace the `foreign_cc` framework.

UebelAndre:
This should still happen

PR #786, opened 2021-09-13
Use rsync with chmod for copying files

### LINKED ISSUE #783: Incompatible with remote exec via bazel-buildfarm
rules_foreign_cc fails on bazel-buildfarm workers. bazel-buildfarm enforces that inputs are read-only and directories are read/execute only. When rules_foreign_cc copies the directory tree, the permissions are preserved and subsequent ./configure commands that attempt to write Makefiles into these read-only directories will fail.

djmarcin:
This ensures that files and directories are writable after being copied, which is necessary in a remote execution environment, such as the one provided by bazel-buildfarm. The existing cp command assumes the source files are writable, so the copies will also be writable, but on bazel-buildfarm, the source files and directories are readonly, so the copies are readonly, and therefore Makefile generation often fails.

The path construction with the trailing `/./` and `-R` flag ensures files are copied with relative paths, which I think is equivalent to what the cp option `--no-target-directory` was accomplishing.

Fixes #783

### DIFF
--- a/foreign_cc/private/framework/toolchains/linux_commands.bzl
+++ b/foreign_cc/private/framework/toolchains/linux_commands.bzl
@@ -86,7 +86,10 @@ fi
     )
 
 def copy_dir_contents_to_dir(source, target):
-    return """cp -L -r --no-target-directory "{source}" "{target}" && find {target} -type f -exec touch -r "{source}" "{{}}" \\;""".format(
+    return """\
+      rsync --chmod=u+w -rRL "{source}/./" "{target}"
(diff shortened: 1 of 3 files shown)


djmarcin:
It looks like rsync isn't available in several of the test environments, unfortunately.

UebelAndre:
Yeah, many of the problems in this project stem from the reliance of bash and assumptions about things installed on the host. I've been wanting to write a process wrapper (https://github.com/bazelbuild/rules_foreign_cc/issues/655) that provides all the functionality needed but have yet to find the time.

PR #789, opened 2021-09-21
Improve failed exports logic

keith:
If something is misconfigured in your setup, these commands can fail.
Example:

```
xcodebuild: error: SDK "iphonesimulator14.4" cannot be located.
xcodebuild: error: SDK "iphonesimulator14.4" cannot be located.
xcrun: error: unable to lookup item 'Path' in SDK 'iphonesimulator14.4
```

In this case, because the execution of the command, and the `export`,
were in the statement, bash ignores the exit code. This breaks that up.

We could also break this up with:

```
foo=$(...)
export foo
```

But that doesn't work in the postprocessing currently done with exports.

### DIFF
--- a/foreign_cc/private/framework.bzl
+++ b/foreign_cc/private/framework.bzl
@@ -292,8 +292,10 @@ def get_env_prelude(ctx, lib_name, data_dependencies, target_root):
         sdk = "{}{}".format(platform.lower(), version)
         env_snippet.extend([
             # TODO: This path needs to take cc_env["XCODE_VERSION_OVERRIDE"] into account
-            "export DEVELOPER_DIR=\"$(xcode-select --print-path)\"",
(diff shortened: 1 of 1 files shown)


UebelAndre:
Nice! Is there some kind of regression test we can add? Or is this dependent upon a particular host (mis)configuration?

keith:
It's dependent on host misconfiguration realistically

keith:
Theoretically we could spin up a specific test job and pass some invalid value and make sure it fails, but probably too much overhead

UebelAndre:
If you could add a comment about why this pattern is important, I'd be happy to get this in 😀

keith:
Updated with a link to the shellcheck details

UebelAndre (approved):
Awesome! Thank you!

keith:
thanks!

PR #796, opened 2021-10-04
Add flags from copts and linkopts attributes

fmeum:
Without this commit, the copts and linkopts attributes on rules such as make and cmake did not actually add any flags to the foreign build system.

This is similar to https://github.com/bazelbuild/rules_foreign_cc/pull/777 and another place where, as far as I can tell, user-supplied flags were silently swallowed. `copts` is being used in https://github.com/bazelbuild/rules_foreign_cc/blob/a7105dafb067c785cc9ade2c70b0d0ce967490fc/foreign_cc/private/cc_toolchain_util.bzl#L224 but that function always returns an empty dictionary for me. @UebelAndre Do you know how this relates to https://github.com/bazelbuild/rules_foreign_cc/blob/a7105dafb067c785cc9ade2c70b0d0ce967490fc/foreign_cc/private/cc_toolchain_util.bzl#L287 which seems much more feature complete?

### DIFF
--- a/foreign_cc/private/cc_toolchain_util.bzl
+++ b/foreign_cc/private/cc_toolchain_util.bzl
@@ -301,9 +301,9 @@ def get_flags_info(ctx, link_output_file = None):
         cc_toolchain = cc_toolchain_,
     )
 
-    copts = (ctx.fragments.cpp.copts + ctx.fragments.cpp.conlyopts) or []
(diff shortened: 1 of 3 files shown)


fmeum:
@UebelAndre I added unittest coverage to this PR. Could you review and merge it before #777? That would make it much easier to add a proper integration test for #777.

UebelAndre (approved):
Very nice! Thank you so much!

PR #803, opened 2021-10-20
Add openssl as a dependency for cmake when building cmake from source

### LINKED ISSUE #709: built cmake is missing dependency on openssl
CMake really needse openssl as a dependency. The current rules to bootstrap cmake don't include this dependency and so this relies on the compilation environment having openssl development headers / libraries available for the build to succeed.

jsharpe:
Update cmake build to include openssl as a dependency.

Closes #709.


UebelAndre:
I think that's an appropriate thing to do. The bigger question is do we value the use of `rules_perl` enough to do that though. I feel like it'd be best to try and build OpenSSL using `rules_cc` if possible but that's a massive undertaking. Right now `rules_perl` does not provide windows support which feels like a blocker there. So if that's true, then we'll likely want to maintain the previous way of building OpenSSL when compiling it for the toolchains here

Issue #720, opened 2021-07-13
Slow `configure_make` builds on windows


UebelAndre:
When running [configure_make](https://bazelbuild.github.io/rules_foreign_cc/0.4.0/configure_make.html) I noticed the builds were unbelievably slow in comparison to linux. This seems to be caused by the [symlink_to_dir](https://github.com/bazelbuild/rules_foreign_cc/blob/821d1efd24f4605fc96dd7f3f8db31c44d017368/foreign_cc/private/framework/toolchains/windows_commands.bzl#L121-L159) function or some other call that walks the directory tree. Windows apparently populates the build directory with some additional content:
```
    Directory: C:\tmp\6ru5hwhq\execroot\rules_foreign_cc_examples\bazel-out\x64_windows-opt-exec-2B5CBBC6\bin\external\rules_foreign_cc\toolchains\make.build_tmpdir


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d-----         7/12/2021   2:15 PM                build-aux
d-----         7/12/2021   2:15 PM                doc
d-----         7/12/2021   2:15 PM                lib
d-----         7/12/2021   2:15 PM                m4
d-----         7/12/2021   2:15 PM                mk
d-----         7/12/2021   2:15 PM                po
d-----         7/12/2021   2:15 PM                src
d-----         7/12/2021   2:15 PM                tests
d-----         7/13/2021   6:57 AM                WinRel
-a----         7/13/2021   6:57 AM          93787 ABOUT-NLS
-a----         7/13/2021   6:57 AM          54579 aclocal.m4
-a----         7/13/2021   6:57 AM           3044 AUTHORS
-ar---         7/12/2021  10:47 AM           5186 Basic.mk
-a----         7/13/2021   6:57 AM            108 BUILD.bazel
-a----         7/13/2021   6:57 AM           1216 build.cfg.in
-a----         7/13/2021   6:57 AM           4531 build.sh
-a----         7/13/2021   6:57 AM           5299 builddos.bat
-a----         7/13/2021   6:57 AM          11011 build_w32.bat
-a----         7/13/2021   6:57 AM         109692 ChangeLog
-a----         7/13/2021   6:57 AM         534917 configure
-a----         7/13/2021   6:57 AM          18566 configure.ac
-a----         7/13/2021   6:57 AM          35147 COPYING
-a----         7/12/2021  10:47 AM           5186 Makefile
[...]

PR #805, opened 2021-10-21
[Ninja] Symlink only top level directory contents

### LINKED ISSUE #804: Symlinking Files on Windows is Slow
I noticed that the `ninja` script calls [symlink_contents_to_dir](https://github.com/bazelbuild/rules_foreign_cc/blob/41f24f3a2f9c7b48db8bcd16d729cff6c26cd115/foreign_cc/ninja.bzl#L54). That function then will walk through the entire directory file per file and symlink one by one (see [declaration](https://github.com/bazelbuild/rules_foreign_cc/blob/4ee863c1fc88b554b9b185f192eaadd2e44c011e/foreign_cc/private/framework/toolchains/windows_commands.bzl#L98)).

I couldn't actually figure out any particular reason for symlinking each file at a time. The one special handling is for a few specific files that are copied instead, where the comments mention it is necessary to copy them to call [replace_in_files](https://github.com/bazelbuild/rules_foreign_cc/blob/4ee863c1fc88b554b9b185f192eaadd2e44c011e/foreign_cc/private/framework/toolchains/windows_commands.bzl#L142). However, `replace_in_files` is used nowhere on the ninja flow.

The issue is that symlinking on windows is _very_ expensive, and anecdotally it can >1hr for symlinking ~25k files on my GCE windows VM with a network SSD. This is likely also the problem behind #720.

My quick hacked local changes show that symlinking the entire folder still "works" (that is, my ninja build invoked through `rules_foreign_cc succeeds). Am I missing the reason why we can't symlink all of them? Should I send in a PR to change that? (At least on windows? Maybe under a flag/attribute?)

rubensf:
Fix #804

### DIFF
--- a/foreign_cc/ninja.bzl
+++ b/foreign_cc/ninja.bzl
@@ -51,7 +51,13 @@ def _create_ninja_script(configureParameters):
     script = []
 
     root = detect_root(ctx.attr.lib_source)
-    script.append("##symlink_contents_to_dir## $$EXT_BUILD_ROOT$$/{} $$BUILD_TMPDIR$$".format(root))
+
+    # On windows, we may get a mix of `\` and `/` path separators.
+    # Not all bash friendly tools know how to handle them.
+    script.append("TMP_EXT=${EXT_BUILD_ROOT//\\\\/\\/}")
+    script.append("TMP_TMP=${BUILD_TMPDIR//\\\\/\\/}")
+    script.append("rm -rf $TMP_TMP/*")
+    script.append("ln -sf $$TMP_EXT$$/{}/* $$TMP_TMP$$".format(root))
 
     data = ctx.attr.data + ctx.attr.build_data
 


UebelAndre:
Sorry for the massive delay here. I unfortunately don't feel confident about these changes or know how to gain confidence here. I personally think `rules_foreign_cc` is bottle-necked by the need for a process wrapper written in some testable language (https://github.com/bazelbuild/rules_foreign_cc/issues/655) before more changes can be made. Otherwise each change seems to bring a higher risk of regression. I'll defer to @jsharpe  for this but think at the very least this change needs to use the `foreign_cc` framework.

jsharpe:
Apologies it has taken me so long to take a look at this PR; I have limited bandwidth at the moment for looking at rules_foreign_cc stuff.

This doesn't fix #720 unfortunately as this change only affects ninja and #720 uses make.

If symlink_to_dir is so slow on windows what happens if we change the implementation to copy instead or symlink on windows? I'll put together a PR to see if this works.
I kind of wanted to see symlink_to_dir modified so that it only links the directories at the highest level it can rather than recursing down and symlinking the files individually.

jsharpe:
@rubensf any change you could take a look at #812 and see if that solves the performance issues for symlinking?

rubensf:
@jsharpe unfortunately the copying is still quite slow - the project we're compiling also checks out some binaries and other dependencies, totally a few gigabytes of data. I updated the comment to remove the #720 fix.

@UebelAndre I'll try to change this in a way that uses the `foreign_cc` framework (but it'll certainly take a while before I get to it :)

PR #812, opened 2021-11-17
Copy instead of symlink on windows

jsharpe:
Testing whether simply copying on windows will resolve the performance issues with symlinking


jsharpe:
This passes CI but windows coverage isn't very complete so this requires further testing ideally.

jheaff1:
I had an issue building OpenSSL with rules_foreign_cc on Windows because some file names were too long and the files could not be copied (windows actually copies files rather than symlinks them if symlinks are not enabled). After enabling symlinks on my windows pc, the problem went away.

I wonder if changing symlinks to copies will cause the issue of long file names to reoccur?

Issue #810, opened 2021-11-17
Apple Silicon can't build hiredis repo, error: ld: symbol(s) not found for architecture x86_64


poinwater:
**Environment**

rules_foreign_cc: 0.6.0
Bazel: 4.2.1
MacOS: Big Sur 11.4
Cmake: 3.21.2
Make: 3.81
Guile: 3.0.7

**WORDSPACE:**
```
workspace(name = "redis")

load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")

http_archive(
    name = "rules_foreign_cc",
    sha256 = "69023642d5781c68911beda769f91fcbc8ca48711db935a75da7f6536b65047f",
    strip_prefix = "rules_foreign_cc-0.6.0",
    url = "https://github.com/bazelbuild/rules_foreign_cc/archive/0.6.0.tar.gz",
)

load("@rules_foreign_cc//foreign_cc:repositories.bzl", "rules_foreign_cc_dependencies")
# This sets up some common toolchains for building targets. For more details, please see
# https://bazelbuild.github.io/rules_foreign_cc/0.6.0/flatten.html#rules_foreign_cc_dependencies
rules_foreign_cc_dependencies()

new_git_repository(
    name = "hiredis",
    branch = "master",
    build_file = "//redis:hiredis.BUILD",
    remote = "https://github.com/redis/hiredis.git",
)

# http_archive(
#     name = "redis-plus-plus",
#     build_file = "//redis:redis-plus-plus.BUILD",
#     sha256 = "3221e1df7bbe95669aecaec1c6c6f4c9a52f472dd4e694e681da951bc738d6bd",
#     strip_prefix = "redis-plus-plus-1.2.3",
#     url = "https://github.com/sewenew/redis-plus-plus/archive/refs/tags/1.2.3.zip",
# )
```

**hiredis.BUILD**
```
load("@rules_foreign_cc//foreign_cc:defs.bzl", "cmake")
# load(
#     "@local_config_tf//:build_defs.bzl",
#     "D_GLIBCXX_USE_CXX11_ABI",
# )

package(
    default_visibility = ["//visibility:public"],
)

licenses(["notice"])  # BSD

filegroup(
    name = "all_srcs",
    srcs = glob(["**"]),
    visibility = ["//visibility:public"],
)

cmake(
    name = "hiredis",
    copts = [
[...]

PR #814, opened 2021-11-18
Explicitly turn off guile support in built version of GNU make

jsharpe:
Might help with the issue seen in #810.


keith (approved):
The error I saw this time when using this with envoy was different, but using this PR seems to have fixed it still.

Issue #755, opened 2021-08-06
autotools toolchain


jsharpe:
We should provide an autotools toolchain providing the common GNU tools, `m4`, `autoconf` and `automake` so that the `configure_make` rule can hermetically consume these components without relying on the host environment having working or new enough versions of these components.

uhthomas:
Hi @jsharpe. What is the status of this issue? Is there any guidance on how best to achieve this?

Thanks :)

jsharpe:
@uhthomas there are toolchain definitions for each of the tools but the rules do not yet make use of them. 
The main issue with completing this is that autoconf is not relocatable in its build - it generates many files with hardcoded paths and so to build a hermetic toolchain of all the autotools components would require heavy patching of autoconf. How far I got is in `examples/third_party/autotools` but I don't currently have much time available to look at this. 

I did experiment with hacking the configure script to pass in `/proc/self/cwd` as part of the install prefix to attempt to fix this but IIRC I ran into some limitations with resolving symlinks in this approach.

My current advise on this is to execute your build within a container with a known deployed toolchain to make this as hermetic as possible.

PRs are welcome though to move towards this goal!

uhthomas:
Not necessarily a "small" example, but as far as I understand capnproto is using autoreconf. Trying to get this working with rules_foreign_cc at the moment and struggling a bit.

https://github.com/capnproto/capnproto/tree/281aacc18003ae103b8a04ea83d7b6f814c97b8b/c%2B%2B

jsharpe:
Its also possible that running autoreconf should be the default behaviour - Debian recommend this: https://wiki.debian.org/Autoreconf#Autoreconfing_packages_-_A_Maintainers.27_Guide.

I guess that the only issue is that the configure_make rules aren't necessarily explicitly an autotools project. Maybe we should have an autotools specific rule to distinguish between autotools specific configure_make projects and more generic ones that just follow that pattern e.g. openssl.

uhthomas:
Sounds reasonable to me. Do you have a recommended path forward for this? As mentioned before, I'd be happy to help where possible. Though my knowledge of the C/C++ toolchains aren't too strong.

jsharpe:
As far as I'm aware all the code paths should already be in place for autoreconf so likely what I'm suggesting is just a macro around the existing rule with some appropriate defaults set

uhthomas:
Cool!

For what it's worth I tried building with the autotools example and didn't have too much luck.

[...]

PR #816, opened 2021-11-19
Add toolchain types for autotools tools

jsharpe:
Add the toolchain types and resolve these to the preinstalled versions of such tools.

Note this only sets up the toolchains and doesn't actually make use of them yet.


jsharpe:
Working towards #755

UebelAndre (approved):

UebelAndre:
It might be worth just adding these when the toolchains are actually ready for use. But your call 😄

PR #818, opened 2021-11-20
Let make build in BUILD_TMPDIR

### LINKED ISSUE #797: `make` builds fail to land in `$$BUILD_TMPDIR$$`
I am not yet sure why, but some of my `make` builds of third-party libraries fail due to the use of `$$EXT_BUILD_ROOT$$/{root}` instead of `$$BUILD_TMPDIR$$` in:
https://github.com/bazelbuild/rules_foreign_cc/blob/a7105dafb067c785cc9ade2c70b0d0ce967490fc/foreign_cc/make.bzl#L59
All of my builds pass with `$$BUILD_TMPDIR$$` in that spot, which is also the value used by `configure_make`.
Edit: There are some issues with dangling symlinks with that change in other builds.

@UebelAndre Could you briefly explain why some rules use `$$EXT_BUILD_ROOT$$/{root}` and others use `$$BUILD_TMPDIR$$`? If I understand this a bit better, I could probably start working on a PR that fixes the build issues I am seeing without introducing new ones.

fmeum:
configure_make already used BUILD_TMPDIR whereas make did not.

Fixes #797.

### DIFF
--- a/foreign_cc/make.bzl
+++ b/foreign_cc/make.bzl
@@ -56,10 +56,9 @@ def _create_make_script(configureParameters):
     make_commands = []
     prefix = "{} ".format(expand_locations(ctx, attrs.tool_prefix, data)) if attrs.tool_prefix else ""
     for target in ctx.attr.targets:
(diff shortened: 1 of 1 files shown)


jsharpe (approved):
LGTM

PR #819, opened 2021-11-22
Make libtool usable as AR in configure_make

fmeum:


foreign_cc/private/configure_script.bzl
```diff
@@ -211,6 +211,11 @@ def _get_configure_variables(workspace_name, tools, flags, user_env_vars):
         if toolchain_val:
             tools_dict[user_var] = [user_env_vars[user_var]]
 
+    if tools_dict["AR"][0].endswith("/libtool"):
+        arflags = vars.get("ARFLAGS")
+        if arflags and arflags[-1] != "-o":
+            vars["ARFLAGS"] = vars["ARFLAGS"] + ["-o"]
```
jsharpe, line 217: Do you not need to set ARFLAGS if its not already defined?

This logic isn't quite right but you get the general idea: 
```suggestion
        if arflags and arflags[-1] != "-o":
            vars["ARFLAGS"] = vars["ARFLAGS"] + ["-o"]
        else:
            vars["ARFLAGS"] = ["-o"]
```

fmeum: Good catch. However, I found some configure scripts (e.g. for expat) to simply execute `$AR cru foo.a foo.o`, so this approach will likely not work. 

I think the proper solution would use a libtool wrapper script with a structure similar to https://github.com/bazelbuild/bazel/blob/37710728712e00e8bd7145662668d72591204146/tools/objc/libtool.sh#L74. @jsharpe Could you give me a hint on how to include an additional tool into the build? Would it be fine to add a shell script to `framework.bzl` as an additional `tools` entry for every build or is there a better way?

fmeum: Thinking more about it, maybe generating a wrapper on the fly (which requires access to `ctx`) would be more generically useful. This could also be used to generate a simple compiler wrapper to properly handle `--sysroot` for configure/make builds. Do you see a good way to add something like this to the current framework structure?

PR #826, opened 2021-12-02
Support propagation of `includes`

ericastor:
This should help support foreign inclusion of libraries that export their headers at multiple locations.

In particular, this should help with libraries where the headers reference each other by locations that don't match the only valid `out_include_dir`.


jsharpe (approved):
LGTM

PR #830, opened 2021-12-04
Expand locations and make variables for `cmake::cache_entries`

UebelAndre:
I was attempting to use [find_package](https://cmake.org/cmake/help/v3.22/command/find_package.html) by setting `<project_name>_DIR` in `cmake::cache_entries` and realized the locations weren't being expanded.

eg:
```python
cmake(
    name = "project",
    cache_entries = {
        "my_module_DIR": "$$(dirname $(execpath @my_module//:module.cmake))",
    },
    data = ["@my_module//:module.cmake"],
    lib_source = ":all_srcs",
)
```

The above snippet did not work without the changes in this PR. I'm happy to use alternative approaches but unfortunately couldn't find any.

### DIFF
--- a/foreign_cc/cmake.bzl
+++ b/foreign_cc/cmake.bzl
@@ -262,7 +262,7 @@ def _create_configure_script(configureParameters):
         install_prefix = "$$INSTALLDIR$$",
         root = root,
         no_toolchain_file = no_toolchain_file,
-        user_cache = dict(ctx.attr.cache_entries),
+        user_cache = expand_locations(ctx, ctx.attr.cache_entries, data),
         user_env = expand_locations_and_make_variables(ctx, "env", data),
         options = attrs.generate_args,
         cmake_commands = cmake_commands,


UebelAndre:
Unfortunately, this would be a breaking change
```
ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/90e2e9b2d72fa9b00c19ff0535bb31ae/external/pcre/BUILD.bazel:12:6: in cache_entries attribute of cmake rule @pcre//:pcre: '${CMAKE_C_FLAGS:-}' syntax is not supported; use '$(CMAKE_C_FLAGS:-)' instead for "Make" variables, or escape the '$' as '$$' if you intended this for the shell
```

jsharpe (approved):
LGTM

jsharpe:
> Unfortunately, this would be a breaking change
> 
> ```
> ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/90e2e9b2d72fa9b00c19ff0535bb31ae/external/pcre/BUILD.bazel:12:6: in cache_entries attribute of cmake rule @pcre//:pcre: '${CMAKE_C_FLAGS:-}' syntax is not supported; use '$(CMAKE_C_FLAGS:-)' instead for "Make" variables, or escape the '$' as '$$' if you intended this for the shell
> ```

I don't think its unreasonable to require use of shell variables / expansions to be escaped with `$$` as this is already required in a number of other locations in the rules so I think this makes it more consistent anyway.

allsey87:
I don't think what was being attempted here would work:
```
"my_module_DIR": "$$(dirname $(execpath @my_module//:module.cmake))",
```

`dirname` is part of `coreutils` which is not only limited to Unix based system but is also not covered by Bazel's make variable expansions.

PR #831, opened 2021-12-04
Two part install in configure_make rules

jsharpe:
Hopefully a stepping stone towards getting a working build of autoconf

### DIFF
--- a/examples/third_party/BUILD.bazel
+++ b/examples/third_party/BUILD.bazel
@@ -6,10 +6,8 @@ test_suite(
     tests = [
         "//apr:apr_build_test",
         "//apr_util:apr_util_build_test",
-        # Builds but not useful as it hardcodes paths so isn't relocateable
-        # "//autotools:autoconf_build_test",
-        # Fails due to hardcoded paths in autoconf
-        # "//autotools:automake_build_test",
+        "//autotools:autoconf_build_test",
+        "//autotools:automake_build_test",
         "//autotools:libtool_build_test",
         "//autotools:m4_build_test",
         "//bison:bison_build_test",
@@ -17,8 +15,8 @@ test_suite(
         "//curl:curl_test_suite",
         "//gn:gn_launch_test",
         "//gperftools:test",
-        "//libgit2:libgit2_build_test",
         "//iconv:iconv_linux_build_test",
+        "//libgit2:libgit2_build_test",
         "//libjpeg_turbo:libjpeg_turbo_build_test",
         "//libpng:test_libpng",
         "//libssh2:libssh2_build_test",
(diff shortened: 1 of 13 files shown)


foreign_cc/configure.bzl
```diff
@@ -207,6 +223,13 @@ def _attrs():
             mandatory = False,
             default = ["", "install"],
         ),
+        "two_part_install": attr.bool(
+            doc = (
+                "Set to True if the configure make script supports two phase install via DESTDIR. (See https://www.gnu.org/software/automake/manual/automake.html#Two_002dPart-Install). Using this option will result in more reproducible builds as --prefix will be set to a consistent value at configure time."
+            ),
+            mandatory = False,
+            default = False,
```
jsharpe, line 231: I would like to be able to set this as the default behaviour as this will produce more hermetic builds but it is a breaking change

foreign_cc/configure.bzl
```diff
@@ -60,12 +63,31 @@ def _create_configure_script(configureParameters):
     attrs = configureParameters.attrs
     inputs = configureParameters.inputs
 
-    install_prefix = _get_install_prefix(ctx)
+    if ctx.attr.install_prefix:
+        if ctx.attr.install_prefix[0] == "/":
+            install_prefix = ctx.attr.install_prefix
```
jsharpe, line 68: Setting prefix seems to cause issues with the output directories and the paths it searches, need to investigate this properly

examples/third_party/autotools/BUILD.automake.bazel
```diff
@@ -9,20 +9,20 @@ filegroup(
 
 configure_make(
     name = "automake",
-    build_data = [
-        "@autoconf",
-        "@m4",
-    ],
     env = {
-        "AC_MACRODIR": "$$EXT_BUILD_DEPS$$/bin/autoconf/share/autoconf",
```
jsharpe, line 17: The fact that this style of env bar doesn't work anymore seems like a regression..

examples/third_party/BUILD.bazel
```diff
@@ -6,19 +6,17 @@ test_suite(
     tests = [
         "//apr:apr_build_test",
         "//apr_util:apr_util_build_test",
-        # Builds but not useful as it hardcodes paths so isn't relocateable
-        # "//autotools:autoconf_build_test",
-        # Fails due to hardcoded paths in autoconf
-        # "//autotools:automake_build_test",
+        "//autotools:autoconf_build_test",
+        "//autotools:automake_build_test",
```
jsharpe, line 10: Doesn't pass yet as autoconf prefix isn't set correctly yet.

PR #834, opened 2021-12-06
Adapt tests to new cc_binary file extension on macOS

fmeum:
As of https://github.com/bazelbuild/bazel/pull/14369, shared libraries produced with cc_binary on macOS with the auto-configured toolchain use the correct file extension for dynamic libraries (`.dylib` rather than `.so`). This requires adapting a test.


UebelAndre (approved):
Good to know! Thanks for fixing this 😄

PR #842, opened 2021-12-19
Bash quoting

jsharpe:


jsharpe:
Not sure why this causes a failure with building python2 on RBE

jsharpe:
> Not sure why this causes a failure with building python2 on RBE

I think this was a bug in the python2 build system - the bump of python2 has now fixed this. @UebelAndre PTAL.

UebelAndre (approved):

PR #843, opened 2021-12-19
Add current_*_toolchain rules to allow passing current toolchain inst…

jsharpe:
…ances to toolchains attribute of rules

Also adds ability to set an environment to be set when using a toolchain instance. These variables are made available as MAKEVARs when used on the toolchains attribute of rules too.


UebelAndre (approved):
Thanks!

PR #846, opened 2021-12-23
use rules_perl toolchain for Windows build of OpenSSL

jheaff1:


examples/third_party/openssl/openssl_repositories.bzl
```diff
@@ -28,25 +28,3 @@ def openssl_repositories():
             "https://www.nasm.us/pub/nasm/releasebuilds/2.15.05/win64/nasm-2.15.05-win64.zip",
         ],
     )
-    maybe(
```
UebelAndre, line 31: Could you leave `rules_perl` as a dependency here? I think unless we have another target that requires it, it should stay grouped with `openssl`

UebelAndre (changes requested):
Thanks! Just one small request 😄

examples/third_party/openssl/openssl_repositories.bzl
jheaff1: Thanks for the review 😁

I moved the declaration of rules perl because autotools also requires it.

please see https://github.com/bazelbuild/rules_foreign_cc/blob/e3a12c7d0f4df8d9805cd18bae5880dfa30a8c36/examples/third_party/autotools/autotools_repositories.bzl#L30

UebelAndre: I think there are plans to move the autotools stuff into `rules_foreign_cc` proper and compile them for use in the toolchains. So I think this as an example will go away soon but I feel more importantly that users should have an easy time copying these examples for use in their own projects. So even if/when the autotools example moves, I'd expect to see the `rules_perl` dependency associated with `openssl`.

jheaff1: Ok sure, I’ll move it back.

I’m on my Christmas break now, I’ll revisit this in the New Year 👍🏻

UebelAndre: Hope you don't mind. I applied the changes since they were pretty straight forward 😄

UebelAndre (approved):

examples/third_party/openssl/openssl_repositories.bzl
jheaff1: Not at all, thank you very much!

Issue #329, opened 2019-10-18
Parallel build support


mattgodbolt:
This might be a big ask; and may be a bazel core project issue; but when consuming larger projects that need to be `configure/make`'d (e.g. OpenSSL, postgres, hdf5...) it would be great to be able to use parallel builds.

Of course, pragmatically one might override `make_commands` to have `make -j12` in it...but that has disastrous consequences with other parallel builds (not in the least also on machines with fewer than 12 cores...)

In the land of gumdrops and unicorns, one could imagine `bazel` being able to trick out `make` by passing it the same magic that `$(MAKE)` or `+make` does in a real makefile, along with a `-j` set from the `--jobs` setting. That way, the `make` process will play nicely with the bazel build.

I realise this is a big ask: but if one doesn't ask, then maybe it won't happen! Any ideas on improving this in general also welcomed!

Thanks all!

mattgodbolt:
In case I was being incoherent, the POSIX jobserver for make is described in more detail here: https://www.gnu.org/software/make/manual/html_node/POSIX-Jobserver.html#POSIX-Jobserver

irengrig:
Thank you for the feedback, I think it is worth doing.
The other problem is how to balance the number of Bazel's jobs with make's jobs (also, there may be several parallel rules_foreign_cc targets being processed.)

HackAttack:
I’d argue that this is vital to make this project usable. For example, running the example on my machine and building OpenBLAS takes 6 minutes, but when I manually build it with `make` it takes 13 seconds!

lawsonAGMT:
I will second this. We have a cmake project dependency that is significantly larger than any other component.

rdelfin:
I'm finding myself with a cmake project whose build time goes up by almost 4x when compared with how it would go with a manual build. As a stop gap, what would be needed to simply add a fixed parameter in the target that lets you specify how many processes you want to create for the job server?

HackAttack:
Related:
bazelbuild/bazel#6477
bazelbuild/bazel#10443

dhbaird:
Not a solution yet: But if I could find a way to let `cmake_external` inherit file descriptors, then a wrapper Makefile would hypothetically get me to a usable workaround, where CPU over-commit would have a 1x upper bound:

`build: ; +bazel @all//... --action_env="MAKEFLAGS=${MAKEFLAGS}"`
[...]

PR #848, opened 2022-01-01
Add parallel make toolchain wrapper

jsharpe:
This can be used by adding
`--extra_toolchains=@rules_foreign_cc//toolchains:built_parallel_make` to your bazel command line

Addresses some of #329


jsharpe:
Currently fails to work due to #433

PR #849, opened 2022-01-01
WIP resolve_tools

jsharpe:
Initial attempt at reviving some old code that I put together that calls `ctx.resolve_tools` on the tools to set the manifest and runfiles for the tools correctly.

This patch is sufficient to resolve the errors seen in using #848 but isn't yet in a state that is ready to merge.

### DIFF
--- a/foreign_cc/private/framework.bzl
+++ b/foreign_cc/private/framework.bzl
@@ -381,7 +381,7 @@ def cc_external_rule_impl(ctx, attrs):
     _print_deprecation_warnings(ctx)
     lib_name = attrs.lib_name or ctx.attr.name
 
-    inputs = _define_inputs(attrs)
+    inputs = _define_inputs(ctx, attrs)
     outputs = _define_outputs(ctx, attrs, lib_name)
     out_cc_info = _define_out_cc_info(ctx, attrs, inputs, outputs)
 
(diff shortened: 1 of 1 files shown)


foreign_cc/private/framework.bzl
```diff
@@ -763,7 +766,8 @@ InputFiles = provider(
             "into $EXT_BUILD_DEPS/include."
         ),
         libs = "Library files built by Bazel. Will be copied into $EXT_BUILD_DEPS/lib.",
-        tools_files = (
+        tools_files = "All files (including runfiles needed for the tools",
```
jsharpe, line 769: I think that we probably also want a tools_manifests entry here and we should merge this with any additional manifests when we call `ctx.actions.run_shell`. I've not fully groked what the manifests input does to an action as its not well documented.

foreign_cc/private/framework.bzl
```diff
@@ -458,16 +458,19 @@ def cc_external_rule_impl(ctx, attrs):
     # TODO: `additional_tools` is deprecated, remove.
     legacy_tools = ctx.files.additional_tools + ctx.files.tools_deps
 
+    inputs_from_tools, manifests_from_tools = ctx.resolve_tools(tools = attrs.tools_deps)
```
jsharpe, line 461: Possibly should be `legacy_tools` here?

jsharpe: Also need to be aware of what this looks like once deprecated attributes are removed. build_data is not included in legacy_tools at the moment.

foreign_cc/private/framework.bzl
```diff
@@ -650,9 +653,9 @@ def _copy_deps_and_tools(files):
     lines += _symlink_contents_to_dir("lib", files.libs)
     lines += _symlink_contents_to_dir("include", files.headers + files.include_dirs)
 
-    if files.tools_files:
+    if files.tools_roots:
         lines.append("##mkdirs## $$EXT_BUILD_DEPS$$/bin")
-    for tool in files.tools_files:
+    for tool in files.tools_roots:
         lines.append("##symlink_to_dir## $$EXT_BUILD_ROOT$$/{} $$EXT_BUILD_DEPS$$/bin/".format(tool))
```
jsharpe, line 659: Does this symlinking break finding runfiles for the tool? It doesn't break it for MAKE but we don't find make via the path so aren't using this mechanism in that case.

PR #851, opened 2022-01-03
Platform agnostic output from test

jsharpe:
Use find to get deterministic output between GNU and BSD based platforms.

### DIFF
--- a/test/BUILD.bazel
+++ b/test/BUILD.bazel
@@ -64,10 +64,6 @@ symlink_contents_to_dir_test_rule(
 
 diff_test(
     name = "shell_method_symlink_contents_to_dir_test",
-    file1 = select({
-        ":freebsd": "expected/out_symlinked_dirs_freebsd.txt",
-        ":macos": "expected/out_symlinked_dirs_macos.txt",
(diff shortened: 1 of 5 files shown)


UebelAndre:
I feel like relying on `find`, a host dependency makes the tests less robust. Can you expand more on what this is solving for?

UebelAndre (changes requested):
I feel like relying on `find`, a host dependency makes the tests less robust. Can you expand more on what this is solving for?

jsharpe:
> I feel like relying on `find`, a host dependency makes the tests less robust. Can you expand more on what this is solving for?

IIRC I made these changes to facilitate refactoring of the symlinking code which is currently very slow for large source trees (e.g. cmake). I wanted to remove the platform specific output for these tests as its likely that the freebsd tests would regress silently as it is not being run anywhere.

Note that our shell toolchains currently make use of find to do a number of the shell actions and so I don't think that using find here is introducing anymore dependencies than we already have in the toolchains? I guess the only thing might be that we should mirror the similar logic for REAL_FIND that is done in the windows toolchain to find find properly on windows to allow this test to also be run on windows which it currently doesn't.

PR #872, opened 2022-02-09
Fixups to use of Labels to support bzlmod

jsharpe:


jsharpe:
For context see PR to bzlmod registry: 
https://github.com/bazelbuild/bazel-central-registry/pull/65#discussion_r791680260

foreign_cc/private/framework/toolchain.bzl
```diff
@@ -72,9 +66,8 @@ framework_toolchain_repository = repository_rule(
     doc = "A repository rule which defines a `@rules_foreign_cc//foreign_cc/private/framework:shell_toolchain` toolchain.",
     implementation = _framework_toolchain_repository_impl,
     attrs = {
-        "commands_src": attr.label(
-            doc = "The label of a `.bzl` source which defines toolchain commands",
-            allow_files = [".bzl"],
+        "commands_src": attr.string(
+            doc = "The string of a `.bzl` source which defines toolchain commands",
```
UebelAndre, line 70: Why remove the `.bzl` restriction?

foreign_cc/private/framework/toolchain.bzl
```diff
@@ -96,7 +89,7 @@ def register_framework_toolchains(register_toolchains = True):
 
     for item in TOOLCHAIN_MAPPINGS:
         # Generate a toolchain name without the `.bzl` suffix
-        toolchain_name = "rules_foreign_cc_framework_toolchain_" + item.file.name[:-len(".bzl")]
+        toolchain_name = "rules_foreign_cc_framework_toolchain_" + item.file.split(":")[1][:-len(".bzl")]
```
UebelAndre, line 92: Always nervous about raw indexing. Can you update the comment to be a bit more descriptive about that?

foreign_cc/private/framework/toolchain.bzl
jsharpe: because it becomes a string here - bzlmod remaps the names of the repos and so we can't evaluate the Label at this point as it gets used in  the context of another repo so unfortunately we need to pass it through as a string here. It's a bit brittle but as its all internal it works but isn't great.
I guess the real solution is to move the `.bzl` into the generated repo itself so we don't need to resolve a label to access it?

UebelAndre: Oooh, I didn't notice that the type changed. I'm more nervous about making larger changes than needing to restrict the values here. This seems fine.

foreign_cc/private/framework/toolchain.bzl
UebelAndre: I think it's fine just want to make sure the expectation is super clear as indexing is easy to miss.

I think it's probably time to discuss writing a process wrapper instead of making more changes to the toolchain. If you could assign `item.file.split(":")[1][:-len(".bzl")]` to another value with some comment I'd be happy.

UebelAndre (approved):
Thanks! Looks good pending the comment about the indexing 😄

jsharpe:
@UebelAndre PTAL when you get a moment, hopefully the changes I've made end up in simpler code in the end despite being a bit more churn.

PR #873, opened 2022-02-09
Pin buildifier version

jsharpe:


UebelAndre (approved):

UebelAndre:
Some of these are something we should fix but I also opened https://github.com/bazelbuild/buildtools/issues/1044 which I think should be implemented before we move to `buildifier v5`.

PR #900, opened 2022-04-20
Fix ninja bootstrap on macOS 12.3+

keith:
macOS 12.3 removed /usr/bin/python, which is what's currently in the
shebang of ninja's configure.py. This forces us to run this with python3
instead.

This can be removed if https://github.com/ninja-build/ninja/pull/2118 is merged

### DIFF
--- a/foreign_cc/built_tools/ninja_build.bzl
+++ b/foreign_cc/built_tools/ninja_build.bzl
@@ -11,7 +11,8 @@ load("//foreign_cc/private/framework:platform.bzl", "os_name")
 
 def _ninja_tool_impl(ctx):
     script = [
-        "./configure.py --bootstrap",
+        # TODO: Drop custom python3 usage https://github.com/ninja-build/ninja/pull/2118
+        "python3 ./configure.py --bootstrap",
         "mkdir $$INSTALLDIR$$/bin",
         "cp -p ./ninja{} $$INSTALLDIR$$/bin/".format(
             ".exe" if "win" in os_name(ctx) else "",


jsharpe (commented):
This makes an assumption that python3 is available on the path which it might not be. Should we take a dependency on the python toolchain and grab the python binary to run from there instead?

keith:
Yes it does, I was kinda thinking since we were technically relying on py2 before as well it might be ok to just swap that to py3. But you're right that it would be better to solve that way. Do you know if there's prior art for that somewhere?

jsharpe:
Hmm, I'm not sure there is prior art; I guess we'd need some way to convert a `py_runtime_pair` into something that returns `platform_common.TemplateVariables` that defines a `PYTHON` variable to be able to call the correct python but I don't know how to do that as the python toolchains is all native Java AFAIK and so I'm not sure how to extract the appropriate information in starlark (if its even currently possible).

Lets be pragmatic and go with the assumption that python3 is available on the path for now as we previously did for py2. I'll open another issue to track the fact that we should technically depend on a toolchain.

keith:
I think we probably could do that, but I guess it would require the user setup something? since I don't think there's consistency around python setup like there is C++ 🤔

jsharpe:
There is an autodetecting python toolchain AFAIK which searches the path for python / python3 and sets up a default toolchain but I've not looked at rules_python recently to know how/if that has changed at all. I use the dropbox rules to build the interpreter but that only works on linux at the moment.

PR #911, opened 2022-05-18
fix: expose CMAKE env variable from cmake toolchains

roman-kashitsyn:
This change updates the definitions of cmake toolchains to expose CMAKE
environment variable that makes it easier to integrate the toolchain
with other build rules (e.g., with rules_rust).


UebelAndre (approved):
Thanks!

PR #914, opened 2022-05-20
Build cmake in parallel

graywolf-at-work:
Without running the build in parallel, it takes forever. Most modern
machines have many cores, and for those that do not, the overhead will
probably not be that large. Most systems will likely use the prebuilt
toolchain anyway.


jsharpe:
https://github.com/bazelbuild/rules_foreign_cc/pull/921 switches the cmake build to use ninja which will by default use all the available cores. I'd rather not hard code a value like 16 because people do compile on local laptops which often don't have more than 4 cores available.

PR #939, opened 2022-06-30
Add ability to specify DLL dir

jheaff1:
C/C++ projects typically install DLLs in the "bin" folder and the
associated interface library in the "lib" folder. This commit changes
the default behaviour accordingly.

The now redundant postfix script in the cmake_hello_world_lib shared test has been removed. The test was also previously broken as no interface library was generated. This commit resolves that issue


jheaff1:
@jsharpe @UebelAndre  Any chance of of a review?

jsharpe (approved):
Its been a while since I've done any windows stuff but this looks reasonable.

PR #948, opened 2022-08-09
Supporting Startfiles

DanielEliad:
Hey :)
I tried compiling a libc using this repo but found out I can't use startfiles (crti.0, crt1.o, crtn.o) in targets that depend on my libc target.

So I added optional support for consuming generated startfiles


jsharpe (changes requested):
The change looks fine, but I think we need a bit more context in the doc strings as to what a startfile is and its purpose as it is a very specialised use case and google doesn't return many results for the term.
Could you also provide an example in the examples directory as this is primarily the way in which code features like this get tested and without this it is likely that the feature may be broken by future refactorings of the code as the initial use case context will likely get lost over time.

DanielEliad:
I added some more documentation in the attribute doc
As for the example I only came across this need when trying to compile uclibc would you like me to put my project in the examples/third_party here?

DanielEliad:
the CI seems to have failed with a non related error:
```
2022/08/10 21:31:23 could not resolve the version 'latest' to an actual version number: unable to determine latest version: could not list available releases for 6.0.0: could not list GCS objects at https://www.googleapis.com/storage/v1/b/bazel/o?delimiter=/&prefix=6.0.0/release/: could not fetch https://www.googleapis.com/storage/v1/b/bazel/o?delimiter=/&prefix=6.0.0/release/: Get "https://www.googleapis.com/storage/v1/b/bazel/o?delimiter=/&prefix=6.0.0/release/": stream error: stream ID 3; INTERNAL_ERROR
```

jsharpe:
It seems the RBE build failed because the imported example build has a hardcoded reference to use gcc whereas the CC compiler is configured to clang. I'm guessing the kernel headers possibly don't compile with clang?

PR #958, opened 2022-08-29
Expand variables in `configure_options`

petoknm:
For cross compiling ffmpeg, I needed to supply the compiler as an argument to `configure`. That is why I needed to add variable expansion to `configure_options` of `configure_make` rule.
```
configure_make(
  ...
  configure_options = [
    ...
    "--cc=$(CC)",
  ]
)
```


jsharpe:
Fixed in https://github.com/bazelbuild/rules_foreign_cc/pull/1102

PR #961, opened 2022-08-30
Specify path to sed on MacOS and FreeBSD

garymm:
The given commands require BSD sed and don't work with GNU sed.
Some users (at least me) have their PATH such that `sed` resolves to GNU
sed, and then the given command fails.


garymm:
Note I didn't update windows because I'm not sure if `/usr/bin/sed` works on Windows and I can't easily test it.

foreign_cc/private/framework/toolchains/linux_commands.bzl
```diff
@@ -74,7 +74,7 @@ if [ -d "$1" ]; then
   for file in ${files[@]+"${files[@]}"}; do
     local backup=$(mktemp)
     touch -r "${file}" "${backup}"
-    sed -i 's@'"$2"'@'"$3"'@g' "${file}"
+    /usr/bin/sed -i 's@'"$2"'@'"$3"'@g' "${file}"
```
jsharpe, line 77: On some linux systems sed is at `/bin/sed` - this is why CI has failed.

foreign_cc/private/framework/toolchains/freebsd_commands.bzl
```diff
@@ -83,7 +83,7 @@ if [ -d "$1" ]; then
   for file in ${files[@]+"${files[@]}"}; do
     local backup=$(mktemp)
     touch -r "${file}" "${backup}"
-    sed -i '' -e 's@'"$2"'@'"$3"'@g' "${file}"
+    /usr/bin/sed -i '' -e 's@'"$2"'@'"$3"'@g' "${file}"
```
jsharpe, line 86: Is this definitely correct on BSD? I don't have a bsd system to be able to test this on - is there some docs on the net that you can point to to back this up?

jsharpe (commented):
Apologies for the delay in getting this reviewed; can you provide a reference that the BSD path is correct? I reverted the linux change as its not correct on all distros.

foreign_cc/private/framework/toolchains/freebsd_commands.bzl
garymm: I couldn't find any documentation but I just created a FreeBSD 12.4 virtual machine and it's /usr/bin/sed.

```
root@freebsd:~ # which sed
/usr/bin/sed
root@freebsd:~ # uname -a
FreeBSD freebsd 12.4-RELEASE-p1 FreeBSD 12.4-RELEASE-p1 GENERIC  amd64
```

PR #964, opened 2022-09-12
Ping MSVC version to 2017

meteorcloudy:
Bazel CI recently upgrade the default MSVC version to 2019.  LLVM 11.0.0 is required to work with MSVC 2019, but not yet available on Bazel CI.


meteorcloudy:
@jsharpe

PR #967, opened 2022-09-20
Disable cc toolchain resolution in examples on Mac

comius:
The examples/cmake_android picks up local instead of Android Cc toolchains.


jsharpe:
I'm not familiar with the android toolchains at all so can you clarify to me here is there an issue with the build files in rules_foreign_cc examples that needs fixing to flip this flag or is it a bazel core issue?

jsharpe (approved):

PR #971, opened 2022-10-06
Add macro that faciliates "bazel run" of binary generated by rules_foreign_cc

### LINKED ISSUE #582: Can't `bazel run` things generated as `out_binaries`

jheaff1:
This resolves #582 

The macro also facilitates the running of dynamically linked executables, which could then be used as tools in dependent targets. This is useful when using LGPL licenced third party software that must be dynamically linked.

Note that the OpenSSL example has been changed to produce shared libraries and to produce the `openssl` binary. This tests that `runnable_binary` can successfully run a dynamically linked binary


jheaff1:
@UebelAndre @jsharpe  Would either of you be able to give me a hand with the MacOS failures in CI? 

It seems now that OpenSSL shared libraries are built rather than static libraries, the curl test fails on MacOS only.

jsharpe:
> @UebelAndre @jsharpe Would either of you be able to give me a hand with the MacOS failures in CI?
> 
> It seems now that OpenSSL shared libraries are built rather than static libraries, the curl test fails on MacOS only.

I haven't got time to spend much time on this right now but looks to be rpath related. I'm guessing the rpath in the binary isn't looking at the right path to find the shared library in the runfiles?

jheaff1:
I think the MacOS issue is related to bazelbuild/bazel#10254

jheaff1:
@jsharpe @UebelAndre As it seems that the use of shared libs on sandboxed macos is a bit broken, i have removed the failing tests from CI. I hope that's acceptable.

foreign_cc/private/runnable_binary_wrapper.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# Allow unbound variable here, in case LD_LIBRARY_PATH or similar is not already set
+set +u
+for dir in "${SHARED_LIBS_DIRS_ARRAY[@]}"; do
+    export ${LIB_PATH_VAR}="${!LIB_PATH_VAR}":"$dir"
+done
+set -u
+
+EXE=BIN
+$(rlocation "${EXE#external/}") $@
```
jsharpe, line 48: exec avoids creating another process and the quotes mean that spaces in any arguments get handled correctly.
```suggestion
exec $(rlocation "${EXE#external/}") "$@"
```

jsharpe (approved):
LGTM

PR #979, opened 2022-11-03
Hermetic pkg config

jheaff1:
This PR adds a a pkg-config toolchain built from source.

On Windows, pkg-config has a couple of shared library dependencies, therefore the `runnable_binary` macro is utilised. As such this PR is built on top of PR #971.

There are many changes in this PR, I have tried to produce atomic commits to make each commit easier to review. If the git commit history should be tidied further, please let me know.

FYI - Once this is merged I will then be submitting a PR to add meson support, which depends on this PR


jsharpe:
@jheaff1 thanks for tackling this - I've started this a few times and never quite got round to finishing it so glad to see it being done!
+1 on getting meson support in too.

jheaff1:
> @jheaff1 thanks for tackling this - I've started this a few times and never quite got round to finishing it so glad to see it being done!
> +1 on getting meson support in too.

No worries. Thanks for reviewing the `runnable_binary` PR. The main difficulty with this one was getting it working on Windows. Typical 😂

jsharpe:
> @jsharpe CI is now passing apart from the Bazel 4.0.0 tests. The issue seems to be that providing `$(execpath )` to `ctx.resolve_command` fails. This must be a bug that was fixed in Bazel 5.
> 
> Any thoughts on how best to move forward? Could we upversion the minimum version examples to bazel 5? I presume Bazel 4.x is not yet end-of-life as Bazel 4.2.3 was released last month.
> 
> I had to replace the usage of `ctx.expand_locations` with `ctx.resolve_command` to facilitate the usage of tools built as a sh_binary or py_binary. This currently only applies to the pkg-config toolchain (built as a sh_binary) but my approach to building Meson is to do so as a py_binary.

Have you tried bumping the minimum version to 4.2.3 to see if the fix has been backported at all?

jheaff1:
> > @jsharpe CI is now passing apart from the Bazel 4.0.0 tests. The issue seems to be that providing `$(execpath )` to `ctx.resolve_command` fails. This must be a bug that was fixed in Bazel 5.
> > Any thoughts on how best to move forward? Could we upversion the minimum version examples to bazel 5? I presume Bazel 4.x is not yet end-of-life as Bazel 4.2.3 was released last month.
> > I had to replace the usage of `ctx.expand_locations` with `ctx.resolve_command` to facilitate the usage of tools built as a sh_binary or py_binary. This currently only applies to the pkg-config toolchain (built as a sh_binary) but my approach to building Meson is to do so as a py_binary.
> 
> Have you tried bumping the minimum version to 4.2.3 to see if the fix has been backported at all?

Yeah I tried that but alas the issue still occurs

jheaff1:
Bazel 6 is due out this month I believe, perhaps once it is released we change the minimum supported version of this repo to Bazel 5.x?

jheaff1:
@jsharpe I managed to get it building on bazel 4.0.0 by adding a special flag, please see my latest commit for details

jsharpe (approved):
Thanks; this is awesome work @jheaff1.

jheaff1:
> Thanks; this is awesome work @jheaff1.

You're most welcome. Next up, Meson!

PR #996, opened 2023-01-03
Update cmake and ninja versions

jsharpe:
Updates the python script to autogenerate the shas for the source files for cmake.


UebelAndre (approved):
Nice!

The failures seem to be the same issues as https://github.com/bazelbuild/rules_rust/issues/1814 which is unfortunate. Not sure if a Googler has looked into it yet but the PR otherwise seems good.

Issue #289, opened 2019-07-08
Cross compiling with CMake for Android


keith:
Currently when building a project for Android, the rules_foreign_cc build seems to still think the build is happening for macOS. This results in downstream CMake config not being correctly set if they are changing behavior based on the fact that they are cross compiling, or if some CMake variables like [`ANDROID`](https://cmake.org/cmake/help/v3.7/variable/ANDROID.html) are set.

The easiest way to reproduce this is to add this:

```cmake
if (NOT ANDROID)
  message(FATAL_ERROR "Expected ANDROID to be set, system name is: ${CMAKE_SYSTEM_NAME}")
endif()
```

at the bottom of `cmake_hello_world_lib/static/src/CMakeLists.txt` and then run `bazel build //cmake_android:app`. At this point if you `cd` into the `BUILD_TMPDIR` you can also see that CMake run check commands like this one:

```
/var/tmp/_bazel_ksmiley/623733d138991bf6c2de613f55fdd0a9/sandbox/darwin-sandbox/179/execroot/rules_foreign_cc_tests/external/androidndk/ndk/toolchains/llvm/prebuilt/darwin-x86_64/bin/clang   \
 -D__ANDROID_API__=28 \
 -isystem/private/var/tmp/_bazel_ksmiley/623733d138991bf6c2de613f55fdd0a9/sandbox/darwin-sandbox/179/execroot/rules_foreign_cc_tests/external/androidndk/ndk/sysroot/usr/include/arm-linux-androideabi \
 -target armv7-none-linux-androideabi \
 -march=armv7-a \
 -mfloat-abi=softfp \
 -mfpu=vfpv3-d16 \
 -gcc-toolchain /private/var/tmp/_bazel_ksmiley/623733d138991bf6c2de613f55fdd0a9/sandbox/darwin-sandbox/179/execroot/rules_foreign_cc_tests/external/androidndk/ndk/toolchains/arm-linux-androideabi-4.9/prebuilt/darwin-x86_64 \
 -fpic \
 -no-canonical-prefixes \
 -Wno-invalid-command-line-argument \
 -Wno-unused-command-line-argument \
 -funwind-tables \
 -fstack-protector-strong \
 -fno-addrsig \
 -Werror=return-type \
 -Werror=int-to-pointer-cast \
 -Werror=pointer-to-int-cast \
 -Werror=implicit-function-declaration  \
[...]

PR #997, opened 2023-01-06
Fix cross compiling for android on macOS

keith:
The way rules_foreign_cc builds cross compilation targets isn't exactly what cmake expects. Because of this when targeting android on macOS, some macOS specific linker flags are added to cmake's test compile invocations, leading to failures. This works around this issue by setting some more variables to trick cmake into only using the expected settings.

Fixes https://github.com/bazelbuild/rules_foreign_cc/issues/289


foreign_cc/private/framework.bzl
```diff
@@ -209,6 +209,7 @@ CC_EXTERNAL_RULE_ATTRIBUTES = {
         cfg = "exec",
         default = [],
     ),
+    "_android_constraint": attr.label(default = Label("@platforms//os:android")),
```
keith, line 212: I chose to only add this one for now since it's a bit of boilerplate to add more, but I can if folks want to

jsharpe (approved):

PR #999, opened 2023-01-07
Add bzlmod support files.

jsharpe:
Adds configuration for publish-to-bcr app and adds the bzlmod files used in the existing BCR entry.

### DIFF
--- a/.bazelci/config.yaml
+++ b/.bazelci/config.yaml
@@ -61,6 +61,8 @@ tasks:
     linux_targets: &linux_targets_bzlmod
       - "//..."
       - "//:third_party_examples_linux_tests"
+      # gives error '//external:databinding_annotation_processor': target 'databinding_annotation_processor' not declared in package 'external' defined by /workdir/examples/WORKSPACE.bazel (Tip: use `query "//external:*"` to see all the targets in that package) and referenced by '@bazel_tools//tools/android:databinding_annotation_processor'
+      - "-//cmake_android/..."
     build_targets: *linux_targets_bzlmod
     test_targets: *linux_targets_bzlmod
     build_flags:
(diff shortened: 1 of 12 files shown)


jsharpe:
TODO: need to add CI jobs for testing bzlmod

.bazelci/config.yaml
```diff
@@ -54,6 +54,19 @@ tasks:
       - "//:third_party_examples_linux_tests"
     build_targets: *linux_targets
     test_targets: *linux_targets
+  ubuntu2004_examples_bzlmod:
+    name: Examples (bzlmod)
```
jheaff1, line 58: No MacOS or windows builds?

.bcr/metadata.template.json
```diff
@@ -0,0 +1,13 @@
+{
+  "homepage": "https://github.com/bazel-contrib/rules_cuda",
```
jheaff1, line 2: Should this be rules_foreign_cc rather than rules_cuda?

.bcr/presubmit.yml
```diff
@@ -0,0 +1,12 @@
+# We recommend included a bcr test workspace that exercises your ruleset with bzlmod.
+# For an example, see https://github.com/aspect-build/bazel-lib/tree/main/e2e/bzlmod.
```
jheaff1, line 2: Should this file be removed given that it’s all commented out?

jsharpe: This file is required for the BCR github app - my plan was to initially have this commented out file and then iterate filling this out in a registry PR where it will actually be tested and then backport it to the .bcr config directory when I have something working.

.bcr/metadata.template.json
jsharpe:
```suggestion
  "homepage": "https://github.com/bazel-contrib/rules_foreign_cc",
```

.bazelci/config.yaml
jsharpe: Most of the issues with bzlmod specific issues happen at analysis time; I initially have just left this scoped to linux just for the sake of keeping CI times down as macOS in particular is slow.

jheaff1 (approved):

PR #1002, opened 2023-01-19
Disable fdo features.

rjogrady:
fdo_instrument and fdo_optimize features should be disabled. These are not supported by external build systems.


jsharpe (approved):
LGTM, Thanks!

Issue #1005, opened 2023-01-27
Flag --incompatible_python_disable_py2 will break rules_foreign_cc in Bazel 7.0


keertk:
Incompatible flag `--incompatible_python_disable_py2` will be enabled by default in the next major release (Bazel 7.0), thus breaking rules_foreign_cc. Please migrate to fix this and unblock the flip of this flag.

The flag is documented here: [bazelbuild/bazel#17293](https://github.com/bazelbuild/bazel/issues/17293).

Please check the following CI builds for build and test results:

- [MacOS OpenJDK 11](https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/1408#0185f0c4-9ffa-40a6-a0f5-bfa65b8f7f88)
- [Ubuntu 18.04 OpenJDK 11](https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/1408#0185f0c4-9ff1-470d-ba0b-821ce1bf338e)
- [MacOS OpenJDK 11](https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/1408#0185f0c4-9ffd-4e5f-a782-b25959a4468c)
- [Ubuntu 20.04 OpenJDK 11](https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/1408#0185f0c4-9ff4-4895-b97a-f76e8b80a389)
- [Ubuntu 16.04 OpenJDK 8](https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/1408#0185f0c4-9ff7-4884-b745-0e82a42ce86e)

Never heard of incompatible flags before? We have [documentation](https://docs.bazel.build/versions/master/backward-compatibility.html) that explains everything.
If you have any questions, please file an issue in https://github.com/bazelbuild/continuous-integration.

PR #1006, opened 2023-01-27
Remove Python 2 configuration and targets

rickeylev:
Bazel is dropping support for Python 2

Also raises the minimum Bazel version in the examples to 5.0, as 4.0 has built-in dependencies on Python 2.

Fixes https://github.com/bazelbuild/rules_foreign_cc/issues/1005

Work towards https://github.com/bazelbuild/bazel/issues/17293


rickeylev:
re: CI failure: The failing case is the "minimum supported version test" because @bazel_tools//tools/android:strip_resources wants to use Python 2.

That was fixed back in 2019: https://github.com/bazelbuild/bazel/commit/d93a146463e20f4eedc5ae658b03ee0260750aa7

As far as I can tell, that code is in the 4.0 release. So I'm not sure what's going on.

.bazelci/config.yaml
```diff
@@ -201,18 +201,13 @@ tasks:
       - "--experimental_enable_aggregating_middleman=False"
   min_supported_version_examples:
     name: "Minimum Supported Version Examples"
-    bazel: "4.0.0"
```
jsharpe, line 204: Technically bazel 4 is still under LTS support until the end of this year and so I'd rather that we just disable the failing tests and keep this at 4 as the rest of these changes only affect the examples / tests and don't change the ability of this ruleset to continue support v4.

jsharpe (changes requested):
Thanks for the fixes but can you just disable the failing tests rather than bump the minimum supported version?

.bazelci/config.yaml
rickeylev: I restored 4.0 as the min version.

I'll try my hand at modifying the config.yaml to skip the offending tests, but I might come back asking for some help on how to do so. (If you happen to have a link that explains these bazelci config files, that'd be great; the only thing i've found is https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/README.md which doesn't explain many of the things i see or how to use them).

rickeylev: OK, I think I figured it out.

Ready for another look

jsharpe (approved):
LGTM

PR #1017, opened 2023-02-22
Add new darwin CC toolchain for tests

keith:
Bazel 7.x is going to remove the CC toolchain that supports iOS and use the unix toolchain for macOS support. Since there are some tests that use the iOS toolchain we need to set that up for testing.

### DIFF
--- a/examples/.bazelrc
+++ b/examples/.bazelrc
@@ -1,4 +1,3 @@
-
 # Required for Windows
 build --enable_runfiles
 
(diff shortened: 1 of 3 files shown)


jsharpe:
@keith it seems the version of `build_bazel_apple_support` unconditionally uses module_extension which doesn't work on 4.0 which is the current minimum supported version for this ruleset. Is there some way we can conditionally load this based on bazel version?

keith:
hrm. it sounds like maybe i could move this into a different file, but that might require users have to change their load statements. given that bazel doesn't support 4.x any more do you know when this repo might drop that compat as well?

jsharpe:
> hrm. it sounds like maybe i could move this into a different file, but that might require users have to change their load statements. given that bazel doesn't support 4.x any more do you know when this repo might drop that compat as well?

https://blog.bazel.build/2020/11/10/long-term-support-release.html - this timeline suggests that 4.x is still LTS until Dec 2023 unless you know of a different timeline that says otherwise?

keith:
oh interesting. i just recently saw some comments about how another 5.x release wasn't going to be cut, which made me assume that was also considered not very well supported anymore, but I guess not

keith:
I added a patch here to just toss that from apple_support, should work for the short term

jsharpe (approved):

PR #1021, opened 2023-02-28
Disable coverage when a C library is not instrumented

linzhp:
When building in coverage mode, Bazel's cc_common.configure_features() adds coverage related flags, such as `--coverage` to the compiler and linker. However, if there is nothing in this library or its deps that are instrumented, we don't need to pass those flags, and avoid unncessary rebuilds.


jsharpe (approved):
LGTM, thanks!

PR #1023, opened 2023-03-01
Add --sysroot flag to CPPFLAGS

### LINKED ISSUE #985: --sysroot parameter should probably be included in CPPFLAGS
I'm investigating a build break that happens when `configure_make` target is built using custom toolchain with `builtin_sysroot`

`--sysroot` parameter is passed to the external project int `CFLAGS`, `CXXFLAGS` and `LDFLAGS` but not in `CPPFLAGS` that only contains include-related flags in my case.

Makefile in question contains pre-processor-only step that references `$(CPPFLAGS)`. This step fails because correct sysroot value is not passed to the gcc.

The fix should probably be adding `--sysroot` parameter to `CPPFLAGS` env variable.

carlosgalvezp:
This is needed for hermetic toolchains that specify a sysroot via the "builtin_sysroot" in the
"create_cc_toolchain_config_info" function.

Fixes #985

### DIFF
--- a/foreign_cc/private/make_env_vars.bzl
+++ b/foreign_cc/private/make_env_vars.bzl
@@ -27,6 +27,11 @@ def get_make_env_vars(
     # https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Preset-Output-Variables.html
     vars["CPPFLAGS"] = deps_flags.flags
 
+    # CPPFLAGS needs to have access to the custom sysroot, if one is set.
+    for flag in vars["CFLAGS"]:
+        if flag.startswith("--sysroot="):
+            vars["CPPFLAGS"].append(flag)
+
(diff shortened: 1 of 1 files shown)


carlosgalvezp:
For some reason I cannot add reviewers in the `Reviewers` field,  @irengrig  @oquenchil would you be able to take a look? Thanks!

carlosgalvezp:
sysroot already is present in CFLAGS so this should not be needed.

timbess:
@carlosgalvezp So I just tripped over this as well I believe that's not correct. CFLAGS are all you need to _compile_ the code, but the configure step often runs preprocessor only compile commands to detect what's available on the system. This will fail unless CPPFLAGS are set and cause weird issues where the configure/compile steps disagree with what exists.

Issue #931, opened 2022-06-20
cmake rule can't decude `CMAKE_CXX_COMPILER` correctedly.


MenglingD:
cc/c++: /opt/rh/devtoolset-8/
os: centos7
rules_foreign_cc versoin: 0.8.0

```bazel
WORKSPACE:

workspace(name = "spdlog")

load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

http_archive(
  ¦ name = "rules_foreign_cc",
  ¦ sha256 = "6041f1374ff32ba711564374ad8e007aef77f71561a7ce784123b9b4b88614fc",
  ¦ strip_prefix = "rules_foreign_cc-0.8.0",
  ¦ url = "https://github.com/bazelbuild/rules_foreign_cc/archive/0.8.0.tar.gz",
)
load("@rules_foreign_cc//foreign_cc:repositories.bzl", "rules_foreign_cc_dependencies")
rules_foreign_cc_dependencies()
```

```bazel
BUILD.bzl
package(default_visibility = ["//visibility:public"])
load("@rules_foreign_cc//foreign_cc:defs.bzl", "cmake")

filegroup(
  ¦ name = "srcs",
  ¦ srcs = glob(["**"]),
  ¦ visibility = ["//visibility:public"],
)

cmake(
  ¦ name = "libspdlog",
  ¦ lib_source = ":srcs",
  ¦ cache_entries = {
  ¦ ¦ ¦ #"SPDLOG_BUILD_EXAMPLE": "OFF",
  ¦ },
  ¦ out_lib_dir = "lib64",
  ¦ out_static_libs = ["libspdlog.a"],
  ¦ # out_shared_libs = ["libspdlog.so"],
)
```

git: [spdlog-bazel](https://github.com/MenglingD/spdlog-bazel)，which would be failed to build as `__var_CMAKE_CXX_COMPILER` has wrong value located in `bazel-out/k8-fastbuild/bin/libspdlog_foreign_cc/build_script.sh`:

```bash
......

__var_CMAKE_CXX_COMPILER="/opt/rh/devtoolset-8/root/usr/bin/gcc"

......
```

The value  of `__var_CMAKE_CXX_COMPILER` should be `/opt/rh/devtoolset-8/root/usr/bin/c++`.

jheaff1:
Sadly this is due to core bazel, which builds C++ files using gcc rather than c++. There isn't a lot that rules_foreign_cc can do about that.

[...]

PR #1030, opened 2023-03-16
Pick correct compiler for CMAKE_CXX_COMPILER

ozanarmagan:
Fix to solve the problem of picking gcc instead of g++ and clang instead of clang++ by a simple function.


ozanarmagan:
Hi @jsharpe, this little trick can fix #931

jheaff1:
Would this also work when cross compiling, eg with a compiler named “arm-none-eabi-gcc”?

PR #1037, opened 2023-04-11
Use python3 instead of python2 in macos commands

### LINKED ISSUE #1036: python 2.x usage on macOS
Hi all, I noticed python 2.x support is dropped per https://github.com/bazelbuild/rules_foreign_cc/issues/1005.

However, the python 3.x installation only produce `python3` under normal `PATH`.

In some macOS commands helpers, we use `python` instead of `python3` like (https://github.com/bazelbuild/rules_foreign_cc/blob/d33d862abb4ce3ba178547ef58c9fcb24cec38ca/foreign_cc/private/framework/toolchains/macos_commands.bzl#L264) and it causes error.

Should these commands be changed to `python3` instead?

fffonion:
Fix #1036

### DIFF
--- a/foreign_cc/private/framework/toolchains/macos_commands.bzl
+++ b/foreign_cc/private/framework/toolchains/macos_commands.bzl
@@ -261,7 +261,7 @@ def replace_symlink(file):
     # as `readlink` is.
     return """\
 if [[ -L "{file}" ]]; then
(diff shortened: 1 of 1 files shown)


jsharpe:
Thanks, this makes sense for recent versions of macOS; CI is on an older version which still has python2 available which is why this hasn't been caught earlier by CI.

PR #1038, opened 2023-04-11
installdir_copy seems meaningless as output

lifengxiang1025:
Hi, I found "installdir_copy" seems meaningless as output.  The "outputs.declared_outputs" has lib/ and include/. I think this's enough.


jsharpe:
I think this is reasonable - if someone requires the previous behaviour they can add gen_dir to the output_groups.

jsharpe (approved):

PR #1046, opened 2023-04-26
Pass must_keep_debug = False to cc_common.create_link_variables

### LINKED ISSUE #1045: Support bazel's `--strip` parameter
See docs for `--strip` [here](https://bazel.build/docs/user-manual#strip).

One might expect `bazel build //... --strip=always` to generate `-Wl,-S` or equivalent parameters when linking foreign CC binaries, but `rules_foreign_cc` doesn't support this parameter.

Turns out adding support is easy: just pass `must_keep_debug = False` when calling `cc_common.create_link_variables`.

The docs say that when `must_keep_debug` is `True` (the default), then the `strip_debug_symbols` will be made available to the toolchain to decide what to do. But the docs are wrong; the value must be set to `False` (https://github.com/bazelbuild/bazel/issues/14322)

jvolkman:
Fixes #1045

### DIFF
--- a/foreign_cc/private/cc_toolchain_util.bzl
+++ b/foreign_cc/private/cc_toolchain_util.bzl
@@ -268,6 +268,7 @@ def get_flags_info(ctx, link_output_file = None):
                 feature_configuration = feature_configuration,
                 is_using_linker = True,
                 is_linking_dynamic_library = True,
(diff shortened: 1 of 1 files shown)


jsharpe (approved):
Good find!

Issue #1049, opened 2023-05-06
The `make()` rule is underdocumented and has confusing behavior


filmil:
See here:

https://github.com/filmil/rules_foreign_cc/commit/cf3471991485fa6b0ba221ed798e81d5f3eb9940

PR #1050, opened 2023-05-06
Adds some documentation specific to the `make` rule.

filmil:
I learned the hard way that the `make` build rule does not work out of the box with just any makefile, unlike other rules in this repository.

The single `make` example does *not* emphasize this, and it is important for the usability of the rule.

These are the things I learned:

- the `make( targets = [...])` invocation must at least execute an equivalent of `make install`, since all `out_*` params are relative to the *installation* directory, not to either the source or the buidkl directory.  So if your default target does not install anything, or you have a `install` make target, but don't invoke it, bazel will be unable to find your build outputs.
- The make file must either be able to recognize the `PREFIX=` env variable and use it, or have a different way of wiring the `$$INSTALLDIR$$` make variable through.
- Finally, the `make install` recipe must ensure that it does not copy symlinks. This will bite you if you use `cp $SOURCE DEST` instead of `cp -L $SOURCE $DEST` since the former will create a bunch of symlinks pointing to the sandbox. Which, of course, is useless once the sandbox is removed. Using `-L` will ensure that the actual file contents are
copied.

Issue: #1049

### DIFF
--- a/foreign_cc/make.bzl
+++ b/foreign_cc/make.bzl
@@ -94,7 +94,9 @@ def _attrs():
         "targets": attr.string_list(
             doc = (
                 "A list of targets within the foreign build system to produce. An empty string (`\"\"`) will result in " +
-                "a call to the underlying build system with no explicit target set"
+                "a call to the underlying build system with no explicit target set. However, in order to extract build " +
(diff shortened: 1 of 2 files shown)


jsharpe (approved):
Thanks!

PR #1056, opened 2023-06-12
Allow selecting individual outputs from a rule

### LINKED ISSUE #768: Link only one of the output libs?
Hi,

I've searched and tried a bit but seems this is a relatively new topic.

I have several external projects (configure_make, cmake), each has multiple output libs (i.e. I specify multiple entries in `out_static_libs`). But for my Bazel target, I want to link toward only one of them each, and found it difficult.

An example (not my real example) would be `openssl`, how can I link only `libcrypto.a` (without `libssl.a`)? Without defining multiple `configure_make` targets (which takes times longer to build).

My real cases are for external projects that

- Output different libs for C and C++;
- Output different libs with a specific feature (e.g. threading / no-threading);
- Output multiple libs each for a specific feature (e.g. file / networking / etc.);

And I do need to use all of them for different Bazel targets (e.g. link only `libcrypto.a` for one target, but both for another).

DolceTriade:
If one wants to select a single output from a build (like a single shared library or a single static library), it's not possible to do so with the current code. The current code only creates an output_group for the declared binaries, which isn't sufficient.

This change creates an output group for all the artifacts including the includes.

Fixes #768

### DIFF
--- a/foreign_cc/private/framework.bzl
+++ b/foreign_cc/private/framework.bzl
@@ -531,7 +531,7 @@ def cc_external_rule_impl(ctx, attrs):
         lib_dir_name = attrs.out_lib_dir,
         include_dir_name = attrs.out_include_dir,
     )
-    output_groups = _declare_output_groups(installdir_copy.file, outputs.out_binary_files)
+    output_groups = _declare_output_groups(installdir_copy.file, outputs.out_binary_files + outputs.libraries.static_libraries + outputs.libraries.shared_libraries + [outputs.out_include_dir])
     wrapped_files = [
         wrapped_outputs.script_file,
         wrapped_outputs.log_file,


DolceTriade:
```


bazel build --show_progress_rate_limit=5 --curses=yes --color=yes --terminal_columns=143 --show_timestamps --verbose_failures --jobs=30 --announce_rc --experimental_repository_cache_hardlinks --disk_cache= --sandbox_tmpfs_path=/tmp --google_default_credentials --remote_cache=remotebuildexecution.googleapis.com --remote_instance_name=projects/bazel-untrusted/instances/default_instance --bes_backend=buildeventservice.googleapis.com --bes_timeout=360s --bes_instance_name=bazel-untrusted --remote_timeout=60 --remote_max_connections=200 --remote_default_platform_properties=properties:{name:"cache-silo-key" value:"f715d4f7e3ad3942ee54f8cd9a9ed93f7808f63c82bbf62d6628653f83f874fa"} --remote_download_toplevel --experimental_enable_aggregating_middleman=False --test_env=HOME --test_env=BAZELISK_USER_AGENT --test_env=USE_BAZEL_VERSION -- //... -//cmake_android:app
--
  | ERROR: --bes_instance_name=bazel-untrusted :: Unrecognized option: --bes_instance_name=bazel-untrusted
  | INFO: Invocation ID: 72a67d79-be10-457d-a3f8-556cbf632730
  | bazel build failed with exit code 2
  | 🚨 Error: The command exited with status 1
  | user command error: The plugin docker command hook exited with status 1


```

Seems like the CI is broken unrelated to this change?

PR #1062, opened 2023-06-20
fix linux cross compiling on macos

isaactorz:
Fixes cross-compiling from macos to linux.

see #997 for background.

I had to make a couple of extra changes to support this:
- Setting `CMAKE_SYSTEM_NAME` manually causes `CMAKE_SYSTEM_PROCESSOR` to not be set. This breaks some builds that expect this variable to be set like [libjpeg-turbo](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/4e52b66f342a803d3b8099b79607e3158d3a241c/CMakeLists.txt#L43).
- I made it so that the above variables are only set when cross-compilation is detected so that `rules_foreign_cc` only takes on responsibility of setting them when necessary.

### DIFF
--- a/foreign_cc/private/cmake_script.bzl
+++ b/foreign_cc/private/cmake_script.bzl
@@ -114,6 +114,9 @@ def create_cmake_script(
         },
     }
 
+    print(target_os)
+    print(host_os)
+    print(arch)
     if target_os != host_os:
         params.cache.update(_target_os_params.get(target_os))
         params.cache.update(_target_arch_params.get(target_arch))
--- a/foreign_cc/private/cmake_script.bzl
+++ b/foreign_cc/private/cmake_script.bzl
@@ -102,16 +102,16 @@ def create_cmake_script(
         },
         "linux": {
             "CMAKE_SYSTEM_NAME": "Linux",
-        }
+        },
     }
 
     _target_arch_params = {
         "x86_64": {
-            "CMAKE_SYSTEM_PROCESSOR": "x86_64"
+            "CMAKE_SYSTEM_PROCESSOR": "x86_64",
         },
         "aarch64": {
             "CMAKE_SYSTEM_PROCESSOR": "aarch64",
-        }
+        },
     }
 
     if target_os != host_os:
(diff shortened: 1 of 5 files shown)


isaactorz:
If anyone happens to peak at this - the issue is setting `CMAKE_SYSTEM_NAME` requires that you also set `CMAKE_SYSTEM_PROCESSOR` manually

This is breaking `libjpeg-turbo`'s build scripts: https://github.com/libjpeg-turbo/libjpeg-turbo/blob/4e52b66f342a803d3b8099b79607e3158d3a241c/CMakeLists.txt#L43.

foreign_cc/private/cmake_script.bzl
```diff
@@ -17,9 +17,30 @@ def _escape_dquote_bash_crosstool(text):
...
+        "CMAKE_SYSTEM_NAME": "Linux",
+    },
+    "linux": {
+        "CMAKE_SYSTEM_NAME": "Linux",
+    },
+}
+
+_TARGET_ARCH_PARAMS = {
+    "aarch64": {
+        "CMAKE_SYSTEM_PROCESSOR": "aarch64",
```
isaactorz, line 32: There's a little more nuance to it than this (mostly on mac) but I think this is a good enough start.

isaactorz:
cc @keith @jsharpe sorry for the direct ping but don't know the etiquette for this project for requesting reviews.

Just wanted to get it on someones radar.

foreign_cc/private/cmake_script.bzl
```diff
@@ -93,13 +116,13 @@ def create_cmake_script(
...
-    if target_os == "android":
-        params.cache.update({
-            "ANDROID": "YES",
-            "CMAKE_SYSTEM_NAME": "Linux",
-        })
+    # Avoid CMake passing the wrong linker flags when cross compiling
+    # by setting CMAKE_SYSTEM_NAME and CMAKE_SYSTEM_PROCESSOR,
+    # see https://github.com/bazelbuild/rules_foreign_cc/issues/289,
+    # and https://github.com/bazelbuild/rules_foreign_cc/pull/1062
+    if target_os != host_os and target_os != "unknown":
```
isaactorz, line 123: There's more to it than this (i.e. linux-x86_64 -> linux-aarch64).

Probably should be a function like `is_cross_compilation` and handle this more rigorously.

jsharpe:
This is on my radar to review - will try to take a look at some point this week!

jsharpe (approved):
LGTM, thanks!

PR #1068, opened 2023-06-29
Set the LD environment variable

dstufft:
Some configure + make systems look for the ``ld`` binary that is supposed to be used, and invoke it directly for various reasons. 

Currently this uses the ``ld`` binary from ``$PATH``, not from the toolchain, so we set the ``LD`` environment variable to ensure that we're using the ``ld`` binary.

Note: This grabs the ``ld`` executable directly from ``cc_toolchain.ld_executable`` rather than using something like ``cxx_linker_executable``, because the former returns ``CC`` generally, but ``LD`` should be the linker directly.


jsharpe (approved):
LGTM, thanks!

PR #1069, opened 2023-06-29
Make pkgconfig Hermetic

### LINKED ISSUE #1065: pkgconfig_tool_unix not hermetic
It looks like when attempting to build pkgtool, rules_foreign_cc is not setting the ``CC`` variable, causing the pkgtool configure to search path for ``gcc``, ``clang``, etc, and using that if it finds one. This can be reproduced using a docker image (to ensure there's no `CC` and a hermetic CC compiler (rather than the default option of letting Bazel search the ``$PATH`` for one.

Then once you get past that, it attempts to just invoke ``make`` without attempting to use the toolchain provided ``make``, so if you also don't have ``make`` on your ``$PATH``, then it will fail then.

Unfortunately, it appears even with all of that, it still fails, but it fails with incompatible integer to pointer conversion, which I'm not sure whose to blame for that.

You can see a minimal reproducer at [dstufft/rules-foreign-cc-hermetic-reproducer](https://github.com/dstufft/rules-foreign-cc-hermetic-reproducer), which you can test doing 

```
$ docker run --rm -v $PWD:/src -it $(docker build -q .) bazel build //...
```

It appears like the ``make_build`` rules do some work to setup an environment prior to building ``make``, however the ``pkgconfig`` rules don't appear to do the same work. I tried just copying what the ``make_build`` rules did on my branch [here](https://github.com/bazelbuild/rules_foreign_cc/compare/main...dstufft:rules_foreign_cc:built-make).

However doing that just causes another error when attempting to build ``pkgconfig``:

```
gatomic.c:392:10: error: incompatible integer to pointer conversion passing 'gssize' (aka 'long') to parameter of type 'gpointer' (aka 'void *') [-Wint-conversion]
  return g_atomic_pointer_add ((volatile gpointer *) atomic, val);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./gatomic.h:170:46: note: expanded from macro 'g_atomic_pointer_add'
    (gssize) __sync_fetch_and_add ((atomic), (val));                         \
                                             ^~~~~
gatomic.c:416:10: error: incompatible integer to pointer conversion passing 'gsize' (aka 'unsigned long') to parameter of type 'gpointer' (aka 'void *') [-Wint-conversion]
  return g_atomic_pointer_and ((volatile gpointer *) atomic, val);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./gatomic.h:177:45: note: expanded from macro 'g_atomic_pointer_and'
    (gsize) __sync_fetch_and_and ((atomic), (val));                          \
                                            ^~~~~
gatomic.c:440:10: error: incompatible integer to pointer conversion passing 'gsize' (aka 'unsigned long') to parameter of type 'gpointer' (aka 'void *') [-Wint-conversion]
  return g_atomic_pointer_or ((volatile gpointer *) atomic, val);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./gatomic.h:184:44: note: expanded from macro 'g_atomic_pointer_or'
    (gsize) __sync_fetch_and_or ((atomic), (val));                           \
                                           ^~~~~
gatomic.c:464:10: error: incompatible integer to pointer conversion passing 'gsize' (aka 'unsigned long') to parameter of type 'gpointer' (aka 'void *') [-Wint-conversion]
  return g_atomic_pointer_xor ((volatile gpointer *) atomic, val);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./gatomic.h:191:45: note: expanded from macro 'g_atomic_pointer_xor'
    (gsize) __sync_fetch_and_xor ((atomic), (val));                          \
                                            ^~~~~
4 errors generated.
```

At this point, I had to quit debugging, I'm not sure if the above errors come from flags that Bazel passes in, from zig (since my reproducer uses zig to get a hermetic CC toolchain), or if ``pkgconfig`` in the repo is broken in some way. I figured I'd record this though.

dstufft:
Make sure that when building ``pkg-config``, that we're using the ``make`` and other tools from our configured toolchain rather than using whatever we happen to find on ``$PATH``.

Fixes #1065


dstufft:
Hmm, I don't know enough about macOS to understand why exactly it's failing here

steeve:
> Hmm, I don't know enough about macOS to understand why exactly it's failing here

For some reason the `ARFLAGS` are not passed:
```

CC       libcharset_la-localcharset.lo
--
  | CCLD     libcharset.la
  | error: /Applications/Xcode14.2.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/libtool: no output file specified (specify with -o output)
```

I can repro it locally.

dmeijboom:
@dstufft We are currently setting `AR` to an empty string in the environment variables, that seems to work (not sure why). Would that be a proper fix?

See: https://github.com/bazelbuild/rules_foreign_cc/commit/48d7b9fa84fef6c2c58f810fa629c39e2572125b

foreign_cc/built_tools/pkgconfig_build.bzl
```diff
@@ -6,14 +6,74 @@ load(
...
     "built_tool_rule_impl",
 )
+load(
+    "//foreign_cc/private:cc_toolchain_util.bzl",
+    "get_env_vars",
+    "get_flags_info",
+    "get_tools_info",
+)
 load("//toolchains/native_tools:tool_access.bzl", "get_make_data")
+load("//foreign_cc/private/framework:platform.bzl", "os_name")
```
jsharpe, line 19:
```suggestion
load("//foreign_cc/private/framework:platform.bzl", "os_name")
load("//toolchains/native_tools:tool_access.bzl", "get_make_data")
```

jsharpe (approved):
Hopefully setting AR to an empty string fixes the macOS build - this just causes it to use ar instead of libtool as configured in the default macOS toolchain.
Otherwise this looks good to me - thanks!

keith:
would be nice to have a release with this one

jsharpe:
> would be nice to have a release with this one

I'll try and cut a release today..

jsharpe:
Annoyingly the CI job for the release process failed; I don't currently have time to debug this - the tag is present on the repo so you can reference that release but cutting the release to the BCR will have to be more manual..

jsharpe:
Actually turned out to be quite straightforward - 0.11.1 should be in the BCR once it's been reviewed there.

PR #1084, opened 2023-08-25
Fix multiple-definition error on windows_config_setting

novas0x2a:
When using runnable_binary twice in the same package, windows_config_setting was being defined twice producing an error like:

Error in config_setting: config_setting rule 'windows_config_setting' in package '' conflicts with existing config_setting rule, defined at .../BUILD.bazel:29:16

This fix prefixes the setting with the name, the same as the other rules defined by the macro.


jsharpe (approved):
Thanks! LGTM.

PR #1091, opened 2023-09-07
change cygpath to mixed

matt-sm:
`cygpath -w` will print `\` eg. `c:\Users\msmith`.  

When this is used to calculate `EXT_BUILD_ROOT`, and `EXT_BUILD_ROOT` is used as a suffix for other paths, it can lead to a mix of `\` amd `/` which can cause `configure_make` build failures. eg.

```
unsafe srcdir value: 'C:\tmp\2b4dntvp\execroot\__main__/external/
```

`cygpath` comes preinstalled in msys2, therefore is always used to calculate `pwd`.

The fallback, `pwd -W` will print `/` eg. `c:/Users/msmith`.

This MR changes the Windows `pwd()` to use the `cygpath -m` "mixed" mode that matches `pwd -W`.

### DIFF
--- a/foreign_cc/private/framework/toolchains/windows_commands.bzl
+++ b/foreign_cc/private/framework/toolchains/windows_commands.bzl
@@ -9,7 +9,7 @@ def script_extension():
     return ".sh"
 
 def pwd():
-    return "$(type -t cygpath > /dev/null && cygpath $(pwd) -w || pwd -W)"
+    return "$(type -t cygpath > /dev/null && cygpath $(pwd) -m || pwd -W)"
 
 def echo(text):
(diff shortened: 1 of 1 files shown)


matt-sm:
It appears meson prefers c:\..\ paths.

`Program win_flex found: YES (C:\tmp\766s67rb\execroot\rules_foreign_cc_examples\bazel-out\x64_windows-opt-exec-2B5CBBC6-ST-2324745666c0\bin\external\winflexbison\copy_winflexbison\winflexbison\win_flex.EXE)`

when c:\..\.. in PATH vs

`Program win_flex found: NO`

when c:/../ in PATH

jsharpe:
Closing in favour of #1204

PR #1095, opened 2023-09-19
Advertise CcInfo provider

### LINKED ISSUE #1079: Libraries not linked by cc_shared_library
Since Bazel 6.3.0, the aspect used in `cc_shared_library` only propagates to dependencies that advertise themselves as providing `CcInfo` via the `provides` attribute of the rule constructor. None of these rules do this, so these libraries are not linked by `cc_shared_library`.

This is the same issue as https://github.com/bazelbuild/rules_rust/issues/2101.

calebzulawski:
Fixes #1079


jsharpe (approved):
Thanks!

PR #1098, opened 2023-09-25
set CC and CXX and their flags in meson

### LINKED ISSUE #1073: Meson doesn't pick the correct compiler
CC / CXX are not set and so meson auto detects the compiler rather than using the version set by the configured toolchain.

xytan0056:
- The PR sets CC and CXX in meson to the ones configured in bazel, instead of system's. 
- It also tries to merge and set CFLAGS, CXXFLAGS, LDFLAGS with implicit flags from bazel
- refactoring `split_arguments` helper function to be more readable and able to process arguments with quotes 

Fixes #1073

### DIFF
--- a/examples/third_party/mesa/BUILD.mesa.bazel
+++ b/examples/third_party/mesa/BUILD.mesa.bazel
@@ -39,7 +39,6 @@ meson_with_requirements(
         },
     }),
     copts = select({
-        "@platforms//os:macos": ["-std=c++"],
         "@openssl//:msvc_compiler": ["/std:c++14"],
         "@platforms//os:linux": ["-std=c++14"],
         "//conditions:default": [],
@@ -51,6 +50,7 @@ meson_with_requirements(
             "ws2_32.lib",
         ],
         "@platforms//os:linux": ["-lstdc++"],
+        # "@platforms//os:macos": ["-mlinker-version=305"],
         "//conditions:default": [],
     }),
     options = select({
--- a/examples/third_party/mesa/BUILD.mesa.bazel
+++ b/examples/third_party/mesa/BUILD.mesa.bazel
@@ -28,6 +28,11 @@ meson_with_requirements(
             "@m4//:m4_exe",
         ],
     }),
+    copts = select({
+        "@openssl//:msvc_compiler": ["/std:c++11"],
+        "@platforms//os:linux": ["-std=c++14"],
+        "//conditions:default": [],
+    }),
     env = select({
         "@bazel_tools//src/conditions:host_windows": {
             "PATH": "$$(dirname $$EXT_BUILD_ROOT$$/$(location @winflexbison//:gen_dir))/winflexbison:$$(dirname $$EXT_BUILD_ROOT$$/$(PYTHON3)):$$PATH",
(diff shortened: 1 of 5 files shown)


examples/third_party/mesa/BUILD.mesa.bazel
```diff
@@ -38,7 +38,21 @@ meson_with_requirements(
...
+
     lib_source = ":all_srcs",
+    linkopts = select({
+        "@openssl//:msvc_compiler": [
+            "ws2_32.lib",
+        ],
+        "@platforms//os:linux": ["-lstdc++"],
+        # "@platforms//os:macos": ["-mlinker-version=305"],
+        "//conditions:default": [],
+    }),
```
jsharpe, line 55: I suspect the reason you're having to add these is because CXX is actually passing the c compiler because bazel by default passes `-x c++` explicitly.

examples/third_party/mesa/BUILD.mesa.bazel
```diff
@@ -38,7 +38,21 @@ meson_with_requirements(
...
+        "@platforms//os:linux": ["-std=c++14"],
+        "//conditions:default": [],
+    }),
+
     lib_source = ":all_srcs",
+    linkopts = select({
+        "@openssl//:msvc_compiler": [
+            "ws2_32.lib",
+        ],
+        "@platforms//os:linux": ["-lstdc++"],
```
jsharpe, line 52: what if you're using libc++ via clang? this arg will break a libc++ build with clang.

xytan0056: by setting `CXX` and `CC` both compilers become `/usr/bin/gcc`, which is then used to compile cpp files in CI.
during [linking with cpp file](https://buildkite.com/bazel/rules-foreign-cc/builds/5111#018ae1da-444c-4dab-8efa-0b0d2612129d/6-3800), it hits 
```
undefined reference to `operator new
```
because of missing C++ support runtime library

examples/third_party/mesa/BUILD.mesa.bazel
xytan0056: yes, 
if we don't set CXX/CC, we get `undefined symbol: fcntl64` with zig cc. However, setting them to bazel configured compiler, `/usr/bin/gcc` for both in CI here,  we will need to pass in c++ library

examples/third_party/mesa/BUILD.mesa.bazel
xytan0056: added additional logic to not use `lstdc++`

linzhp (approved):
@jsharpe Can you take another look?

foreign_cc/meson.bzl
```diff
@@ -201,3 +233,11 @@ def meson_with_requirements(name, requirements, **kwargs):
         toolchain = full_label("built_meson_toolchain_for_{}".format(name)),
         **kwargs
     )
+
+# TODO: converge with cmake_script.bzl
```
jheaff1, line 237: Is this TODO to be resolved as part of this PR?

xytan0056: just did another experiment, we can just use the same with 
https://github.com/bazelbuild/rules_foreign_cc/blob/bae11c9a508953c1ec3a564102630878737de99d/foreign_cc/private/cmake_script.bzl#L313-L316

updated

jheaff1 (approved):

xytan0056:
@jsharpe another look?

jheaff1 (approved):

xytan0056:
@jsharpe

PR #1100, opened 2023-10-04
Allow runnable_binary to work on non-external targets

novas0x2a:
Before this change, runnable_binary would only work on targets in external repositories. This is due to a limitation of the rootpath function, with more context in https://github.com/bazelbuild/bazel/issues/10923

This switches the macro to use rlocationpath, which handles this use-case. Using rlocationpath requires bazel 5.4.0, so this also updates the docs accordingly.


jsharpe (approved):
LGTM

PR #1109, opened 2023-10-23
add libsodium to examples

xytan0056:
add libsodium to examples


jsharpe:
Just a general note that in general we're not expanding the examples to build extra projects that don't exercise some unique configuration of the rules as CI in slow as it is and this would only serve to make it slower. Longer term the correct place to contribute this is to the bazel central registry.

xytan0056:
@jsharpe got it. 
We were just trying repro an error in our repository to build libsodium with the same config : 

this is what we get from our repository **only happening on mac os**
```
libtool: warning: '-version-info/-version-number' is ignored for convenience libraries
ar: only one of -a and -[bi] options allowed
usage:  ar -d [-TLsv] archive file ...
        ar -m [-TLsv] archive file ...
        ar -m [-abiTLsv] position archive file ...
        ar -p [-TLsv] archive [file ...]
        ar -q [-cTLsv] archive file ...
        ar -r [-cuTLsv] archive file ...
        ar -r [-abciuTLsv] position archive file ...
        ar -t [-TLsv] archive [file ...]
        ar -x [-ouTLsv] archive [file ...]
```
turns out #928 introduced `cxx_linker_static` to `AR_FLAGS` that has `-static -s`, but `static` is not a viable flag  for `ar`. 

Upon further look, it looks like `libtool` here in examples points to `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/libtool`. But ours points to `../../libtool` in bazel sandbox. 

Do you happen to know how that's configured?

PR #1117, opened 2023-11-09
Allow file targets in native_tool_toolchain()

sitaktif:
Some users might want to specify simple files (such as pre-built tools) in the target attribute of native_tool_toolchain.


UebelAndre (approved):
Thanks!

PR #1127, opened 2023-12-05
Use gazelle to generate bzl_library targets

jsharpe:


jsharpe:
The docs generation seems broken on main too - will investigate in a followup PR (suspect we just need to update stardoc)

jheaff1 (approved):

PR #1134, opened 2023-12-16
Increase resiliency of `runnable_binary`

mishazharov:
In the same vein as https://github.com/bazelbuild/rules_foreign_cc/pull/1093, we want to allow `runnable_binary` to run on targets without shared libraries

Copy and pasted from #1093 

> When using the runnable_binary macro with a binary e.g. diff that does not involve any shared libraries, the following error is produced:
>
> line 46: SHARED_LIBS_ARRAY[@]: unbound variable
>
> As find . -name "*${SHARED_LIB_SUFFIX}" -print0 does not return any results (thus SHARED_LIBS_ARRAY) , along with set -uo earlier in the script, this assumes that there will always be some shared library files.

Also, this PR introduces checks for whether the directory which is specified by `$RUNFILES_DIR` actually exists. If it does not exist, `cd`ing into it will crash the script (even though the rest of the script would have run fine). This is applicable for cases where the binary has runfiles, and is also a runfile itself.

### DIFF
--- a/foreign_cc/private/runnable_binary_wrapper.sh
+++ b/foreign_cc/private/runnable_binary_wrapper.sh
@@ -1,24 +1,5 @@
 #!/usr/bin/env bash
 
-# RUN_UNDER_RUNFILES is set in the "bazel test" environment, where all transitive runfiles are placed into one directory
-# Otherwise, first cd to the runfiles dir for the wrapped executable before searching for shared libraries for the wrapped executable
-if [[ -z $RUN_UNDER_RUNFILES ]]; then
-    SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
-    RUNFILES_DIR_TMP=${SCRIPT_DIR}/SH_BINARY_FILENAME.runfiles
-
-    if [[ -d "$RUNFILES_DIR_TMP" ]]; then
-        RUNFILES_DIR="$RUNFILES_DIR_TMP"
-    fi
-
-    unset RUNFILES_DIR_TMP
-fi
-
-if [[ -n "$RUNFILES_DIR" ]] && [[ -d "$RUNFILES_DIR" ]]; then
-    cd ${RUNFILES_DIR}
-else
-    cd "$(pwd)"
-fi
-
 # --- begin runfiles.bash initialization v2 ---
 # Copy-pasted from the Bazel Bash runfiles library v2. (@bazel_tools//tools/bash/runfiles)
 set -uo pipefail; f=bazel_tools/tools/bash/runfiles/runfiles.bash
@@ -29,6 +10,8 @@ source "$(grep -sm1 "^$f " "$0.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/null
 source "$(grep -sm1 "^$f " "$0.exe.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/null || \
 { echo>&2 "ERROR: cannot find $f"; exit 1; }; f=; set -e
 # --- end runfiles.bash initialization v2 ---
+
+cd "${RUNFILES_DIR}"
 
 EXE=EXECUTABLE
 EXE_PATH=$(rlocation "${EXE#external/}")
@@ -71,5 +54,4 @@ if [ ${#SHARED_LIBS_DIRS_ARRAY[@]} -ne 0 ]; then
 fi
 
 cd - &> /dev/null
-
 exec ${EXE_PATH} "$@"
--- a/foreign_cc/private/runnable_binary_wrapper.sh
+++ b/foreign_cc/private/runnable_binary_wrapper.sh
@@ -15,6 +15,8 @@ fi
 
 if [[ -n "$RUNFILES_DIR" ]] && [[ -d "$RUNFILES_DIR" ]]; then
     cd ${RUNFILES_DIR}
+else
+    cd "$(pwd)"
 fi
 
 # --- begin runfiles.bash initialization v2 ---
@@ -68,8 +70,6 @@ if [ ${#SHARED_LIBS_DIRS_ARRAY[@]} -ne 0 ]; then
     set -u
 fi
 
-if [[ -n "$RUNFILES_DIR" ]] && [[ -d "$RUNFILES_DIR" ]]; then
-    cd - &> /dev/null
-fi
+cd - &> /dev/null
 
 exec ${EXE_PATH} "$@"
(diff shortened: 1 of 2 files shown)


foreign_cc/private/runnable_binary_wrapper.sh
```diff
@@ -4,9 +4,20 @@
 # Otherwise, first cd to the runfiles dir for the wrapped executable before searching for shared libraries for the wrapped executable
 if [[ -z $RUN_UNDER_RUNFILES ]]; then
     SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
-    RUNFILES_DIR=${SCRIPT_DIR}/SH_BINARY_FILENAME.runfiles
+    RUNFILES_DIR_TMP=${SCRIPT_DIR}/SH_BINARY_FILENAME.runfiles
```
jsharpe, line 7: Under what conditions do the runfiles not exist? I have to admit I'm not familiar enough with the details of runfile layout to know if this is correct or not.

@fmeum is this logic correct or are we missing something else runfiles related here?

fmeum: I think that this logic isn't sound, but it also wasn't sound before the PR (not entirely sure though, this is genuinely tricky). `RUNFILES_DIR` must always take precedence if already set, but this logic appears to overwrite based in whether the current script has a runfiles directory, which can result in non-hermetic behavior.

Since the script uses the runfiles library below, could that setup code be moved up? That would provide the correct variables and this script wouldn't have to worry about the details. I also don't see a need to have a special case for `--run_under`.

mishazharov: Thank you for taking a look, I've just removed the special case related to `--run_under`. Since we don't use `--run_under`, I'm not sure if that preserves other users workflows (but it looks like they should continue working)

> Under what conditions do the runfiles not exist?

I've found that using the previous logic would fail if the runnable binary was a data input to another binary. The rules_foreign_cc binary would be present in the runfiles directory, and wouldn't have a runfiles directory itself (since all the runfiles were aggregated in a top level directory)

fmeum: That makes sense. If a binary is a data dependency of some other binary, the outer binary has to use the runfiles libraries to set `RUNFILES_DIR`, which the inner binary can then read and use.

mishazharov:
@jsharpe @fmeum what do you all think about the new revision? A test is failing on windows because of a bad SSL certificate, but I think that's just because I busted the cache:
```
Downloading pcre2 patch from https://wrapdb.mesonbuild.com/v2/pcre2_10.40-3/get_patch
WARNING: failed to download with error: WrapDB connection failed to https://wrapdb.mesonbuild.com/v2/pcre2_10.40-3/get_patch with error <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:997)>.. Trying after a delay...
WARNING: failed to download with error: WrapDB connection failed to https://wrapdb.mesonbuild.com/v2/pcre2_10.40-3/get_patch with error <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:997)>.. Trying after a delay...
WARNING: failed to download with error: WrapDB connection failed to https://wrapdb.mesonbuild.com/v2/pcre2_10.40-3/get_patch with error <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:997)>.. Trying after a delay...
WARNING: failed to download with error: WrapDB connection failed to https://wrapdb.mesonbuild.com/v2/pcre2_10.40-3/get_patch with error <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:997)>.. Trying after a delay...
WARNING: failed to download with error: WrapDB connection failed to https://wrapdb.mesonbuild.com/v2/pcre2_10.40-3/get_patch with error <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:997)>.. Trying after a delay...
A fallback URL could be specified using patch_fallback_url key in the wrap file
```

(although I'm able to successfully download the patch locally, so I'm not too sure what's up there)

foreign_cc/private/runnable_binary_wrapper.sh
```diff
@@ -19,6 +11,8 @@ source "$(grep -sm1 "^$f " "$0.exe.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/
 { echo>&2 "ERROR: cannot find $f"; exit 1; }; f=; set -e
 # --- end runfiles.bash initialization v2 ---
 
+cd "${RUNFILES_DIR}"
```
fmeum, line 14: We may want to catch the case where `RUNFILES_DIR` isn't set or doesn't exist here and show a nice error message (e.g. on Windows, where `RUNFILES_MANIFEST_FILE` would be used instead).

foreign_cc/private/runnable_binary_wrapper.sh
```diff
@@ -19,6 +11,8 @@ source "$(grep -sm1 "^$f " "$0.exe.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/
...
+
 EXE=EXECUTABLE
 EXE_PATH=$(rlocation "${EXE#external/}")
```
fmeum, line 17: If `RUNFILES_DIR` isn't set to an absolute path, then the `EXE_PATH` may be incorrect after the `cd`. This can't easily be caught in tests since those `cd` to the `RUNFILES_DIR` anyway.

mishazharov: Thanks @fmeum, I added the directory existence check, and made the `RUNFILES_DIR` absolute

foreign_cc/private/runnable_binary_wrapper.sh
```diff
@@ -19,6 +11,15 @@ source "$(grep -sm1 "^$f " "$0.exe.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/
 { echo>&2 "ERROR: cannot find $f"; exit 1; }; f=; set -e
 # --- end runfiles.bash initialization v2 ---
 
+if [[ ! -d "${RUNFILES_DIR}" ]]; then
+    >&2 echo "RUNFILES_DIR is set to '${RUNFILES_DIR}' which does not exist";
+    exit 1;
+fi
+
+RUNFILES_DIR=$(realpath ${RUNFILES_DIR})
```
fmeum, line 19: `realpath` isn't available on macOS. Is this script used on Linux only?

mishazharov: Nice, I didn't know that. I've replaced it with a cross platform variant. Thanks!

fmeum (approved):

jsharpe (approved):

Issue #1135, opened 2023-12-16
Several builds under examples/third_party fail at HEAD


bcsgh:
The stuff under `examples/third_party` looks like it could be a highly valuable resource, but I'm not actually sure *any* of it works. Which is annoying.

# Reproduction:
```
T=rules_foreign_cc
git clone https://github.com/bazelbuild/rules_foreign_cc.git $T
cd $T/examples/third_party

bazel build -k :all
# More complete:
bazel build -k ...
# Most (likely too)  complete:
bazel query 'kind("http_archive|git_repository",//external:*)' |
    sed "/:.*[/]/d;s/.*:/@/;s~$~//:all~" |
    bazel build -k $(cat)
```

# ERRORS

## `@rules_perl//` (needed by `@openssl//`)
```
ERROR: [...]/external/rules_perl/BUILD:74:23: While resolving toolchains for target @rules_perl//:current_toolchain: No matching toolchains found for types @rules_perl//:toolchain_type.
To debug, rerun with --toolchain_resolution_debug='@rules_perl//:toolchain_type'
```

## `@mesa//`
```
Error in download_and_extract: java.io.IOException: Error downloading [https://archive.mesa3d.org//mesa-22.1.4.tar.xz] to [...]/external/mesa/temp12780688070917179032/mesa-22.1.4.tar.xz: No route to host (Host unreachable)
```

## `@libpng//`
```
ERROR: [...]/external/libpng/BUILD.bazel:13:6: Foreign Cc - CMake: Building libpng failed:
[...]
external/libpng/contrib/libtests/pngstest.c:3216:24: error: '%d' directive writing between 1 and 11 bytes into a region of size between 10 and 32 [-Werror=format-overflow=]
 3216 |       sprintf(name, "%s%d.png", tmpf, ++counter);
      |                        ^~
[...]
```

## `@gperftools//`
```
ERROR: [...]/external/gperftools/BUILD.bazel:11:15: Foreign Cc - Configure: Building gperftools_build failed:
[...]
external/gperftools/src/tcmalloc.cc:2098:36: error: 'void tc_delete_aligned_nothrow(void*, std::align_val_t, const std::nothrow_t&)' alias between functions of incompatible types 'void(void*, std::align_val_t, const std::nothrow_t&)' and 'void(void*)' [-Werror=attribute-alias=]
[...]

PR #1137, opened 2023-12-17
Fix `//openssl/...` by calling `openssl_setup()`

bcsgh:
This is the proximate cause that resulted in https://github.com/bazelbuild/rules_foreign_cc/issues/1135

### DIFF
--- a/examples/third_party/WORKSPACE.bazel
+++ b/examples/third_party/WORKSPACE.bazel
@@ -17,3 +17,7 @@ local_repository(
 load("//:repositories.bzl", "repositories")
 
 repositories()
+
+load("//:setup.bzl", "setup")
+
+setup()


bcsgh:
I looks like the failing checks are failing at HEAD?

jsharpe (approved):
We test CI from the examples WORKSPACE rather than the examples/third_party WORKSPACE so its not surprising that examples/third_party doesn't work directly.

Happy to approve any PRs that get examples/third_party working standalone but do bear in mind that now bazel 7.0.0 has been released this will be primarily migrating towards a MODULE.bazel file which won't have the same shortcomings.

bcsgh:
Until `WORKSPACE` support is removed (as in can't be turned on) from all still supported version of bazel, keeping official examples working that way should be (IMHO) an official goal.

Sure people **should** move over, but if you push _too_ hard on that, some people will see switching to a build tool that is less demanding about its opinions start looking just as easy in the short term and less hassle in the long run. In my case, if I'd know how much I'd need to fix I'd likely have deferred upgrading to 7.0.0 for another 2-3 months and there definitely isn't room in the project's calendar to switch to `MODULE.bazel` until then.

PR #1148, opened 2023-12-22
Build msvc cmake with prebuilt cmake

matt-sm:
On Windows using the msvc compiler, CMake can only be built using an existing binary CMake:

https://github.com/Kitware/CMake/blob/master/README.rst#windows

This means the existing `cmake_tool` fails to build for msvc (it tries to pass mingw compiler flags).  This change enables CMake to be built using a preinstalled CMake.

This change follows the pattern for pkg-config whereby unix/msvc have different build rules.

### DIFF
--- a/foreign_cc/built_tools/cmake_build.bzl
+++ b/foreign_cc/built_tools/cmake_build.bzl
@@ -1,14 +1,40 @@
-""" Rule for building CMake from sources. """
+""" Macro for building CMake from sources. """
 
-load("//foreign_cc:defs.bzl", "configure_make")
+load("//foreign_cc:defs.bzl", "cmake_variant", "configure_make")
 
 def cmake_tool(name, srcs, **kwargs):
+    """Macro for building CMake
+
+    Args:
+        name: A unique name for this target
+        srcs: The target containing the build tool's sources
+        **kwargs: Remaining args
+    """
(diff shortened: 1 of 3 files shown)


matt-sm:
@jsharpe is this something you would consider merging?

foreign_cc/built_tools/cmake_build.bzl
```diff
@@ -1,14 +1,40 @@
...
+            ":msvc_compiler": "{}_msvc".format(name),
+            "//conditions:default": "{}_default".format(name),
+        }),
+    )
+
+    cmake_variant(
+        name = "{}_msvc".format(name),
+        # to prevent errors with incompatible _WIN32_WINNT in cmlibarchive
+        # override NTDDI_VERSION to match _WIN32_WINNT set in the default cc_toolchain
+        copts = ["/D NTDDI_VERSION=0x06010000"],
```
jsharpe, line 27: I'm not familiar enough with Windows to understand why this is needed but going to trust you on this one!

jsharpe (approved):
Thanks, this looks good to me.

foreign_cc/built_tools/cmake_build.bzl
matt-sm: Yeah I would very much have liked not to include this but there is definitely a error with this package and the default toolchain.

PR #1151, opened 2024-01-01
Allow `prefix` to be set as an absolute path in `configure`

### LINKED ISSUE #823: make install fails when --prefix needs to be an absolute path
I am trying to build nginx from sources. With nginx, it is fairly standard to set `--prefix` to an absolute path. For example, the default value is `/usr/local/nginx`.

http://nginx.org/en/docs/configure.html

If I try to use `install_prefix` to do this, rules_foreign_cc still appends `$$BUILD_TMPDIR$$` to my install_prefix. This will make nginx think its config and such are in /some/long/bazel/tmpdir/usr/local/nginx.

It seems that the correct tool to use in this case is DESTDIR. I want to be able to set --prefix like any other arg I pass to configure, then use DESTDIR to install it in the tmp dir.

Here is a patch that seems to accomplish what I want:
https://gist.github.com/aptenodytes-forsteri/99f84a715b481db59442cde579c432b1

Here is an example of building nginx from sources:

```
filegroup(
    name = "srcs",
    srcs = glob(["**/*"]),
)

configure_make(
    name = "nginx_all",
    args = ["--debug"],
    configure_in_place = True,
    configure_options = [
       # Note that it is standard to have absolute paths here, so --prefix doesn't work.
        "--sbin-path=/usr/sbin/nginx",
        "--modules-path=/usr/lib/nginx/modules",
        "--conf-path=/etc/nginx/nginx.conf",
        "--pid-path=/var/run/nginx/nginx.pid",
        "--lock-path=/var/run/nginx/nginx.lock",
       # Redacted - some other flags to include nginx modules
    ],
    lib_source = ":srcs",
    # This matches the --sbin-path flag above
    out_bin_dir = "usr/sbin",
    out_binaries = ["nginx"],
   # This is the part I added in my patch.
    override_install_prefix = "/usr/share/nginx"
)
```

cc: @alexeagle

mishazharov:
This PR allows setting the `--prefix` flag to an absolute path for `configure`. When `make` is invoked, targets that start with `install` or `uninstall` are given a `DESTDIR` flag which instructs `make` to place the install output into the directory that the rules expect the installed files to be in.

Since there could be some shortfalls with this approach of setting the `prefix` flag (namely if `make` doesn't support `DESTDIR` for some reason) this new option is specified as "experimental"

Closes https://github.com/bazelbuild/rules_foreign_cc/issues/823

### DIFF
--- a/foreign_cc/configure.bzl
+++ b/foreign_cc/configure.bzl
@@ -78,12 +78,19 @@ def _create_configure_script(configureParameters):
     configure_prefix = "{} ".format(expand_locations_and_make_variables(ctx, ctx.attr.configure_prefix, "configure_prefix", data)) if ctx.attr.configure_prefix else ""
     configure_options = [expand_locations_and_make_variables(ctx, option, "configure_option", data) for option in ctx.attr.configure_options] if ctx.attr.configure_options else []
 
+    if ctx.attr.experimental_absolute_install_prefix:
+        destdir_arg = 'DESTDIR="$$BUILD_TMPDIR$$/" '
+    else:
+        destdir_arg = ""
+
     for target in ctx.attr.targets:
         # Configure will have generated sources into `$BUILD_TMPDIR` so make sure we `cd` there
         make_commands.append("{prefix}{make} {target} {args}".format(
             prefix = prefix,
             make = attrs.make_path,
-            args = args,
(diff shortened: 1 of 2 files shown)


mishazharov:
@jsharpe Can you please take a look at this PR if you have a chance? It's simplified some of our build setup because binaries aren't confused about where they're being executed from

PR #1158, opened 2024-01-12
Allow configuring meson, make and pkgconfig versions via bzlmod exten…

jsharpe:
…sion

### DIFF
--- a/foreign_cc/extensions.bzl
+++ b/foreign_cc/extensions.bzl
@@ -1,51 +1,69 @@
 """Entry point for extensions used by bzlmod."""
 
-load("//foreign_cc:repositories.bzl", "rules_foreign_cc_dependencies")
-load("//toolchains:prebuilt_toolchains.bzl", "prebuilt_toolchains")
-
-_DEFAULT_CMAKE_VERSION = "3.23.2"
-_DEFAULT_NINJA_VERSION = "1.11.1"
+load("//foreign_cc:repositories.bzl", "rules_foreign_cc_dependencies", "DEFAULT_CMAKE_VERSION", "DEFAULT_MAKE_VERSION", "DEFAULT_MESON_VERSION", "DEFAULT_NINJA_VERSION", "DEFAULT_PKGCONFIG_VERSION")
+load("//toolchains:built_toolchains.bzl", "make_toolchain", "meson_toolchain", "pkgconfig_version")
+load("//toolchains:prebuilt_toolchains.bzl", "cmake_toolchains", "ninja_toolchains")
 
 cmake_toolchain_version = tag_class(attrs = {
-    "version": attr.string(doc = "The cmake version", default = _DEFAULT_CMAKE_VERSION),
+    "version": attr.string(doc = "The cmake version", default = DEFAULT_CMAKE_VERSION),
+})
+
+make_toolchain_version = tag_class(attrs = {
+    "version": attr.string(doc = "The GNU Make version", default = DEFAULT_MAKE_VERSION),
+})
+
+meson_toolchain_version = tag_class(attrs = {
+    "version": attr.string(doc = "The meson version", default = DEFAULT_MESON_VERSION),
 })
 
 ninja_toolchain_version = tag_class(attrs = {
-    "version": attr.string(doc = "The ninja version", default = _DEFAULT_NINJA_VERSION),
+    "version": attr.string(doc = "The ninja version", default = DEFAULT_NINJA_VERSION),
+})
+
+pkgconfig_toolchain_version = tag_class(attrs = {
+    "version": attr.string(doc = "The pkgconfig version", default = DEFAULT_PKGCONFIG_VERSION),
 })
 
 def _init(module_ctx):
+    for mod in module_ctx.modules:
+        if mod.is_root:
+            for toolchain in mod.tags.cmake:
+                cmake_toolchains(toolchain.version, register_toolchains = False)
+
+            for toolchain in mod.tags.make:
+                make_toolchain(toolchain.version, register_toolchains = False)
+
+            for toolchain in mod.tags.meson:
+                meson_toolchain(toolchain.version, register_toolchains = False)
+
+            for toolchain in mod.tags.ninja:
+                ninja_toolchains(toolchain.version, register_toolchains = False)
+
+            for toolchain in mod.tags.pkgconfig:
+                pkgconfig_toolchain(toolchain.version, register_toolchains = False)
+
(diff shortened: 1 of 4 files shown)


foreign_cc/meson.bzl
```diff
@@ -208,8 +208,8 @@ def meson_with_requirements(name, requirements, **kwargs):
 
     meson_tool(
         name = "meson_tool_for_{}".format(name),
-        main = "@meson_src//:meson.py",
-        data = ["@meson_src//:runtime"],
+        main = "@meson_1.1.1_src//:meson.py",
+        data = ["@meson_1.1.1_src//:runtime"],
```
jsharpe, line 212: @jheaff1 any ideas how I could avoid this hardcoded version number?
I want the end user to not have to do use_repo(tools, "meson_src") if they're using this macro which they currently have to; maybe we should make the toolchain variants have to be declared via a module extension and then generate this toolchain into the toolchain_hub e.g.
```
tools.meson(version = "1.1.1")
tools.meson(version = "1.1.1", requirements = ["@pypi__numpy"])
```
?

jsharpe:
@jheaff1 @UebelAndre One part of this I'm yet to design is what the module extension should look like for selecting between preinstalled tools and prebuilt / built from source tools. Any ideas?

The extension API currently looks like this:
```
tools.cmake(version = "3.29.1")
tools.ninja(version = "1.12.0")
```

we could do something like:

```
tools.cmake(version = "3.29.1", allow_build_from_source = False)
tools.ninja(version = "1.11.1", use_prebuilt_binaries = False)
tools.pkgconfig(preinstalled_version = True)
```
but very much open to ideas on this.

examples/MODULE.bazel
```diff
@@ -7,17 +7,16 @@ local_path_override(
...
 
 bazel_dep(name = "platforms", version = "0.0.11")
-bazel_dep(name = "rules_rust", version = "0.56.0")
-bazel_dep(name = "rules_swift", version = "1.6.0", repo_name = "build_bazel_rules_swift")
+bazel_dep(name = "rules_swift", version = "1.17.0", repo_name = "build_bazel_rules_swift")
 bazel_dep(name = "rules_apple", version = "3.4.0", repo_name = "build_bazel_rules_apple")
-bazel_dep(name = "apple_support", version = "1.12.0", repo_name = "build_bazel_apple_support")
-bazel_dep(name = "rules_python", version = "1.2.0")
+bazel_dep(name = "apple_support", version = "1.14.0", repo_name = "build_bazel_apple_support")
+bazel_dep(name = "rules_python", version = "0.23.1")
```
jsharpe, line 17:
```suggestion
```

jsharpe:
> Out of curiosity, did this PR die? I was hoping to use a new version of Cmake but it looks like this PR was abandoned.

Not dead, its still on my todo list but haven't had the time to work through all the failures in CI!

jwhpryor:
Ah wow, truthfully I was not expecting this thread to get revived, I really appreciate you taking a look!  I would love to understand how to use your work when it lands. No pressure, but thanks for your effort, and I look forward to it landing!

test/BUILD.bazel
```diff
@@ -9,11 +9,11 @@ load(":utils_test.bzl", "utils_test_suite")
...
-    "//toolchains/private:cmake_tool",
-    "//toolchains/private:make_tool",
-    "//toolchains/private:meson_tool",
-    "//toolchains/private:ninja_tool",
-    "//toolchains/private:pkgconfig_tool",
+    # "//toolchains/private:cmake_tool",
+    # "//toolchains/private:make_tool",
+    # "//toolchains/private:meson_tool",
+    # "//toolchains/private:ninja_tool",
+    # "//toolchains/private:pkgconfig_tool",
```
jsharpe, line 16: Need a strategy to handle these targets - maybe these tests should be generated in the toolchain hub now?

foreign_cc/meson.bzl
novas0x2a: I'm far from an expert on bzlmod (just started trying to port our stuff) but my observed-so-far impression of best practice here is that use_all_repos is the best you can do: https://github.com/bazel-contrib/rules_foreign_cc/pull/1420

(If you reuse the same repo name for multiple different versions, my observation is that it causes bazel to delete/recreate the repo; this isn't necessarily a problem for development on HEAD, but if you maintain multiple support branches, it avoids the need for bazel to do extra work when you switch between them if the repo name contains the version).

novas0x2a:
Came over here from https://github.com/bazel-contrib/rules_foreign_cc/pull/1418#issuecomment-2993800614; I'm definitely not an expert on bzlmod, but in the (internal) repo I'm currently converting it to bzlmod, this MR appears to work (and is slightly easier if you combine it with #1420)

MODULE.bazel
```diff
@@ -24,25 +24,20 @@ python.toolchain(python_version = "3.12")
...
 )
 
 register_toolchains(
-    "@rules_foreign_cc_framework_toolchains//:all",
-    "@cmake_3.31.8_toolchains//:all",
-    "@ninja_1.13.0_toolchains//:all",
     "@python_3_12//:all",
+    "@prebuilt_cmake_toolchains//:all",
+    "@prebuilt_ninja_toolchains//:all",
+    "@rules_foreign_cc_framework_toolchains//:all",
```
novas0x2a, line 40: I'm actually wondering if some of these should be marked as dev_dependencies instead; If I were building my own cmake, for example, i definitely wouldn't want the prebuilt toolchain to be registered at all. To give you an idea of what we do today in our workspace file (pre-migration to bzlmod, since bzlmod can't match this yet...)
```
rules_foreign_cc_dependencies(
    cmake_version = "3.31.7",
    make_version = "4.4.1",
    ninja_version = NINJA_VERSION,
    register_built_pkgconfig_toolchain = False,
    register_preinstalled_tools = False,
)

register_toolchains(
    "//thirdparty:pkgconfig_noop_toolchain",
    "//thirdparty:autoconf_noop_toolchain",
    "//thirdparty:automake_noop_toolchain",
    "//thirdparty:m4_noop_toolchain",
)
```
To avoid host divergence, we disable every host tool we can and replace them with noop versions (the automake one sets `ACLOCAL=false AUTOMAKE=false`, for example).

The infrastructural framework toolchains make sense to register unconditionally, I think, but it seems like anything that is affected by configuration at all should only be registered from the root module. (I also should reiterate that I don't really have any idea what "good" looks like for bzlmod, and you shouldn't necessarily take anything I say about it as informed :))

PR #1165, opened 2024-01-25
Fix tool path for nmake

jsun-splunk:
There is a problem (or feature) with [nmake](https://learn.microsoft.com/en-us/cpp/build/reference/nmake-reference?view=msvc-170) where if you don't wrap the path for tools like CC in double quotes, nmake will only print out the command instead of executing it.

For example, given the following makefile:

```
CC = C:/my/toolchain/path/cl.exe

foo: main.cpp
 $(CC) main.cpp
```

running nmake outputs:

```
F:\USERS\jsun\test>nmake

Microsoft (R) Program Maintenance Utility Version 14.16.27051.0
Copyright (C) Microsoft Corporation.  All rights reserved.

        C:/my/toolchain/path/cl.exe main.cpp
```

I believe this problem hasn't surfaced till now as it is already quoting tool path on windows if they have a space in it. [ref](https://github.com/bazelbuild/rules_foreign_cc/blob/main/foreign_cc/private/make_env_vars.bzl#L120-L121)

We encountered this issue as we are using our own msvc toolchain and it did not have any space in the tool path.

This problem also surfaces when rules_foreign_cc builds its own pkg-config. i.e. 

```
bazel build @rules_foreign_cc//toolchains/private:pkgconfig_tool
```

This PR attempts to fix this by checking if nmake is used as a part of make_commands and if so wrap the tool path in quotes.

### DIFF
--- a/foreign_cc/private/make_env_vars.bzl
+++ b/foreign_cc/private/make_env_vars.bzl
@@ -10,8 +10,9 @@ def get_make_env_vars(
         flags,
         user_vars,
         deps,
-        inputs):
-    vars = _get_make_variables(workspace_name, tools, flags, user_vars)
+        inputs,
+        make_commands = []):
+    vars = _get_make_variables(workspace_name, tools, flags, user_vars, make_commands)
     deps_flags = _define_deps_flags(deps, inputs)
 
     # For cross-compilation.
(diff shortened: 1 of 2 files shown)


jsharpe (approved):
LGTM; thanks!

PR #1167, opened 2024-01-25
Add make 4.4.1

novas0x2a:
It was released a while ago, and, although I haven't been able to track down the exact reason why, I keep seeing jobserver failures (on linux) that don't happen on either 4.3 or 4.4.1, leading me to believe there's probably a bug in the new 4.4 jobserver code that was fixed in 4.4.1.

### DIFF
--- a/foreign_cc/repositories.bzl
+++ b/foreign_cc/repositories.bzl
@@ -10,7 +10,7 @@ def rules_foreign_cc_dependencies(
         native_tools_toolchains = [],
         register_default_tools = True,
         cmake_version = "3.23.2",
-        make_version = "4.4",
+        make_version = "4.4.1",
         ninja_version = "1.11.1",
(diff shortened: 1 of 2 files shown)


jsharpe (approved):
Thanks!

novas0x2a:
One thing to note is this will emit a warning until make-4.4.1 is uploaded to the bazel mirror; are you able to do that?

jsharpe:
> One thing to note is this will emit a warning until make-4.4.1 is uploaded to the bazel mirror; are you able to do that?

There's an issue template in bazelbuild/bazel to request a new package be added to the mirror. Anyone can fill that out.

novas0x2a:
Ah, didn't realize that. Filed: https://github.com/bazelbuild/bazel/issues/21039

PR #1169, opened 2024-02-06
Cater for different drive letters when normalizing path

matt-sm:
` _normalize_path()` fails to normalize paths containing different drive letters eg.

`C:\foo;F:\bar` is converted to `/c/foo:F:\bar`

This change will ensure paths containing different drive letters are also normalized eg.

`C:\foo;F:\bar` is converted to `/C/foo:/F/bar`

### DIFF
--- a/foreign_cc/private/framework.bzl
+++ b/foreign_cc/private/framework.bzl
@@ -661,10 +661,10 @@ def _print_env():
     ]
 
 def _normalize_path(path):
-    # Change Windows style paths to Unix style. E.g. change "C:" to "/c"
+    # Change Windows style paths to Unix style.
     if path[0].isalpha() and path[1] == ":":
-        path = path.replace(path[0:2], "/" + path[0].lower())
-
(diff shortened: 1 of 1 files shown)


matt-sm:
@jsharpe is this something you would consider merging?

jsharpe:
> @jsharpe is this something you would consider merging?

Yes; but its annoying that windows CI is getting SSL certificate errors on Windows so that we can't easily verify this PR; its been a while since I've had to develop on windows, and I've never used bazel on Windows so we definitely rely on contributions from the community that do use windows to enhance the capability on that platform!

matt-sm:
What is the fix for the CI certificate errors?

jsharpe:
> What is the fix for the CI certificate errors?

I don't know I'm afraid - not sure if this is infra related, bazel 7, or a problem in our rules..

mishazharov:
I took a quick look and was able to fix the CI issue: https://github.com/bazelbuild/rules_foreign_cc/pull/1171

jsharpe (approved):
LGTM thanks!

PR #1176, opened 2024-02-23
Add aarch64 to prebuild library constraints for macos

dmeijboom:
At least since Ninja 1.10.2 the macos binaries are actually universal binaries which means that they support both `x86_64` as well as `aarch64`. I tested all versions manually to make sure we only mark ninja binaries as universal binaries when it's build for both architectures.

See:

```
➜ file ~/Downloads/ninja
/Users/dmeijboom/Downloads/ninja: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64:Mach-O 64-bit executable arm64]
~/Downloads/ninja (for architecture x86_64):     Mach-O 64-bit executable x86_64
~/Downloads/ninja (for architecture arm64):      Mach-O 64-bit executable arm64
```

This PR adds a new target `mac_aarch64` which contains the same artifact as `x86_64` where the CPU constraint is changed to `aarch64`.

Tested and seems to work just fine.

### DIFF
--- a/toolchains/prebuilt_toolchains.bzl
+++ b/toolchains/prebuilt_toolchains.bzl
@@ -8303,6 +8303,20 @@ def _ninja_toolchains(version, register_toolchains):
 
         maybe(
             http_archive,
+            name = "ninja_1.11.1_mac_aarch64",
+            urls = [
+                "https://github.com/ninja-build/ninja/releases/download/v1.11.1/ninja-mac.zip",
+            ],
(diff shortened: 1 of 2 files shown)


jsharpe (approved):
LGTM! Thanks for this.

jsharpe:
@dmeijboom if you update your branch with main then this should auto-merge once its run through CI.

dmeijboom:
> @dmeijboom if you update your branch with main then this should auto-merge once its run through CI.

Thanks, done!

PR #1180, opened 2024-03-07
adjust relative msvc paths

matt-sm:
RFCC relies on the PATH and INCLUDE msvc env vars to find the compiler/linker and includes files. 

When using an msvc toolchain with only relative paths, many of the foreign_cc rules break due to a reliance on absolute paths.

This MR prepends any relative paths in the msvc PATH and INCLUDES with `$EXT_BUILD_ROOT` to ensure all paths are still absolute without breaking hermeticity.

### DIFF
--- a/foreign_cc/private/framework.bzl
+++ b/foreign_cc/private/framework.bzl
@@ -304,7 +304,8 @@ def get_env_prelude(ctx, lib_name, data_dependencies, target_root):
     env = dict()
 
     # Add all environment variables from the cc_toolchain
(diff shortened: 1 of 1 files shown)


matt-sm:
@jsharpe I'm not sure if you have any brighter ideas

also wondering how these changes are related to the buildkite failures

jsharpe (approved):
This seems a reasonable workaround - the only other approach would be some form of wrapper scripts for the toolchains.
Looks like previous CI failures were infrastructure flakes as it seems to have run through ok now.
Apologies in the slow review of this change!

PR #1184, opened 2024-03-22
Use a sed delimiter that is unlikely to appear in the input

### LINKED ISSUE #867: Problems when outputUserRoot has @-sign
The problematic line is https://github.com/bazelbuild/rules_foreign_cc/blob/0.7.1/foreign_cc/private/framework/toolchains/linux_commands.bzl#L77.

I found this out when building with remote cache turned on (gcloud in this use-case). Here, `outputUserRoot` looks like `outputRoot/_bazel_foo@bar.com` and the substitution fails. It's tricky to debug as in the terminal you can only see something as
```
sed: -e expression #1, char 207: unknown option to `s'
```

A simple workaround is to define e.g. `--output_base=/tmp/bazel/output`.

mkauf:
Fixes #867


jsharpe (approved):
Thanks for doing this!

PR #1187, opened 2024-03-31
fix macos build

zirain:
fix: #1186

### DIFF
--- a/foreign_cc/built_tools/pkgconfig_build.bzl
+++ b/foreign_cc/built_tools/pkgconfig_build.bzl
@@ -75,6 +75,7 @@ def pkgconfig_tool(name, srcs, **kwargs):
         **kwargs
     )
 
+    kwargs.pop("env", {})
     make_variant(
         name = "{}_msvc_build".format(name),
         lib_source = srcs,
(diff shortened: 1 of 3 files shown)


foreign_cc/built_tools/pkgconfig_build.bzl
```diff
@@ -18,6 +22,9 @@ def _pkgconfig_tool_impl(ctx):
         "%s install" % make_data.path,
     ]
 
+    if os_name(ctx) == "macos":
+        script[0] = "CFLAGS=-Wno-int-conversion CXXFLAGS=-Wno-int-conversion ./configure --with-internal-glib --prefix=$$INSTALLDIR$$"
```
jsharpe, line 26: The problem with this command is it completely overrides CFLAGS and CXXFLAGS which means the flags from the exec configuration toolchain will be ignored. I suspect this will be better done by a select in the BUILD file for pkgconfig that conditionally adds these as copts so that any other flags are maintained when building the pkg config tool.

zirain: updated, not sure if it's the right direction.

zirain:
people who meet this, can use following workaround:

```
build:macos --host_action_env=CXXFLAGS=-Wno-int-conversion
build:macos --action_env=CXXFLAGS=-Wno-int-conversion
build:macos --host_action_env=CFLAGS=-Wno-int-conversion
build:macos --action_env=CFLAGS=-Wno-int-conversion
```

PR #1188, opened 2024-04-03
Fix - Copy Dir ignores hidden files on macOS

LaurenceTews:
The copy_dir_contents_to_dir command currently ignores hidden files (e.g. dotfiles), leading to missing files after copying. Removing the * character in this case maintains the same behaviour, only with the inclusion of hidden files (.clang-tidy, .gitignore, etc).


jsharpe (approved):
Thanks; I suspect that the freebsd toolchain also has the same issue.

PR #1199, opened 2024-04-30
Exclude filenames with spaces from cmake toolchains

illicitonion:
Without this, on macOS runfiles fail to build because of `cmake-3.23.2-macos-universal/doc/cmake/html/_sources/generator/Borland Makefiles.rst.txt`


jsharpe (approved):
LGTM - I suspect we can exclude the entire doc directory TBH but this catches any future files elsewhere in the tree too..

PR #1202, opened 2024-05-16
WIP: provisional parallelization support

novas0x2a:
This is a proof of concept for one potential (limited) version of parallelization which doesn't require complex implementation and doesn't put the number of CPUS into the cache key (and break caching across machines, the way using `--action_env=CMAKE_BUILD_PARALLEL_LEVEL`). This is not fully implemented in this MR, known things that are missing:
1) Windows and FreeBSD don't have impls because I'm not currently sure what to put there.
2) The ability to enable and disable this on a per-rule basis is not currently implemented and is probably necessary (since otherwise, this would be a bit of a compat break, and it's not uncommon for projects to break under parallelization). I'm open to suggestions on the best way to plumb this.

This might eventually implement a simplified form of https://github.com/bazelbuild/rules_foreign_cc/issues/329

@jsharpe I know you did some similar work on this in https://github.com/bazelbuild/rules_foreign_cc/issues/848; do you think an approach like this would be mergeable (if we figured out good ways of doing the TODOs above?)

### DIFF
--- a/foreign_cc/private/framework/toolchains/linux_commands.bzl
+++ b/foreign_cc/private/framework/toolchains/linux_commands.bzl
@@ -38,6 +38,13 @@ def enable_tracing():
 def disable_tracing():
     return "set +x"
 
+def enable_parallel_build():
+    return """
+    _cpu_count=$(nproc)
+    export MAKEFLAGS="-j$_cpu_count${MAKEFLAGS:+ }${MAKEFLAGS:-}"
+    export CMAKE_BUILD_PARALLEL_LEVEL="$_cpu_count"
+"""
+
 def mkdirs(path):
     return "mkdir -p " + path
 
@@ -264,6 +271,7 @@ commands = struct(
     define_sandbox_paths = define_sandbox_paths,
     disable_tracing = disable_tracing,
     echo = echo,
+    enable_parallel_build = enable_parallel_build,
     enable_tracing = enable_tracing,
     env = env,
     export_var = export_var,
--- a/foreign_cc/boost_build.bzl
+++ b/foreign_cc/boost_build.bzl
@@ -30,6 +30,7 @@ def _create_configure_script(configureParameters):
         "##copy_dir_contents_to_dir## $$EXT_BUILD_ROOT$$/{}/. .".format(root),
         "chmod -R +w .",
         "##enable_tracing##",
+        "##enable_parallel_build##",
         "./bootstrap.sh {}".format(" ".join(ctx.attr.bootstrap_options)),
         "./b2 install {} --prefix=.".format(" ".join(user_options)),
         "##disable_tracing##",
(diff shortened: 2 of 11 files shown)


foreign_cc/private/framework/toolchains/linux_commands.bzl
```diff
@@ -38,6 +38,13 @@ def enable_tracing():
 def disable_tracing():
     return "set +x"
 
+def enable_parallel_build():
+    return """
+    _cpu_count=$(nproc)
+    export MAKEFLAGS="-j$_cpu_count${MAKEFLAGS:+ }${MAKEFLAGS:-}"
```
jsharpe, line 44: I'd also suggest setting `-l$_cpu_count` here just to be a good citizen to the machine as bazel has only assigned a single core to the action in its scheduling and so limiting the parallelism under load is a useful thing to do.
I'm not familiar enough with CMake's parallelism config to know if there's something equivalent there?

foreign_cc/private/framework/toolchains/linux_commands.bzl
```diff
@@ -38,6 +38,13 @@ def enable_tracing():
 def disable_tracing():
     return "set +x"
 
+def enable_parallel_build():
+    return """
+    _cpu_count=$(nproc)
```
jsharpe, line 43: This is a slightly naiive way of getting a cpu account as nproc is not cgroup aware AFAIK.
When running under RBE its likely that you will be constrained in a cgroup and so this added parallelism there will likely actually result in slower builds in those environments. I've yet to see a good way of interrogating cgroups to work out how many processes you actually want to spawn in a reliable way.

jsharpe (commented):
The main challenge I see is that all current approaches to parallelism over provision the resources that bazel has assigned to the action; this is often fine in single user systems but when you are on a multi-tenant system or RBE then this is not being a good citizen of the resources.
I think if we were to land something this it would have to be opt-in via a config_setting to turn it on, and then overridable on a per target basis.
I'm also not that thrilled at the idea of adding to the shell toolchain; longer term the shell toolchain should probably be replaced by a cross-platform wrapper tool - TBH I think I prefer the approach of placing this logic in tool wrappers rather than centrally - this means that any consumers of e.g. make would be able to get the parallel invocation semantics without needing support at the callsite; this also solves the opt-in problem as it just becomes a toolchain selection and so then there is no extra attributes or config settings required.
Incidently ninja defaults to parallel builds, we likely should be explicitly wrapping that and limiting it to 1 thread but defaulting CMake builds to the Ninja generator is a quick hack to get parallel builds.

Looking at https://github.com/bazelbuild/bazel/issues/10443 though -i think resource_sets are probably the correct way to pass parallelism info to bazel but only allows a fixed size, and not a dynamically allocated value based on how much resource is currently available.

PR #1204, opened 2024-05-22
fix: change cygpath to mixed mode

jsun-splunk:
'autoconf' applications often has a ./configure stage that checks the $srcdir (source directory). This is represented in foreign_cc by a combination of $EXT_BUILD_ROOT and a relative path to the source. This check fails on windows if $EXT_BUILD_ROOT contains '\\' since it is an escape char under unix. Therefore, we need to generate $EXT_BUILD_ROOT with '/' instead.

Generating $EXT_BUILD_ROOT with '/' present a different problem as now the resulting PATH variable becomes the format "/C:/path/to/a:/C:/path/to/b". Due to ':' being used as both window drive leter assignment and linux path delimiter, MSYS2 gets very confused and doesn't resolve this correctly. Therefore, we need this to become the format "/C/path/to/a:/C/path/to/b".

$EXT_BUILD_ROOT is sometimes passed through the 'env' attribute by the user. We need to handle this scenario as well.

### DIFF
--- a/foreign_cc/private/framework.bzl
+++ b/foreign_cc/private/framework.bzl
@@ -327,6 +327,12 @@ def get_env_prelude(ctx, lib_name, data_dependencies, target_root):
     user_vars = expand_locations_and_make_variables(ctx, ctx.attr.env, "env", data_dependencies)
     env.update(user_vars)
 
+    if cc_toolchain.compiler == "msvc-cl":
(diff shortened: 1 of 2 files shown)


jsun-splunk:
Hi @jsharpe,

Could I trouble you to review?

jsun-splunk:
This MR has the same purpose as https://github.com/bazelbuild/rules_foreign_cc/pull/1091 but also fixes the meson issue.

jsharpe (approved):
LGTM

PR #1207, opened 2024-06-06
fix: path expansion on windows

jsun-splunk:
This is an extension of https://github.com/bazelbuild/rules_foreign_cc/pull/1204.

PATH variable updated by the `##path##` primitive does not get expanded correctly on windows. 

For example:

```
##path## $$EXT_BUILD_ROOT$$
```

gets expanded to 

```
export PATH=C:/users/me/bazel_cache/xxxxxx/execroot/myrepo:$PATH
```

At the beginning, the `C:` gets incorrectly interpreted as 2 paths under msys.

1. `C`
2. `/users/me/bazel_cache/xxxxxx/execroot/myrepo`

The desired expansion here is:

```
export PATH=/C/users/me/bazel_cache/xxxxxx/execroot/myrepo:$PATH
```

Which correctly gets interpreted as 1 path.

### DIFF
--- a/foreign_cc/private/framework/toolchains/windows_commands.bzl
+++ b/foreign_cc/private/framework/toolchains/windows_commands.bzl
@@ -27,7 +27,25 @@ def env():
     return "env"
 
 def path(expression):
+    # In windows we cannot add vars like $$EXT_BUILD_DEPS$$ directly to PATH as PATH is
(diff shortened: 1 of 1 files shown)


jsun-splunk:
Hi @jsharpe, could you help have look? Thanks!

jsharpe (approved):
This seems reasonable to me, thanks

Issue #1200, opened 2024-05-03
pkg-config build broken on clang 15+ and gcc 14+


fuhry:
# Background (what is pkg-config?)

pkg-config 0.29.2 (the latest release as of this writing) bundles an ancient version of glib 2.36 ([released in August 2013](https://download.gnome.org/sources/glib/2.36/)) to workaround the pkg-config/glib circular dependency. It should be noted that pkg-config itself is not well-maintained, with the last major release having been in 2015 and the last point release in 2017, however it remains an important dependency of virtually all build systems as a standardized way to transform dependency names into CFLAGS, LDFLAGS, etc.

Note that [glib](https://docs.gtk.org/glib/) is not to be confused with [GNU libc](https://www.gnu.org/software/libc/), often abbreviated to glibc.

# The issue

glib 2.36 performed implicit integer conversion for its atomic pointer arithmetic functions, using g(s)size (which is aliased to (u)long) instead of (u)intptr. This was a warning in most compilers for many years but was upgraded to an error with [clang 15](https://releases.llvm.org/15.0.0/tools/clang/docs/ReleaseNotes.html#:~:text=The%20%2DWint%2Dconversion%20warning%20diagnostic%20for%20implicit%20int%20%3C%2D%3E%20pointer%20conversions%20now%20defaults%20to%20an%20error%20in%20all%20C%20language%20modes.%20It%20may%20be%20downgraded%20to%20a%20warning%20with%20%2DWno%2Derror%3Dint%2Dconversion%2C%20or%20disabled%20entirely%20with%20%2DWno%2Dint%2Dconversion.) and [gcc 14](https://gcc.gnu.org/gcc-14/porting_to.html#int-conversion). This was [addressed in upstream glib](https://gitlab.gnome.org/GNOME/glib/-/commit/c762d511346d3cb84cea3557a246ccf8873b4a1c) in Dec 2022.

# Patch and next steps

I've [published a patch to the upstream pkg-config mailing list](https://lists.freedesktop.org/archives/pkg-config/2024-May/001122.html). (GitHub unfortunately is refusing to allow me to attach the `.patch` file here, despite claiming that `.patch` is a permitted file extension.) Depending on how quickly the maintainers process it, it may be advisable to bring this patch into rules_foreign_cc in advance of the upstream pkg-config release to unblock builds with newer clang and gcc.

[...]

PR #1211, opened 2024-06-11
Add in custom glib patch

andrewkatson:
@jsharpe So I was able to add in glib but not exactly sure where to point it to. I see in `MODULE.bazel` that we pull in `glib_src` and `glib_dev`. And I followed that path to `extensions.bzl` but I am not quite sure what is going on. Can you give me some hints?


jsharpe:
AFAIK pkg-config bundles a vendored copy of glib -you'll need to add glib as a dependency to the pkg-config build target in the toolchains directory and then update the configure comand to point at that version of glib?
The existing glib_src and glib_dev I think are just in the examples if I remember correctly so those may be red herrings.

WORKSPACE.bazel
```diff
@@ -72,3 +72,10 @@ load("@bazel_skylib_gazelle_plugin//:setup.bzl", "bazel_skylib_gazelle_plugin_se
 bazel_skylib_gazelle_plugin_setup()
 
 gazelle_dependencies(go_repository_default_config = "//:WORKSPACE.bazel")
+
+http_archive(
+    name = "glib",
+    urls = ["https://gitlab.gnome.org/GNOME/glib/-/archive/2.80.3/glib-2.80.3.zip"],
+    strip_prefix = "glib-2.80.3",
+    patches = ["https://lists.freedesktop.org/archives/pkg-config/attachments/20240503/7e621a18/attachment.bin"],
```
jsharpe, line 80: Its probably a good idea to mirror this into the repo as I assume this file isn't that big and relying on a mail archive might not be the most reliable.

andrewkatson:
> AFAIK pkg-config bundles a vendored copy of glib -you'll need to add glib as a dependency to the pkg-config build target in the toolchains directory and then update the configure comand to point at that version of glib? The existing glib_src and glib_dev I think are just in the examples if I remember correctly so those may be red herrings.

Somewhere here? https://github.com/bazelbuild/rules_foreign_cc/blob/37383aaeba8dfbdbf2b0f765c2102c3963d2c8ab/toolchains/built_toolchains.bzl#L271

WORKSPACE.bazel
andrewkatson: Okay Ive got it mirrored locally and patch is applied. Are there any files I should be excluding so I dont upload them?

jsharpe:
See: https://github.com/bazelbuild/rules_foreign_cc/issues/1200#issuecomment-2166538237 - the patch linked here should be sufficient to fix this - @andrewkatson if you could turn that into a PR against this repo then that'd be great!

andrewkatson:
Okay Ill do that

andrewkatson:
@jsharpe I have a question. You said to mirror glib but whenever I try to mirror it I get prompted for their ssh password and then cannot. I've tried a few different ways of mirroring. Is it sufficient to just copy code from their latest release then apply the patch to that?

jsharpe:
I actually don't think we need to pull in an updated glib looking at the previous patch I linked from the other issue - seems that pkg-config can be patched to fix this without needing to bring in a newer glib version. - you should be able to take in the patch that was linked, apply to the rules_foreign_cc and then fixup the Label references to not reference the repo that patch comes from and that should be sufficient.

andrewkatson:
Sorry for all the questions I have never even heard of patching before this! @jsharpe I applied one of the patches. The other one talked about `.bazelrc` which I didnt see in the repo.

andrewkatson:
Also tried to reproduce the errors above using the commands from the remote server and it just hangs trying to build the cmake toolchain. I am still trying to debug but if you can tell what the issue is from the buildkite checks then let me know

andrewkatson:
@jsharpe thank you for that merge! Everything passes now! When can we merge it?

barbu110:
Hey folks, just want to mention this is very important for our project - it's currently blocking builds. Trying to understand if there's a way to get this patch applied from our own project (that is, until this PR gets merged). Any tips?

jsharpe (approved):
LGTM - we're not currently checking in the MODULE.bazel.lock file so I've removed that.

andrewkatson:
> Hey folks, just want to mention this is very important for our project - it's currently blocking builds. Trying to understand if there's a way to get this patch applied from our own project (that is, until this PR gets merged). Any tips?

Follow this https://aur.archlinux.org/packages/envoyproxy from https://github.com/bazelbuild/rules_foreign_cc/issues/1200#issuecomment-2166538237. Does that help?

andrewkatson:
@jsharpe when do you think you'll publish a new version to the bazel central repo?

PR #1212, opened 2024-06-11
chore(deps): cmake/ninja: add new versions

novas0x2a:


toolchains/prebuilt_toolchains.py
```diff
@@ -166,6 +172,7 @@
 }
 
 NINJA_VERSIONS = (
+    "1.12.1",
```
jsharpe, line 175: Is the update to the ninja repo list missing for this update?

jsharpe:
Apologies I broke main by accidently committing directly to the main branch from github.dev - revert is in #1213

toolchains/prebuilt_toolchains.py
novas0x2a: oh, I guess it is; when I run the script by hand and try to commit the result, the repo hooks don't let me; when I let the repo hooks run the script themselves; it looks like they don't stage the ninja file, just the cmake file

novas0x2a: Oh, I see what's happening. The hook requires using buildifier 6.1.0, but my buildifier is 7.1.2, and they produce different output. PR updated.

novas0x2a:
> Apologies I broke main by accidently committing directly to the main branch from github.dev - revert is in #1213

No worries, rebased :)

jsharpe (approved):

PR #1215, opened 2024-06-13
Update framework.bzl to account for externalIncludes

layus:
During our migration to Bazel 7, some of our cmake() rules broke. It turns out there is a new field in CcCompilationContext to account for external includes, as opposed to internal system_includes.

See https://github.com/bazelbuild/bazel/commit/a6ef0b341a8ffe8ab27e5ace79d8eaae158c422b


jsharpe (approved):
LGTM, thanks!

PR #1223, opened 2024-07-08
Add setup_args and run `meson setup` explictly

### LINKED ISSUE #1170: Add a `setup_args` argument to the meson rule
As discussed in #986,  a `setup_args` argument to the meson rule similar to `build_args` and `install_args` for the other steps should be added.

_Originally posted by @sfc-gh-mfickert in https://github.com/bazelbuild/rules_foreign_cc/issues/986#issuecomment-1432936301_

CC @jheaff1

allsey87:
Resolves #1170


allsey87:
The failed test seems to be an issue with the MacOS runners. I have seen this on a couple projects/CI systems lately.

jsharpe:
The buildifier failure is an issue in this PR:
image

The macOS failures do seem to correlate with the switch to use VMs in bazel's CI; I suspect they've either bumped macOS version or are missing some preinstalled tooling that we depend on as main is also failing.

jsharpe:
@jheaff1 do you have time to review the changes in this PR?

allsey87:
> The buildifier failure is an issue in this PR

Fixed in c7bcf7d

jsharpe:
Looks like the other macOS failures are due to the runners upgrading the version of XCode installed.. The fix for this likely needs to come in the apple toolchain.

allsey87:
I think this is ready for review?

jsharpe (approved):
Without really knowing meson, this seems a reasonable change and CI is now all green so I'm happy with this.

PR #1240, opened 2024-07-30
Apply expand_locations_and_make_variables to cache_entries

froody:
When building proj I need to pass the sqlite3 binary via a cache entry, like `{"EXE_SQLITE3": "$(execpath @sqlite3//:shell)"}` but that requires that cache entries are also expanded.


jsharpe (approved):
LGTM

jsharpe:
Apologies for having to revert this @froody - I should have checked CI, but the already broken state meant that I missed the failures this introduced. Feel free to resubmit this as I should soon have CI back to a green state again.

PR #1241, opened 2024-07-31
Fix pkgconfig_build.bzl for Windows build with Bzlmod

meteorcloudy:
Make sure repo names are generated correctly and toolchain label is passed instead of pure string.


jsharpe:
We might want to actually enable bzlmod in CI for windows to test this!: https://github.com/bazelbuild/rules_foreign_cc/blob/fb209674f326b7020f64d5d022cbe7fa16f3209c/.bazelci/config.yaml#L31

jsharpe:
Just updating to latest main to get a clearer idea of what's broken in CI

PR #1246, opened 2024-08-06
Portable copy_dir_contents_to_dir

lamcw:
On BSD-based distro/MacOS, using `cp -L -R {source}/` ignores files or directories with a leading dot. If users have GNU-based coreutils installed on their BSD/Mac system (via Nix for e.g.), `cp -L -R {source}/` would also copy the source directory _underneath_ the target directory, which is not what we want.

This PR changes both the `copy_dir_contents_to_dir` implementation for BSD and MacOS such that it works regardless of whether user has GNU or BSD coreutils installed.


jsharpe (approved):
LGTM, thanks!

Issue #1082, opened 2023-08-18
binfmt_misc confuses configure about cross-compilation


novas0x2a:
This specifically has to do with cross-compilation. autoconf has a cross-compilation check that's normally capable of automatically enabling cross-compilation mode (this is the check I mean): 
```
configure:4434: checking whether we are cross compiling
```

Unfortunately, it's extremely common on linux machines now to have `binfmt_misc` mounted, which permits linux to automatically pass non-native binaries to another process for execution (usually qemu). This can allow static binaries for (e.g.) arm64 platforms to run on (e.g.) amd64 hosts transparently.

Unfortunately, this does not work for dynamic binaries without a bunch of extra setup, because they look for (and can't find) the loader. This is a (redacted) example config.log from when that happens:

```
configure:4434: checking whether we are cross compiling
configure:4442: $CACHEDIR/execroot/$WORKSPACE/external/toolchain-linux-x86_64--linux-arm64/bin/gcc -o conftest --sysroot $CACHEDIR/execroot/$WORKSPACE/external/toolchain-linux-x86_64--linux-arm64_files/aarch64-stage2-linux-gnu/sysroot [TRUNCATED] conftest.c  >&5
configure:4446: $? = 0
configure:4453: ./conftest
aarch64-binfmt-P: Could not open '/lib/ld-linux-aarch64.so.1': No such file or directory
configure:4457: $? = 255
configure:4464: error: in `$CACHEDIR/execroot/$WORKSPACE/bazel-out/k8-fastbuild/bin/external/make/make.build_tmpdir':
configure:4466: error: cannot run C compiled programs.
If you meant to cross compile, use `--host'.
```

Unfortunately, autoconf doesn't handle this case very well, and just breaks. The usual workaround for this is to pass `--host` to `configure` (as the message says) but I'm relatively new to bazel and I'm not sure how to wire this up to automatically detect if the bazel `--platform` (the target platform) is compatible with the bazel `--host_platform` (the execution platform).

I was hoping that someone would see this and know exactly how to add this to `configure_make` so it works automatically :)

matt-sm:
@jsharpe we are interested in submitting a PR to fix this, as it's a blocker to us cross-compiling.

Is this something you would consider merging?

jsharpe:
[...]

PR #1247, opened 2024-08-07
configure_make: support autotools cross-compilation

novas0x2a:
This teaches the autotools-based configure script to pass --host when
cross-compilation is detected. This fixes at least part of #1082. This
functionality is off by default (because configure_make can be used for
non-autotools builds) and can be enabled by passing configure_xcompile
to configure_make.

This functionality is also extended to the built_tools that use autoconf
(pkg-config and make).

It also does a bit of generalization to make it easier to support more
platforms in the future, and adds s390x to the detected arches.

### DIFF
--- a/foreign_cc/built_tools/make_build.bzl
+++ b/foreign_cc/built_tools/make_build.bzl
@@ -15,6 +15,7 @@ load(
     "get_flags_info",
     "get_tools_info",
 )
+load("//foreign_cc/private:detect_xcompile.bzl", "detect_xcompile")
 load("//foreign_cc/private/framework:platform.bzl", "os_name")
 
 def _make_tool_impl(ctx):
@@ -74,6 +75,25 @@ def _make_tool_impl(ctx):
         if os_name(ctx) == "macos":
             non_sysroot_ldflags += ["-undefined", "error"]
 
+        configure_options = [
+            "--without-guile",
+            "--with-guile=no",
+            "--disable-dependency-tracking",
+            "--prefix=$$INSTALLDIR$$",
+        ]
+
+        install_cmd = ["./make install"]
+
+        xcompile_options = detect_xcompile(ctx)
+        if xcompile_options:
+            configure_options.extend(xcompile_options)
+
+            # We can't use make to install make when cross-compiling
+            install_cmd = [
+                "mkdir -p $$INSTALLDIR$$/bin",
+                "cp -p make $$INSTALLDIR$$/bin/make",
+            ]
+
         env.update({
             "AR": absolute_ar,
             "ARFLAGS": _join_flags_list(ctx.workspace_name, arflags),
@@ -85,11 +105,10 @@ def _make_tool_impl(ctx):
 
         configure_env = " ".join(["%s=\"%s\"" % (key, value) for key, value in env.items()])
         script = [
-            "%s ./configure --without-guile --with-guile=no --disable-dependency-tracking --prefix=$$INSTALLDIR$$" % configure_env,
+            "%s ./configure %s" % (configure_env, " ".join(configure_options)),
             "cat build.cfg",
             "./build.sh",
-            "./make install",
-        ]
+        ] + install_cmd
 
     return built_tool_rule_impl(
         ctx,
--- a/foreign_cc/private/cmake_script.bzl
+++ b/foreign_cc/private/cmake_script.bzl
@@ -31,6 +31,9 @@ _TARGET_ARCH_PARAMS = {
     "aarch64": {
         "CMAKE_SYSTEM_PROCESSOR": "aarch64",
     },
+    "s390x": {
+        "CMAKE_SYSTEM_PROCESSOR": "s390x",
+    },
     "x86_64": {
         "CMAKE_SYSTEM_PROCESSOR": "x86_64",
     },
@@ -38,6 +41,7 @@ _TARGET_ARCH_PARAMS = {
 
 def create_cmake_script(
         workspace_name,
+        current_label,
         target_os,
         target_arch,
         host_os,
@@ -59,6 +63,7 @@ def create_cmake_script(
 
     Args:
         workspace_name: current workspace name
+        current_label: The label of the target currently being built
         target_os: The target OS for the build
         target_arch: The target arch for the build
         host_os: The execution OS for the build
@@ -120,7 +125,10 @@ def create_cmake_script(
     # by setting CMAKE_SYSTEM_NAME and CMAKE_SYSTEM_PROCESSOR,
     # see https://github.com/bazelbuild/rules_foreign_cc/issues/289,
     # and https://github.com/bazelbuild/rules_foreign_cc/pull/1062
-    if target_os != host_os and target_os != "unknown":
+    if target_os == "unknown":
+        # buildifier: disable=print
+        print("target_os is unknown, please update foreign_cc/private/framework/platform.bzl and foreign_cc/private/cmake_script.bzl; triggered by", current_label)
+    elif target_os != host_os:
         params.cache.update(_TARGET_OS_PARAMS.get(target_os, {}))
         params.cache.update(_TARGET_ARCH_PARAMS.get(target_arch, {}))
 
--- a/foreign_cc/private/framework/platform.bzl
+++ b/foreign_cc/private/framework/platform.bzl
@@ -1,8 +1,40 @@
 """A helper module containing tools for detecting platform information"""
 
+SUPPORTED_CPU = [
+    "aarch64",
+    "s390x",
+    "x86_64",
+]
+
+SUPPORTED_OS = [
+    "android",
+    "freebsd",
+    "ios",
+    "linux",
+    "macos",
+    "none",
+    "openbsd",
+    "qnx",
+    "tvos",
+    "watchos",
+    "windows",
+]
+
+PLATFORM_CONSTRAINTS_RULE_ATTRIBUTES = {
+    "_{}_constraint".format(i): attr.label(default = Label("@platforms//os:{}".format(i)))
+    for i in SUPPORTED_OS
+}
+
+# this would be cleaner as x | y, but that's not supported in bazel 5.4.0
+PLATFORM_CONSTRAINTS_RULE_ATTRIBUTES.update({
+    "_{}_constraint".format(i): attr.label(default = Label("@platforms//cpu:{}".format(i)))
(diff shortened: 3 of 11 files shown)


novas0x2a:
@jsharpe looks like this probably works now! The only failing builds are the ones that seem to be failing on main, too. (I've tested this in our internal setup on linux, and @matt-sm should be testing on darwin).

foreign_cc/built_tools/make_build.bzl
```diff
@@ -74,6 +75,25 @@ def _make_tool_impl(ctx):
...
+        configure_options = [
+            "--without-guile",
+            "--with-guile=no",
+            "--disable-dependency-tracking",
+            "--prefix=$$INSTALLDIR$$",
+        ]
+
+        install_cmd = ["./make install"]
+
+        xcompile_options = detect_xcompile(ctx)
```
jsharpe, line 87: This all looks correct but when does the built make tool (or pkgconfig?) ever get cross compiled? I would have thought that it was always in the `exec` configuration and so would always be built natively?

foreign_cc/private/cmake_script.bzl
```diff
@@ -31,6 +31,9 @@ _TARGET_ARCH_PARAMS = {
     "aarch64": {
         "CMAKE_SYSTEM_PROCESSOR": "aarch64",
     },
+    "s390x": {
```
jsharpe, line 34: There's probably a lot more values that are missing from here as `CMAKE_SYSTEM_PROCESSOR` can take many values (see https://stackoverflow.com/questions/70475665/what-are-the-possible-values-of-cmake-system-processor). At a minimum PowerPC is missing, although its not that common and is even less comonly used by bazel users IME. I guess we can let people add additional cases as they come across them though..

foreign_cc/private/framework/platform.bzl
```diff
@@ -1,8 +1,40 @@
 """A helper module containing tools for detecting platform information"""
 
+SUPPORTED_CPU = [
+    "aarch64",
+    "s390x",
+    "x86_64",
+]
```
jsharpe, line 7: Ideally we'd mirror whats in platforms: https://github.com/bazelbuild/platforms/blob/main/cpu/BUILD but not sure what a good way of automating the extraction of those values is, especially as these are used to create Labels in the platforms repo..

foreign_cc/private/framework/platform.bzl
```diff
@@ -1,8 +1,40 @@
...
+
+SUPPORTED_OS = [
```
jsharpe, line 9: Same comment as above but for the os values from platforms!

foreign_cc/private/framework/platform.bzl
```diff
@@ -92,10 +152,37 @@ def target_os_name(ctx):
...
+        elif arch == "s390x":
+            return "s390x-ibm-linux-gnu"
+        elif arch == "x86_64":
+            return "aarch64-pc-linux-gnu"
+
+    elif os == "macos":
+        if arch == "aarch64":
+            return "aarch64-apple-macos"
+        elif arch == "x86_64":
+            return "x86_64-apple-macos"
```
jsharpe, line 186: Is macos correct? On my mac at least the triplet names used for gcc installed via homebrew are aarch64-apple-darwin23

jsharpe (commented):
LGTM - will wait on @matt-sm to test on darwin though as I had some questions about whether the macos triplet names were correct..

foreign_cc/built_tools/make_build.bzl
novas0x2a: In rfcc that's true, but due to annoying historical reasons, we build make for both the exec and the target. For autoconf, anything that doesn't do the --host thing when cross-compiling causes it to just break, so I defaulted to adding it everywhere that used autotools (we don't build pkg-config for the target, I was just being consistent to minimize surprise)

foreign_cc/private/framework/platform.bzl
novas0x2a: Yeah, I took these from platforms. Originally, I had every CPU in platforms in this list, and then I realized I wasn't actually sure how to map some of the cpus to gnu-triples or to cmake-system-names; I also didn't have a way to test them. My gut said it was potentially more confusing to get it wrong than to skip it until someone who could test came along and needed it, but I'm happy to restore the original list if you'd like.

One thing I also deliberated doing (even though it's slightly disapproved of in the bazel docs) was issue a `print()` warning every time the unknown case was detected, with a bread crumb back so people know to add to the list. I'd be glad to add that if you want.

foreign_cc/private/framework/platform.bzl
novas0x2a: yeah, see comment there, basically.

foreign_cc/private/framework/platform.bzl
novas0x2a: Yeah, the convention for homebrew is `${arch}-apple-darwin${version}`. Other examples we saw internally (from running config.guess on those platforms) were (aarch64|x86_64)-apple-darwin23.(4|5).0 and I was slightly nervous about lying about the library version, so I checked to see what config.sub would do (since that's what autotools uses to normalize) and it uses `macos` as the default os name, but then also preserves the `darwinXX` when it sees it: 
```
$ ./config.sub-63acb96f92473ceb5e21d873d7c0aee266b3d6d3 aarch64-apple-macos
aarch64-apple-macos
$ ./config.sub-63acb96f92473ceb5e21d873d7c0aee266b3d6d3 aarch64-apple
aarch64-apple-macos
$ ./config.sub-63acb96f92473ceb5e21d873d7c0aee266b3d6d3 aarch64-apple-darwin123
aarch64-apple-darwin123
```

I took this as a sign that macos was fine. A slightly more correct way to guess the host might actually be to ship config.guess and run it as an action to see what it says, and then use that, but since there's no good way to do that reliably for the target tuple, I biased towards internally-consistent normalization.

One _other_ consideration is, I don't currently pass `--build`, just `--host`. This means it's possible that  autoconf's cross_compiling flag will get set to true when it shouldn't be (like in the darwin case above). I did that because it's more likely to result in a working result when the target's not in our tuple, but it does also mean passing `--host aarch64-apple-darwin23` when config.guess thinks it's `aarch64-apple-darwin23.4` triggers cross-compiling mode. The fix for this would be to pass both `--host` and `--build` all the time, so our consistent normalization prevents that from being a problem. (I did confirm this in our CI, our darwin builds do end up emitting `checking whether we are cross compiling... yes`)

I'm happy to adjust this to whichever variation you think is best! After having written this out, I'm actually more inclined to pass --build under every scenario we pass --host and rely on other users updating the data as-needed, but I'll defer to you.

foreign_cc/private/cmake_script.bzl
novas0x2a: I was going to add ppc64le to the list (that's a platform I can actually test on) but ppc64le was added to `@platforms` late, and it's not actually available in the version of `platforms` currently loaded by the tests. That's fixable with a bump of `@platforms`, though, so I'd be glad to do that if you want.

novas0x2a:
> LGTM - will wait on @matt-sm to test on darwin though as I had some questions about whether the macos triplet names were correct..

Mentioned in the review comments above, but I ended up running this in our darwin CI and it worked.

foreign_cc/private/framework/platform.bzl
novas0x2a: I did think of a disadvantage of passing --build: there's no way (that I know of) for bazel to guess which libc the compiler is expecting, because there aren't consistent libc constraints to use. I don't think this actually breaks anything, though, because it appears that when autoconf finds musl it actually internally replaces the triplet with gnu anyway.

I think I'll make the change anyway (it can't break any _existing_ code because autotools would already be breaking in that case) but just something to keep an eye on.

foreign_cc/private/cmake_script.bzl
novas0x2a: Okay, I elected not to do this because it requires that all consuming projects update their platforms version. (I'd be happy to do that if you think it's important, but I realized it's a breaking change)

foreign_cc/private/framework/platform.bzl
novas0x2a: I did add a print

novas0x2a:
I pushed a new version with a few fixes, including fixing two brown-paper-bag fixes (I had two bugs that cancelled each other out, that I've now fixed)

foreign_cc/private/cmake_script.bzl
jsharpe: I've not come across anyone actively adding any fixes for ppc64le so lets assume for now that there is no one using this to avoid the breaking change (although we're still pre 1.0.0 so technically there is no stability guarantee between versions)

jsharpe (approved):
LGTM - I'm sure we'll get fixes to the triple logic if people run into it when cross compiling if we've missed something there...

allsey87:
@novas0x2a could you add a brief example here of how to use `xcompile_options` with `configure_make`?

I might make another PR to add the `wasm32` CPU and `emscripten` OS to the supported platforms.

novas0x2a:
> @novas0x2a could you add a brief example here of how to use `xcompile_options` with `configure_make`?
Looks like the docs generation hasn't run yet, but, the version that landed just has a bool, `configure_xcompile`; defaults to false, if you set it to true it will pass --host and --build (as detected by bazel) to configure. Note that this is causing rfcc to _detect_ the cross-compilation, but only if bazel is already operating in [cross-compile mode](https://bazel.build/extending/platforms#specifying-build-platform).

> I might make another PR to add the `wasm32` CPU and `emscripten` OS to the supported platforms.
It'd need to match the upstream tokens:
- https://github.com/bazelbuild/platforms/blob/main/os/BUILD
- https://github.com/bazelbuild/platforms/blob/main/cpu/BUILD

allsey87:
> > I might make another PR to add the `wasm32` CPU and `emscripten` OS to the supported platforms.
> 
> It'd need to match the upstream tokens:
> 
> * https://github.com/bazelbuild/platforms/blob/main/os/BUILD
> * https://github.com/bazelbuild/platforms/blob/main/cpu/BUILD

They are there:

https://github.com/bazelbuild/platforms/blob/main/os/BUILD#L110
https://github.com/bazelbuild/platforms/blob/main/cpu/BUILD#L168

PR #1250, opened 2024-08-07
Revert "Apply expand_locations_and_make_variables to cache_entries"

jsharpe:
Reverts bazelbuild/rules_foreign_cc#1240


jsharpe:
Unfortunately this PR breaks too many of the examples - it changes the escaping in the values passed in to the cache_entries - see #1249 for some of the changes I was attempting to make to resolve this, but there's just too many to fix at the moment and I need to get CI back to a green state.

PR #1253, opened 2024-08-08
runnable_binary: use package_relative_label

lamcw:
With `--enable_bzlmod`, `native.repository_name` returns the apparent repo name (e.g. `@foo~`) instead of the canonical repo name. This can be a problem in the `runnable_binary` macro (esp. genrule targets) as `rlocationpath` may not be able to "see" `srcs` that are already listed, when some other modules depend on the `runnable_binary` target.

```
label '@@[unknown repo 'foo~' requested from @@foo~]//:foo_exe_fg' in $(location) expression is not a declared prerequisite of this rule. Since this rule was created by the macro 'runnable_binary', the error might have been caused by the macro implementation
```

Instead, `package_relative_label` should be used in the `runnable_binary` macro such that Bazel correctly converts the `filegroup` label in the context of the package currently being initialized.


lamcw:
@jsharpe is there any plan to drop support for 5.4.0? `package_relative_label` is not supported until 6.1.0 https://github.com/bazelbuild/bazel/pull/17435

jsharpe:
As per the LTS schedule - https://bazel.build/release we'll continue to support 5.4.1 until Jan 2025.
Are you able to use bazel-contrib/bazel_features to detect support for package_relative_label and conditionally use it?

lamcw:
> Are you able to use bazel-contrib/bazel_features to detect support for package_relative_label and conditionally use it?

`bazel_features` does not support detecting `package_relative_label` -- instead I am using a `hasattr` guard to make sure it exists before invoking on it. Please take a look 🙏

jsharpe (approved):
LGTM!

PR #1254, opened 2024-08-08
Enable bzlmod on windows

jsharpe:


jsharpe:
Just seeing what breaks!

Issue #1244, opened 2024-08-01
[Bazel CI] CcMesonMakeRule Fails with Linker Error, Build Fails



sgowroji:
CI: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/4026#01910c40-d7fb-4158-b4cb-8bf56fa9423d

 Platform: MacOS

 Logs: 
```
ERROR: [0m/private/var/tmp/_bazel_buildkite/4dbdaa322633a6e7887d5b601b0cf2f5/external/libffi/BUILD.bazel:10:6: Foreign Cc - Meson: Building libffi failed: (Exit 1): bash failed: error executing CcMesonMakeRule command (from target @@libffi//:libffi) 
ERROR: Unable to detect linker for compiler `/private/var/tmp/_bazel_buildkite/4dbdaa322633a6e7887d5b601b0cf2f5/execroot/rules_foreign_cc_examples/external/local_config_apple_cc/wrapped_clang -Wl,--version -lc++ -Wl,-oso_prefix,__BAZEL_EXECUTION_ROOT__/ -headerpad_max_install_names -no-canonical-prefixes -target x86_64-apple-macosx14.2 -fobjc-link-runtime -Xlinker -no_deduplicate -framework Foundation -Wl,-no_warn_duplicate_libraries`
error: linker command failed with exit code 1 (use -v to see invocation)
ERROR: [0m/private/var/tmp/_bazel_buildkite/4dbdaa322633a6e7887d5b601b0cf2f5/external/rules_foreign_cc_examples_third_party/glib/BUILD:10:11 Executing genrule @@rules_foreign_cc_examples_third_party//glib:glib_build_test_0__deps failed: (Exit 1): bash failed: error executing CcMesonMakeRule command (from target @@libffi//:libffi) 
ERROR: [0mBuild did NOT complete successfully
Error: The command exited with status 1[0m
error: exit status 1
``` 
 Culprit: 

 Steps:
  ``` 
 git clone https://github.com/bazelbuild/rules_foreign_cc
git reset 9d5727d5e51bf1be8423dff7996bdb4d847b47e6  --hard
export USE_BAZEL_VERSION=a7264d99e2d14463cb683a4091d5ad0e08431205
bazel build -- //... //:third_party_examples_macos_tests -//cmake_android/... -@rules_foreign_cc_examples_third_party//curl:curl_test -@rules_foreign_cc_examples_third_party//openssl:openssl_test 
 ``` 
[...]

PR #1260, opened 2024-08-14
fix meson tests for macOS CI

jsun-splunk:
This is an attempt to fix the macOS CI issues mentioned in https://github.com/bazelbuild/rules_foreign_cc/issues/1244.

Also noting https://github.com/bazelbuild/rules_foreign_cc/pull/1251 disabled the broken tests. Whereas this MR should fix the tests.

The original problem should be a meson issue and not a foreign_cc one. see https://github.com/mesonbuild/meson/pull/12574. Although, I couldn't track down what introduced the problem.

I got the idea that it is a meson problem, given `-Wl,--version` is not generated as a part of our `cxx_linker_executable` or `cxx_linker_shared`.

The quick solution is to update our default meson version to a version that includes the fix.

Had to also update glib, as after the meson fix it was running to this an issue similar to https://gitlab.gnome.org/GNOME/glib/-/issues/2995.

### DIFF
--- a/examples/third_party/BUILD.bazel
+++ b/examples/third_party/BUILD.bazel
@@ -86,8 +86,7 @@ test_suite(
         "//cares:test_c_ares",
         "//cares:test_c_ares_ios",
         "//curl:curl_test_suite",
-        # Fails due to linker error in ffi (https://github.com/mesonbuild/meson/issues/12282)
-        # "//glib:glib_build_test",
+        "//glib:glib_build_test",
         "//gn:gn_launch_test",
         "//gperftools:test",
         "//iconv:iconv_macos_build_test",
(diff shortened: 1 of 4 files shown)


jsharpe (approved):
Awesome, thanks for working out this workaround!

bbatliner-ocient:
The first URL tried doesn't seem to exist: https://mirror.bazel.build/github.com/mesonbuild/meson/releases/download/1.5.1/meson-1.5.1.tar.gz

I recognize this was copying the pattern from the other Meson versions.

What is the process for getting this file to exist on `mirror.bazel.build`? The lack of this file's existence produces warnings of the form:
```
WARNING: Download from https://mirror.bazel.build/github.com/mesonbuild/meson/releases/download/1.5.1/meson-1.5.1.tar.gz failed: class java.io.FileNotFoundException GET returned 404 Not Found
```

PR #1261, opened 2024-08-14
Merge user provided CPPFLAGS with build system CPPFLAGS

allsey87:
Another PR regarding flags being set in `env`. Currently there is no solution to set `CPPFLAGS` via `configure_make`. If you set them in the environment, they will be clobbered by the build system flags. If you set them via `configure_options` or `configure_prefix` (I saw someone doing that somewhere), then they clobber the build system `CPPFLAGS` meaning you cannot use dependencies (since this is how header files from dependencies are made during the build).

### DIFF
--- a/foreign_cc/private/make_env_vars.bzl
+++ b/foreign_cc/private/make_env_vars.bzl
@@ -85,9 +85,9 @@ _MAKE_FLAGS = {
     "AR_FLAGS": "cxx_linker_static",
     "ASFLAGS": "assemble",
     "CFLAGS": "cc",
+    "CPPFLAGS": "",
     "CXXFLAGS": "cxx",
     "LDFLAGS": "cxx_linker_executable",
-    "CPPFLAGS": "",
     # missing: cxx_linker_shared
 }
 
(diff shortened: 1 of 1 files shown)


foreign_cc/private/make_env_vars.bzl
```diff
@@ -82,6 +85,7 @@ _MAKE_FLAGS = {
     "AR_FLAGS": "cxx_linker_static",
     "ASFLAGS": "assemble",
     "CFLAGS": "cc",
+    "CPPFLAGS": "",
```
allsey87, line 88: What do you think about this @jsharpe? It is a bit of a hack that exploits that `getattr(flags, _MAKE_FLAGS[flag], [])` below will default to an empty list since an attribute cannot have an empty key.

allsey87: I have changed this in 799f91ca74aa722ae3f66362726b1eceaf4cf8a6, I think it is better to be more explicit here.

jsharpe (approved):

jsharpe:
@allsey87 if you update this branch then it will auto merge once CI has run - I don't have permissions on your fork to update it myself.

PR #1266, opened 2024-08-20
fix: keep original source tree when symlink into ext_include_dir

gulu-goolu:


foreign_cc/private/framework.bzl
```diff
@@ -767,13 +767,23 @@ def _symlink_contents_to_dir(dir_name, files_list):
...
-$$EXT_BUILD_ROOT$$/{} $$EXT_BUILD_DEPS$$/{} True".format(path, dir_name))
+$$EXT_BUILD_ROOT$$/{} $$EXT_BUILD_DEPS$$/{} True".format(path, _get_dir_name(dir_name, file)))
 
     return lines
 
 def _file_path(file):
     return file if type(file) == "string" else file.path
 
+def _get_dir_name(dir, file):
+    if dir == "include":
```
jsharpe, line 778: Don't we always want to do this? Surely if it was linking say share/ into the tree you'd also want the folder structure to be maintained?

PR #1269, opened 2024-08-23
Remove dependencies on rules_cc

jsharpe:
`glib_dev` was incorrectly loading a symbol from `rules_cc` which causes visibility issues on windows bzlmod.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -15,7 +15,6 @@ bazel_dep(name = "rules_python", version = "0.23.1")
 bazel_dep(name = "gazelle", version = "0.34.0", dev_dependency = True, repo_name = "bazel_gazelle")
 bazel_dep(name = "bazel_skylib_gazelle_plugin", version = "1.5.0", dev_dependency = True)
 bazel_dep(name = "stardoc", version = "0.5.6", dev_dependency = True, repo_name = "io_bazel_stardoc")
-bazel_dep(name = "rules_cc", version = "0.0.9", dev_dependency = True)
 
 python = use_extension("@rules_python//python/extensions:python.bzl", "python")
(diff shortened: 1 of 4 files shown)


jsharpe:
@meteorcloudy - it seems we need to do a bit of work on the bazel_ci_rules bzlmod package as the generated repo has a dev_dependency on rules_cc that isn't declared; I guess the WORKSPACE macro should also be migrated to a extension for proper bzlmod support?

meteorcloudy:
Yes, this is how Bazel uses it:
https://github.com/bazelbuild/bazel/blob/master/MODULE.bazel#L354-L364
https://github.com/bazelbuild/bazel/blob/master/rbe_extension.bzl

jsharpe:
> Yes, this is how Bazel uses it: [bazelbuild/bazel@`master`/MODULE.bazel#L354-L364](https://github.com/bazelbuild/bazel/blob/master/MODULE.bazel?rgh-link-date=2024-08-26T07%3A42%3A25Z#L354-L364) [bazelbuild/bazel@`master`/rbe_extension.bzl](https://github.com/bazelbuild/bazel/blob/master/rbe_extension.bzl?rgh-link-date=2024-08-26T07%3A42%3A25Z)

This only works though because bazel has a dep on rules_cc

meteorcloudy:
Oh, I see, but isn't rules_cc a nature dependency of rules_foriegn_cc?

jsharpe:
> Oh, I see, but isn't rules_cc a nature dependency of rules_foriegn_cc?

Well it does use `cc_library` and `cc_binary` but obviously these are available in bazel natively at the moment.. The rules_cc toolchain config is often out of sync with what's in bazel and so I've tended not to take deps on rules_cc.

Issue #1267, opened 2024-08-22
runnable_binary generated a zero-length wrapper, but succeeded


novas0x2a:
Sorry for how vague this is, still trying to figure out what might have happened. What we observed is:
- The shell script that gets generated (i.e. `${name}_wrapper.sh`) ended up as a zero-length file, but the action succeeded. 
- This meant that the empty script was uploaded to our remote cache, and broke other builds until we figured out what happened and cleaned it up.
- We've only observed this ~once~ twice now.
- This happened on Windows.

I'm still trying to figure out how it might have happened, but I'm guessing it's somehow Windows-related. I was curious if anyone else had seen this happen. Looking at the [code](https://github.com/bazelbuild/rules_foreign_cc/blob/d9367d28ad5292dd2ed42ae85223e79e845d0f3f/foreign_cc/utils.bzl#L45-L47), it seems like the only way this could have happened is if the sed failed and the script didn't notice (the `>$@` would have created the output regardless of the sed). Maybe `errexit` should have caught it? I don't understand currently, since genrules are supposed to run with errexit. I don't have a reliable repro or anything, unfortunately, though I'm trying.

novas0x2a:
I believe my above-referenced commits have fixed the issue; we haven't seen a reoccurrence since then.

PR #1270, opened 2024-08-24
runnable_binary: avoid using > on the output file

novas0x2a:
In #1267 there was an issue where, somehow, the wrapper script ended up as a zero-length file but the action succeeded (resulting in storing a zero-length file instead of the script in the caches). genrules are documented as running inside `-ueo pipefail` so it seems like this should be impossible...

In any case, this avoids using >$@ because the > runs before the command itself does, creating a zero-length file before populating it, and is the only way I could think of that this ends up as an empty file. Hopefully, if the problem repeats, genrule will notice that the file didn't get created and fail.

SH_BINARY_FILENAME also wasn't used, so this skips a step.

### DIFF
--- a/foreign_cc/utils.bzl
+++ b/foreign_cc/utils.bzl
@@ -44,7 +44,7 @@ def runnable_binary(name, binary, foreign_cc_target, match_binary_name = False, 
 
     wrapper_cmd = """
     sed s@EXECUTABLE@$(rlocationpath {name})@g $(location @rules_foreign_cc//foreign_cc/private:runnable_binary_wrapper.sh) > tmp
-    sed s@SH_BINARY_FILENAME@{sh_binary_filename}@g tmp > $@
+    cp tmp $@
     """
 
     if hasattr(native, "package_relative_label"):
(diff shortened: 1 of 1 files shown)


foreign_cc/utils.bzl
```diff
@@ -44,7 +44,7 @@ def runnable_binary(name, binary, foreign_cc_target, match_binary_name = False,
 
     wrapper_cmd = """
     sed s@EXECUTABLE@$(rlocationpath {name})@g $(location @rules_foreign_cc//foreign_cc/private:runnable_binary_wrapper.sh) > tmp
-    sed s@SH_BINARY_FILENAME@{sh_binary_filename}@g tmp > $@
```
aij, line 47: note[nonblocking]: I was wondering about this, so I checked on the history.  `SH_BINARY_FILENAME` was added in 41c937a7a94e0c20bd19d02983d0e8c4f760fcde and the last (and only) use was removed in 4831827e291e3b4a1f73679a5c1197a26d0ac3ed, so it seems like a pretty safe clean up.

aij (approved):

jsharpe (approved):

PR #1272, opened 2024-08-30
Support location expansion and make variable substitution in CMake cache entries

### LINKED ISSUE #1013: cmake not support $(execpath //foo:bar)?
I defined a  cache_entries: "MYSQL_INCLUDE_DIR": "$(locations @mysqlclient)" + "/include",  but $(locations @mysqlclient) is empty, so it's cmake not support  [make variables](https://bazel.build/reference/be/make-variables)? or just cannot used in cache_entries or way I used not correct?

allsey87:
Another attempt at #1124, #1240, #1249, and #1250

Closes #1013

Let's see if I can fix the examples and make CI green again!


allsey87:
@jsharpe Buildkite is giving us those sweet, sweet green lights. What do you think?

jsharpe (approved):
+1 - can you resolve the conflicts then we can get this merged.

PR #1278, opened 2024-09-06
fix undefined variables in the wrap_outputs script

novas0x2a:
BUILD_TMPDIR and EXT_BUILD_DEPS were used in the wrapper script, but they weren't actually defined. This wasn't caught because the script prelude (which sets bash -u) wasn't used either.

This fix:
- Adds the missing script prelude to the wrapper script
- Generalizes get_env_prelude slightly so built_tools can use it directly, without needing to override variables
- Passes the env prelude to wrapped_outputs so it can set those same variables
- Prints out the paths to the build and deps dir at the end of a failed script, to make it easier to debug failed builds.


novas0x2a:
ping @jsharpe :)
(Also, I'll be going to bazelcon if you want to meet in person)

jsharpe (approved):
Nice catch!

PR #1285, opened 2024-09-11
correct overeager slash conversion on windows

novas0x2a:
Bazel's `File().path` method will return forward-slashed paths; this means if you pass a flag like `"/FI" + f.path` it'll end up looking like `/FIpath/to/thing`; however, `_convert_flags` was converting all `/` to `-`, which ended up breaking the path. This makes sure to only replace the _first_ `/`.


novas0x2a:
ping @jsharpe :)

Issue #1271, opened 2024-08-29
Sandbox makes build non-deterministic when building with debug symbols


aaron-michaux:
The sandbox has a unique incrementing number in its path, which in turn finds its way into the debug symbols. This means that binary targets are non-deterministic. For example, when I build `@zlib//:zlib` with debug symbols, the bit-pattern of `libz.a` depends on sandbox path, which changes from build to build. Thus, anything that links to `libz.a` is also non-deterministic.

Gcc has an option `-fdebug-prefix-map=old=new`, which can mitigate against this; however, to set this option, you'd have to _a priori_ know the path to the sandbox. That is, we want to have a flag that looks like: `-fdebug-prefix-map=$SANDBOX_PATH=.`, where the bazel rule fills in `$SANDBOX_PATH` for you.

In my mind, this is an issue with the bazel sandbox itself; however, the issue can be mitigated in `rules_foreign_cc`.

Thanks to @fmeum for prompting me to create this ticket.

fmeum:
> In my mind, this is an issue with the bazel sandbox itself; however, the issue can be mitigated in rules_foreign_cc.

It's not an issue with the sandbox itself if you follow its golden rule: exclusively use relative paths. If you do that, then compilers would generally emit relative paths into debug output. But rules_foreign_cc may not be able to ensure this as it forks out to external build systems.

PR #1288, opened 2024-09-12
Fix for #1271, set_file_prefix_map attribute

aaron-michaux:
One way to fix https://github.com/bazelbuild/rules_foreign_cc/issues/1271 is to pass `-ffile-prefix-map=$SANDBOX=.` This fix creates a `set_file_prefix_map` attribute for the `cmake` rule. If `True`, will set the gcc/clang flag to eviscerate the sandbox path from the debug symbols.

### DIFF
--- a/foreign_cc/cmake.bzl
+++ b/foreign_cc/cmake.bzl
@@ -408,11 +408,11 @@ def _attrs():
         ),
         "set_file_prefix_map": attr.bool(
             doc = (
-                "Use -ffile-prefix-map with the intention to remove the sandbox path from " +\
+                "Use -ffile-prefix-map with the intention to remove the sandbox path from " +
                 "debug symbols"
             ),
             mandatory = False,
-            default = False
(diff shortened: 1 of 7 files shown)


foreign_cc/cmake.bzl
```diff
@@ -393,6 +393,14 @@ def _attrs():
             doc = "Arguments for the CMake install command",
             mandatory = False,
         ),
+        "set_file_prefix_map": attr.bool(
```
jsharpe, line 396: Couldn't we just add this to `CC_EXTERNAL_RULE_ATTRIBUTES` rather than having to modify each of the individual rulesets?

aaron-michaux: I've just checked this works. Obviously better to add to `CC_EXTERNAL_RULE_ATTRIBUTES`

jsharpe (approved):

PR #1289, opened 2024-09-14
Update the broken URL on BCR

asymingt:
I noticed that the homepage URL for `rules_foreign_cc` is broken on BCR, and I *think* this PR is how to go about fixing this. See: https://registry.bazel.build/modules/rules_foreign_cc


jsharpe:
Yes that url was wrong but we're going to be migrating this repo out of bazelbuild very soon: https://github.com/bazelbuild/rules_foreign_cc/issues/1290 so we can probably skip merging this as we'll just end up having to revert this as soon as the migration happens!

asymingt:
There was a wider context that I didn't know. Thanks for letting me know. I'll close this PR.

PR #1295, opened 2024-10-01
Add ffmpeg example

mering:
This does not build with the configured hermetic toolchain.

# Instructions

## (Optional) Run in container
Create container:
```sh
docker run -it --rm -v $(pwd):/work -w /work debian:12
```
Setup container:
```sh
apt update && apt install -y curl
curl -LSs "https://github.com/bazelbuild/bazelisk/releases/download/v1.20.0/bazelisk-linux-amd64" -o "/usr/bin/bazelisk" && chmod +x "/usr/bin/bazelisk" && ln -s "/usr/bin/bazelisk" "/usr/bin/bazel"
useradd --shell /bin/bash --create-home builder
su builder
```

## Build
```sh
cd examples/third_party
USE_BAZEL_VERSION=7.x bazelisk build @ffmpeg
```

### DIFF
--- a/examples/third_party/.bazelrc
+++ b/examples/third_party/.bazelrc
@@ -1,0 +1,13 @@
+# Enable Bzlmod for every Bazel command
+common --enable_bzlmod
+
+# Always use the pre-configured toolchain.
+build --repo_env=BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1
+build --action_env=BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1
+
+# Use a static value for `PATH` and does not inherit `LD_LIBRARY_PATH`. Doesn't let environment
+# variables like `PATH` sneak into the build, which can cause massive cache misses when they change.
+# Use `--action_env=ENV_VARIABLE` if you want to inherit specific environment variables from the
+# client, but note that doing so can prevent cross-user caching if a shared cache is used.
+# Docs: https://bazel.build/reference/command-line-reference#flag--incompatible_strict_action_env
+build --incompatible_strict_action_env
--- a/examples/third_party/BUILD.sysroot.bazel
+++ b/examples/third_party/BUILD.sysroot.bazel
@@ -1,0 +1,46 @@
+filegroup(
+    name = "all_files",
+    srcs = glob([
+        "lib/x86_64-linux-gnu/ld*",
+        "lib/x86_64-linux-gnu/libc*",
+        "lib/x86_64-linux-gnu/libdl*",
+        "lib/x86_64-linux-gnu/libgcc*",
+        "lib/x86_64-linux-gnu/libm*",
+        "lib/x86_64-linux-gnu/libpthread*",
+        "lib/x86_64-linux-gnu/librt*",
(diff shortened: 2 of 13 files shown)


voxeljorge:
I've also run into this, while trying to build ffmpeg with `rules_foreign_cc`

I'm able to get a successful build by letting bazel just select the local c compiler but ideally I would like to use a hermetic llvm + sysroot.

voxeljorge:
my build fails with errors like this:

```
gcc: error: unrecognized command line option '--target=x86_64-unknown-linux-gnu'
gcc: error: unrecognized command line option '-fcolor-diagnostics'
gcc: error: unrecognized command line option '-Wthread-safety'
gcc: error: unrecognized command line option '-Wself-assign'
```

in `config.log`

voxeljorge:
Seems like this bug is probably related: https://trac.ffmpeg.org/ticket/9310

voxeljorge:
Adding `"--cc=$$EXT_BUILD_ROOT/$(CC)"` to `configure_options` fixed my build. I suspect that might work for this PR as well.

voxeljorge:
I found that I had similar errors because not only is configure not using `CC`, it is also not using `CFLAGS` or any of the other env vars. I ended up writing a wrapper `configure_wrapper` which would read the env vars and set `--cc`, `--extral-cflags` etc

mering:
> I found that I had similar errors because not only is configure not using `CC`, it is also not using `CFLAGS` or any of the other env vars. I ended up writing a wrapper `configure_wrapper` which would read the env vars and set `--cc`, `--extral-cflags` etc

@voxeljorge Could you share your wrapper?

I noticed that my build was missing the `--sysroot` as `CFLAGS`. I updated my branch accordingly but the hardcoded paths are ugly. Is this available in some makevar or similar?

`libvpx` and `libdav1d` are still not building in my branch. Do these codec libraries build for you?

mering:
I created two fixes for `libdav1d` with `meson`: #1302 #1303

voxeljorge:
I don't build libvpx or libdav1d at all. The configure wrapper is trivial:

```bash
#!/bin/bash

# ffmpeg ignores flags like CC/CXX and CFLAGS: https://trac.ffmpeg.org/ticket/9310
# so we have a wrapper which assigns them correctly
exec "$(dirname "$0")"/configure \
	--cc="$CC" \
	--cxx="$CXX" \
	--extra-cflags="$CFLAGS" \
	--extra-cxxflags="$CXXFLAGS" \
	--extra-ldflags="$LDFLAGS" \
	"$@"
```

The wrapper can be installed a couple of ways, you could inject it during the repository creation time what I did was copy the sources with something like this:

```bazel
load("@aspect_bazel_lib//lib:copy_to_directory.bzl", "copy_to_directory")

copy_to_directory(
    name = "src",
    srcs = [
        "configure_wrapper",
        "@ffmpeg_7_0_sources",
    ],
    include_external_repositories = ["ffmpeg_7_0_sources"],
)

configure_make(
    name = "ffmpeg",
    lib_source = "//third_party/ffmpeg:src",
    configure_command = "src/configure_wrapper",
)
```

I do have ffmpeg building with a limited set of codecs currently, even with a modern llvm and sysroot.

PR #1302, opened 2024-10-16
Configure AR and STRIP tools for meson

mering:
This is required to fix `libdav1d` in #1295.


jsharpe:
This doesn't work on macOS from the look of things - commonly libraries don't build with AR=libtool on macos so that is probably the issue. Not sure what the correct fix is though

PR #1303, opened 2024-10-16
Improve flags handling for meson

mering:
This aligns this to what other tools are doing. The `_join_flags_list()` pattern is also used for other foreign build tools.

This is required to fix `--sysroot` flag even without cross compiling for example when using in combination with a hermetic toolchain (for example `toolchains_llvm`). This flag is only available in `get_flags_info()` and for the complicated ffmpeg build this works flawlessly. Without it being absolute, it doesn't work.

This is required to fix `libdav1d` in #1295.

### DIFF
--- a/examples/third_party/glib/BUILD.glib.bazel
+++ b/examples/third_party/glib/BUILD.glib.bazel
@@ -22,6 +22,13 @@ DEPS = [
 
 meson(
     name = "glib",
+    copts = select({
+        ":msvc_compiler": [
+            # This string is used as enum in gio/glib-compile-resources.c:718
+            "-UCOMPILER_MSVC",
+        ],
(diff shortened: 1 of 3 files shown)


voxeljorge:
I have been able to reproduce the error in a similar build and can confirm that this patch fixes my error, which also was due to the `--sysroot` flag being a relative path instead of an absolute path before being passed to meson.

mering:
@voxeljorge @jsharpe could you help me fix the Windows examples build? I don't have this platform available and don't know much about building on it.

mering:
I could fix the CI with the help of @feuerste and it is all green now!

@jsharpe so this is ready for review and merge.

jsharpe (approved):

PR #1314, opened 2024-11-05
Remove deprecated use of incompatible_use_toolchain_transition

EdSchouten:
This is only necessary to support versions of Bazel prior to 5.0.0. According to the README, we only need to support Bazel 5.4.0 and later.


EdSchouten:
CI failure seems to be unrelated to the change at hand?

jsharpe (approved):
Thanks

PR #1316, opened 2024-11-06
fix: docs pages

jjmaestro:
### fix: deprecated versions of actions/upload-artifact

The [pages workflow is failing](https://github.com/bazel-contrib/rules_foreign_cc/actions/runs/11706830028/job/32604817856) due to deprecated versions of actions/upload-artifact.

I believe this is why the docs are not updating (e.g. the Meson docs are still blank after #1060 landed).

For more info, see:
 * https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/
 * https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/
 * https://github.com/actions/upload-artifact
 
### chore: bump version.bzl to the current released version
This is also needed for docs generation, see `docs/tools/workspace_status.sh`

### docs: update releases

### DIFF
--- a/.github/docs-0.1.0.patch
+++ b/.github/docs-0.1.0.patch
@@ -40,3 +40,16 @@ index 0000000..98bb722
 +"""A module represeting the version of rules_foreign_cc"""
 +
 +VERSION = "0.1.0"
+diff --git a/docs/.bazelrc b/docs/.bazelrc
+new file mode 100644
+index 0000000..9eb0697
+--- /dev/null
++++ b/docs/.bazelrc
+@@ -0,0 +1,7 @@
++# Bazel configuration flags for rules_foreign_cc_docs
++
++# https://github.com/bazelbuild/stardoc/issues/112
++common --incompatible_allow_tags_propagation
++
++common --noenable_bzlmod
++build --workspace_status_command=tools/workspace_status.sh
--- a/.github/docs-0.10.1.patch
+++ b/.github/docs-0.10.1.patch
@@ -1,16 +1,3 @@
-diff --git a/docs/BUILD.bazel b/docs/BUILD.bazel
-index 2bdbf6e..418dd13 100644
---- a/docs/BUILD.bazel
-+++ b/docs/BUILD.bazel
-@@ -97,7 +97,7 @@ set -euo pipefail
- short_commit="$${SHORT_COMMIT}"
- commit="$${COMMIT}"
- release="$${RELEASE}"
--if [[ -n "\\$${RELEASE}" ]]; then
-+if [[ -n "\\$${RELEASE:-}" ]]; then
-     release="\\$${RELEASE}"
- fi
- mkdir -p \\$${BUILD_WORKSPACE_DIRECTORY}/src
 diff --git a/docs/.bazelrc b/docs/.bazelrc
 index dc6a3f7..9eb0697 100644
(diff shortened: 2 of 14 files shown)


jjmaestro:
I was hoping this would work... but there's a bunch of checks failing for the generation of the docs.

I tried to test the workflow in my fork but I wasn't able to :S  @jsharpe if you know how to do it, I'll be happy to try.

jsharpe:
I don't actually know how the doc generation works - @UebelAndre put it together. I believe the issue though is that its trying to run the latest version of stardoc on older branches to generate the docs for older releases and so its failing because bzlmod syntax such as `@@` didn't exist when the earlier releases were done.

jjmaestro:
Yeah, I was checking the checks and it does seem like it was doing that, e.g. running Bazel 7.4 with Bzlmod on the `0.7.0` branch and failing... it's a pity cos it stops the other checks from running with more recent tags, e.g. `0.12.0` :-/

For the time being, I'm removing the old releases to see if the checks pass for the new stuff, and if so, I'll revert the change and we can at least know that it works for the new stuff 😅

jjmaestro:
OK, so this worked for the latest releases 🎉  Now, there's some other stuff failing...

jjmaestro:
Aaaah, I found this! https://github.com/bazel-contrib/rules_foreign_cc/blob/9edaefe13cc8aff6ba3bdc5736dd7d2adff5af44/.github/workflows/pages.yaml#L42-L47

So, I guess the correct approach is to produce patches for all the old branches! A bit cumbersome but yeah, it's a good escape hatch @UebelAndre 😄 

I'm going to try to patch `docs/.bazelrc` for those branches forcing to run without Bzlmod...

jjmaestro:
OK, I think this last commit will fix the issue, but the checks fail because the `curl` is trying to get the patches from `main` and the patches don't exist for a bunch of releases that I'm now expecting to be patched :-/

The only way out I see is to create a separate PR with the patches and land it first, then rebase this PR. But that would also mean landing the PR with the patches "blind", without much testing so... maybe we could land this PR and see if it fixes the issues.

@jsharpe let me know what you want to do!

jjmaestro:
I've tried reproducing the checksum error with [another script](https://gist.github.com/jjmaestro/98fcd9d83374596e73bef39aa1137a83) but I can't, the checksum of the downloaded artifact always matches the expected checksum... so I'm not sure what's happening.

@jsharpe since the doc generation is currently broken, I don't think that landing this PR would create further issues, so... maybe we can try and see if the CI jobs finally go green? :-?

jjmaestro:
@jsharpe ping here too, I think this PR should land, the checks breaking seem completely unrelated (the weird GNOME `glib` checksum error that I haven't been able to reproduce).

jsharpe (approved):
Thank you for looking into why this was failing - I believe what you've done should work - will merge it and check!

jsharpe:
This is almost there - but think some of the patches don't quite apply e.g. 0.3.0 - github cancels all the actions as soon as one fails so its a bit annoying to check whether any of the others are also bad..

jjmaestro:
Ok, I'll give it another thought and will also try to modify the workflow to try and run all of the jobs so at least there's more info to debug!

PR #1323, opened 2024-11-18
chore(docs): Remove very old versions of the docs.

jsharpe:
The patches for these branches no longer apply and these are very old versions of the rules which won't work with a supported version of bazel so remove them.


jjmaestro:
I think I actually fixed these in #1327 😅  I think if you stamp that PR, it'll fix all the issues, plus the changes to the GH workflow I added allow for easy filtering and deprecation of the versions / patches :)

PR #1325, opened 2024-11-19
chore: more precommit hooks

jjmaestro:
> [!NOTE]  
>  Stacked on top of #1327

Here's a bit of a winter house cleaning :)

Add a bunch of pre-commit hooks:
* Some basic hooks to prevent usual whitespace nits (mixed line endings, trailing whitespace and missing EOF) and merge conflicts
* Lint YAML (actions) and validate GH Actions (two validators because they are both quite useful for different reasons!)
* Also validate shell scripts and Makefiles (NOTE: one of the GH Action validators also shellchecks the scripts in them :)
* Finally, skip checks on generated files (diff/patches and testing expected outputs).

Then, run `pre-commit run --all-files` to lint and fix the things that can be automatically fixed.

Finally, a few more commits to fix the stuff that can't be fixed manually:

* some outdated GH Actions + shellcheck parts of the GH actions
* check Makefiles
* shellcheck scripts and exclude `examples/` because there's quite a bunch of warnings so it's probably better to leave those for another PR.


jjmaestro:
Ah, forgot to double-check tests locally :S  The `.pre-commit` hooks modified some expected test output. I'll exclude those files as well.

jjmaestro:
@jsharpe hopefully you'll like this one, it makes it easier to contribute stuff :) This one should be all green once #1327 lands!

jsharpe (approved):

jsharpe:
Sigh, more download issues in CI.. Seems the python.org download site has some bad permissions at the moment.

PR #1339, opened 2024-12-11
add cmake configuration attr

matt-sm:
`cmake()` appends a `--config` arg to every `cmake --install` and `cmake --build`.

The default value of this arg is derived from the COMPILATION_MODE env var.

This PR allows `cmake()` to override this value.


matt-sm:
@jsharpe can this be merged?

jsharpe (approved):

PR #1343, opened 2024-12-15
fix: `pyc` files in Meson `mesonbuild` repo causing rebuilds

### LINKED ISSUE #1342: `pyc` files in Meson `mesonbuild` repo causing rebuilds
When I was using the `meson` rule, I saw that on the second consecutive `bazel build //example` call, Bazel seemed to be rebuilding / doing something when I thought it should just straight return since nothing should have changed.

I tried to see what was going on using `--explain=` and `--execution_log_json_file=` flags and found out it was due to `pyc` appearing in Meson's `mesonbuild` repository. 

Repro: https://github.com/jjmaestro/rfcc-meson-rebuilding

TL;DR I still don't know how this is happening, I tried updating `rules_python` to the latest version and using some `py_binary` flags that I thought would ensure the `pyc` files would appear "before" (`precompile`d) but nothing seemed to help.

I had also tried excluding the `__pycache__` folders and didn't see any effect... until @rickeylev pointed me to a better example in `rules_python`!

https://github.com/bazelbuild/rules_python/blob/727ab43107fb0b2d528140f609b873670a5c6c26/python/private/python_repository.bzl#L187-L195

The "fix" seems to be as simple as adding `exclude = ["**/__pycache__/*"]` to the `glob` in the Meson repo that imports the downloaded Meson tgz, although I'd love to understand more of what's happening and why! 😅 (e.g. why the `pyc` files seem to be coming from python 3.11 when I've tried forcing 3.13, etc).

For more context, see [Slack 🧵](https://bazelbuild.slack.com/archives/CA306CEV6/p1734196424601509).

jjmaestro:
Fixes #1342

It seems that pyc files appear in the Meson repo between the first and second time a python toolchain is used and if these are not excluded, the new pyc files invalidate dependent rules, forcing a rebuild.

For more context, see #1342 and [Slack 🧵](https://bazelbuild.slack.com/archives/CA306CEV6/p1734196424601509).


jjmaestro:
I force-pushed the same commit to re-trigger the checks but they all seem to be failing for some reason :(  I'll investigate and see if I can fix it in another PR!

jsharpe:
Its because CI has upgraded to bazel 8 - I've started fixing it here: https://github.com/bazel-contrib/rules_foreign_cc/pull/1340 but need to wait for bazel 8.0.1 as theres a regression with respect to handling of $(location) that is returning more than one file now.

jsharpe:
The fix seems sensible to me though, although like you I'd like to know why its necessary!

jjmaestro:
> The fix seems sensible to me though, although like you I'd like to know why its necessary!

Yeah, totally 😅

From what I gathered, there's a bunch of issues with the `pyc` files but what really surprises me is the Python version in those files... the fact that it seems to be compiled with the standalone interpreter and not with the Python version that's selected in the toolchain.

But alas, I think this is as good as it gets! Maybe when I learn more about how Bazel & some of the rules work I'll have an aha! moment 😄

PR #1346, opened 2024-12-19
framework: avoid forced path suffixing

novas0x2a:
Calling foreign build systems can be fraught, since a lot of custom-written rules do things like execute gcc from the path, instead of using the CC variable. This makes PATH ordering important, to make sure that even if that happens, the toolchain tools get used anyway (and not, e.g. from /usr/bin)

One way to handle that at the toolchain level is to use env_entries (e.g.
https://github.com/bazelbuild/rules_cc/blob/8c94e11c2b05b5f25ced5f23cd07d0cdd36edc1a/cc/private/toolchain/windows_cc_toolchain_config.bzl#L1246) to add the toolchain to the path. You can control whether the toolchain gets prepended or appended by using the token $PATH, e.g. /my/path:$PATH vs $PATH:/my/path.

Unfortunately, the _correct_path_variable helper was unconditionally prepending $PATH to the toolchain path, even though it had just retrieved the value of PATH and no longer needed the original value, making it impossible for the toolchain to prepend a path.

### DIFF
--- a/foreign_cc/private/framework.bzl
+++ b/foreign_cc/private/framework.bzl
@@ -724,11 +724,23 @@ def _correct_path_variable(toolchain, env):
                 corrected_env[key] = ";".join(path_paths)
         env = corrected_env
 
-    value = env.get("PATH", "")
-    if not value:
+    value = env.get("PATH")
+    if value == None:
+        # avoid setting PATH if it isn't set, and vice-versa
         return env
-    value = _normalize_path(env.get("PATH", ""))
-    env["PATH"] = "$PATH:" + value
+
+    value = _normalize_path(value)
+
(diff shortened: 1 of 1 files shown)


novas0x2a:
It looks like CI might currently be broken, unrelated to this PR.

jsharpe:
@novas0x2a looks like there are some genuine issues with this change on windows

novas0x2a:
@jsharpe fixed it. It looks like the problem was that CI depends on strict_action_env not being set (my testing used strict_action_env). This version now intentionally looks for $PATH in the updated PATH, and only prepends it if the PATH token isn't present in the string.

jsharpe (approved):

PR #1348, opened 2025-01-01
Added formatting checks for C/C++ and python code

UebelAndre:


.bazelci/config.yaml
```diff
@@ -252,6 +252,5 @@ tasks:
     test_targets: *min_supported_targets
 
 buildifier:
-  version: "6.1.0"
```
jsharpe, line 255: The version in pre-commit-config.yaml needs to be kept in sync with this

.github/workflows/formatting.yaml
```diff
@@ -0,0 +1,34 @@
...
+          source: '.'
+          extensions: 'h,c,cc,cpp,proto,java'
+          clangFormatVersion: 14
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.11
+      - name: Install dependencies
+        run: |
+          pip install 'black==24.10.0' 'isort==5.13.2'
```
jsharpe, line 28: Any particular reason to choose black over ruff? Ruff can check formatting and isort in a single path with the correct config file.

UebelAndre: I'm old and tend to stick with what I know 😅

jsharpe: Fair enough, lets go with this for now; I can upgrade to ruff at some point if I have spare cycles..

jsharpe (approved):
LGTM

PR #1370, opened 2025-02-27
Allow cmake to use pkg-config for dependencies

chouquette:
Target built with `cmake` are currently unable to find the dependencies `.pc` files.
This PR attempts to fix it

### DIFF
--- a/foreign_cc/private/cmake_script.bzl
+++ b/foreign_cc/private/cmake_script.bzl
@@ -1,6 +1,7 @@
 """ Contains all logic for calling CMake for building external libraries/binaries """
 
 load(":cc_toolchain_util.bzl", "absolutize_path_in_str")
+load("//foreign_cc/private:make_script.bzl", "pkgconfig_script")
 
 def _escape_dquote_bash(text):
     """ Escape double quotes in flag lists for use in bash strings that set environment variables """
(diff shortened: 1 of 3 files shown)


chouquette:
I'm not sure the remaining CI failures are related to the changes, and I can't reproduce locally, I'd appreciate some guidance on this!

jsharpe:
> I'm not sure the remaining CI failures are related to the changes, and I can't reproduce locally, I'd appreciate some guidance on this!

Unfortunately they are related; you need to run the correct version of buildifier on cmake_script.bzl and you need to update the bzl_library target that gets used in the docs generation as you've added a new import to cmake_script.bzl. You should be able to reproduce that one locally by going into the docs folder and running the build for that locally.

chouquette:
Thanks a lot for the pointers! The remaining issues should now be fixed

jsharpe (approved):

foreign_cc/private/cmake_script.bzl
```diff
@@ -130,6 +131,7 @@ def create_cmake_script(
         "CMAKE_BUILD_TYPE": build_type,
         "CMAKE_INSTALL_PREFIX": install_prefix,
         "CMAKE_PREFIX_PATH": merged_prefix_path,
+        "PKG_CONFIG_ARGN": "--define-variable=EXT_BUILD_DEPS=$$EXT_BUILD_DEPS$$",
```
szym, line 134: I've found that building https://github.com/capstone-engine/capstone produces a capstone.pc file with `${EXT_BUILD_ROOT}` left in. So should this also define that variable?

PR #1383, opened 2025-03-17
Add support for Emscripten

allsey87:
This PR adds the definitions for the CPUs and OSes used in WebAssembly, specifically, for Emscripten. Just testing in CI for now.


allsey87:
I have been testing this for a couple weeks and it works well, although perhaps I should remove `wasi` from the `SUPPORTED_OS` array since I don't define any triplets for it.

The only issue I have at the moment is that we have not yet integrated Meson-cross compile support, but I would defer that to another PR.

jsharpe (approved):
LGTM, feel free to add wasi triplets and Meson support in additional PRs.

PR #1390, opened 2025-04-01
Add shellcheck tests for generated build scripts

jsharpe:


jsharpe:
Unfortunately there aren't binaries for shell check on windows so the tests are limited to macOS and Linux.

UebelAndre:
> Unfortunately there aren't binaries for shell check on windows so the tests are limited to macOS and Linux.

I filed https://github.com/bazelbuild/bazel-central-registry/issues/4225 which might one day help with that.

PR #1395, opened 2025-04-26
fix: Make CMAKE_RANLIB the no-op command

benjamin-branchware:


benjamin-branchware:
Hmm, this doesn't seem to be as clear-cut as I thought. Now, the colon is sitting right up against the preceding path for some reason.

Is there some different when using `clang-cl` that I am not aware of? It seems to have differing behaviour in that case.

PR #1401, opened 2025-05-20
meson: add a shared_ldflags_option attr

matt-sm:
Currently the meson rule will always set LDFLAGS to `flags.cxx_linker_executable`.

If there are any executable-specific flags eg. `-pie` in that list, any meson `shared_library()` will fail to link.

Cmake has a feature whereby you can set target-specific linker flags eg. `CMAKE_CXX_LINK_EXECUTABLE`.  Meson does not have a similar concept: [upstream issue](https://github.com/mesonbuild/meson/issues/14621)

As a workaround, this MR adds a new `shared_ldflags_option` attribute to the `meson` rule that will add shared flags to a build option with the specified name.  This attribute follows the `make`  `shared_ldflags_vars` pattern.

Additional change:  the `options` values are not currently expanded.  This PR also adds location expansion to this attr.


matt-sm:
@cerisier keen to merge this

matt-sm:
@novas0x2a I have updated this PR with tests (I eventually figured out how to get rpath working for the darwin shared lib).

The remaining breakage seems to be some kind of remote cache error?

PR #1404, opened 2025-05-27
fix: full_label

### LINKED ISSUE #1403: Invalid registered toolchain `built_meson_toolchain_for_`
I am running into an error when trying to use `meson_with_requirements` from a module introduced via `use_repo_rule` and `new_local_repository`. I have put together an MRE for this issue here: https://github.com/allsey87/meson_bazel_mre. The error that I am getting is as follows:

```
ERROR: /home/developer/.cache/bazel/_bazel_developer/ebf0e4d82830df4ff316ae5fdbbf4910/external/_main~_repo_rules~external_project/BUILD.bazel:11:24: While resolving toolchains for target @@_main~_repo_rules~external_project//:cxx_binary_ (b54b4f9): invalid registered toolchain '@_main~_repo_rules~external_project//:built_meson_toolchain_for_cxx_binary': No repository visible as '@_main~_repo_rules~external_project' from main repository
```

The MRE above has the following layout and attempts to build a simple meson binary via `meson_with_requirements` rule.
```
.
├── BUILD.bazel # empty
├── external_project
│   ├── meson.build
│   └── src
│       └── test.cxx
├── external_project.BUILD # contains the meson_with_requirements invocation
└── MODULE.bazel # creates repository using new_local_repository on external_project
```

Building this project with `bazel build @external_project//:all` gives me the error above. Note that in my real build system I am using `http_archive` instead of `new_local_repository` with a project that is actually external. Moreover, the requirements argument of the `meson_with_requirements` invocation is not an empty list.

The error seems to suggest that the problem is not that the toolchain itself does not exist, but rather the repository that is being referred to does not exist. Perhaps the difference is the extra `@` character when comparing `@@_main~_repo_rules~external_project` and `@_main~_repo_rules~external_project`?
 
Any ideas @jjmaestro @jheaff1 @jsharpe?

jjmaestro:
Fixes #1403

Similar to #1253, `full_label` should use `native.package_relative_label` to build the full canonical label.

Hopefully all of this can be deleted once v6 is out of maintenance (Dec 2025).

### DIFF
--- a/examples/third_party/curl/curl_test.cc
+++ b/examples/third_party/curl/curl_test.cc
@@ -11,7 +11,7 @@ int main(int argc, char* argv[]) {
 
     assertm(std::string(data->version) == std::string("7.74.0"),
             "The version of curl does not match the expected version");
-    assertm(std::string(data->ssl_version) == std::string("OpenSSL/1.1.1o"),
+    assertm(std::string(data->ssl_version) == std::string("OpenSSL/1.1.1w"),
             "The version of openssl does not match the expected version");
 
     return 0;
--- a/examples/third_party/openssl/openssl_repositories.bzl
+++ b/examples/third_party/openssl/openssl_repositories.bzl
@@ -8,12 +8,12 @@ def openssl_repositories():
         http_archive,
         name = "openssl",
         build_file = Label("//openssl:BUILD.openssl.bazel"),
-        sha256 = "9384a2b0570dd80358841464677115df785edb941c71211f75076d72fe6b438f",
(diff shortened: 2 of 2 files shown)


allsey87:
Resolves #1403

allsey87:
@jsharpe could you trigger a rerun of the tests on this PR?

jsharpe:
> @jsharpe could you trigger a rerun of the tests on this PR?

For reference I can't, a new commit is needed to retrigger buildkite runs as I don't have a login for bazel's buildkite.

PR #1416, opened 2025-06-12
Mark extension as reproducible

fmeum:
This avoids a lot of unnecessary lockfile noise.


fmeum:
@jsharpe Could you take a look?

jsharpe (approved):

PR #1418, opened 2025-06-16
feat: extend the extension to support more version pins

novas0x2a:
This makes it possible to pin meson, make, and pkgconfig in the same way it was possible with workspaces. I should say: I don't actually know if this is the correct way to do this, I was just following the existing example and it Seems To Work.

### DIFF
--- a/foreign_cc/extensions.bzl
+++ b/foreign_cc/extensions.bzl
@@ -2,20 +2,41 @@
 
 load("@bazel_features//:features.bzl", "bazel_features")
 load("//foreign_cc:repositories.bzl", "rules_foreign_cc_dependencies")
+load("//foreign_cc/private:default_versions.bzl", _DEFAULT_VERSIONS = "DEFAULT_VERSIONS")
 load("//toolchains:prebuilt_toolchains.bzl", "prebuilt_toolchains")
 
-_DEFAULT_CMAKE_VERSION = "3.31.8"
-_DEFAULT_NINJA_VERSION = "1.13.0"
-
 cmake_toolchain_version = tag_class(attrs = {
-    "version": attr.string(doc = "The cmake version", default = _DEFAULT_CMAKE_VERSION),
+    "version": attr.string(doc = "The cmake version", default = _DEFAULT_VERSIONS["cmake"]),
+})
+
+make_toolchain_version = tag_class(attrs = {
+    "version": attr.string(doc = "The make version", default = _DEFAULT_VERSIONS["make"]),
+})
+
+meson_toolchain_version = tag_class(attrs = {
+    "version": attr.string(doc = "The meson version", default = _DEFAULT_VERSIONS["meson"]),
 })
 
 ninja_toolchain_version = tag_class(attrs = {
-    "version": attr.string(doc = "The ninja version", default = _DEFAULT_NINJA_VERSION),
+    "version": attr.string(doc = "The ninja version", default = _DEFAULT_VERSIONS["ninja"]),
+})
+
+pkgconfig_toolchain_version = tag_class(attrs = {
+    "version": attr.string(doc = "The pkgconfig version", default = _DEFAULT_VERSIONS["pkgconfig"]),
 })
 
 def _init(module_ctx):
+    versions = dict(_DEFAULT_VERSIONS)
+
+    pinned_kw = {}
+
+    for mod in module_ctx.modules:
+        if not mod.is_root:
+            for toolchain_type in versions.keys():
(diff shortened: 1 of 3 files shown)


jsharpe:
Unfortunately its not quite enough as the version name is in the generated repository that the toolchain lives in and so as soon as you switch away from the default versions you will get errors due to unresolved repo names.

novas0x2a:
> Unfortunately its not quite enough as the version name is in the generated repository that the toolchain lives in and so as soon as you switch away from the default versions you will get errors due to unresolved repo names.

Hm, you mean, specifically, like [here](https://github.com/bazel-contrib/rules_foreign_cc/blob/main/MODULE.bazel#L45)? (I'd actually expect that I'd have to keep the register_toolchains list in sync, so maybe that's not unexpected?).

I also discovered [use_all_repos](https://github.com/bazelbuild/bazel-skylib/blob/main/docs/modules_doc.md#modulesuse_all_repos) which provides enough information for `bazel mod tidy` to update the `use_repo` list (but unfortunately not the register_toolchains list).

novas0x2a:
@jsharpe this is the error you meant, right? ```
WARNING: /home/milundy/Code/bazelbuild/rules_foreign_cc/MODULE.bazel:26:22: The module extension tools defined in @rules_foreign_cc//foreign_cc:extensions.bzl reported incorrect imports of repositories via use_repo():

Not imported, but reported as direct dependencies by the extension (may cause the build to fail):
    cmake-3.31.8-linux-aarch64, cmake-3.31.8-linux-x86_64, cmake-3.31.8-macos-universal, cmake-3.31.8-windows-i386, cmake-3.31.8-windows-x86_64, ninja_1.13.0_linux, ninja_1.13.0_linux-aarch64, ninja_1.13.0_mac, ninja_1.13.0_mac_aarch64, ninja_1.13.0_win

Fix the use_repo calls by running 'bazel mod tidy'.
```

If so, I put up a solution over here to help keep the list in sync: https://github.com/bazel-contrib/rules_foreign_cc/pull/1420 (hopefully I understood what you meant?)

jsharpe:
> @jsharpe this is the error you meant, right?
> 
> ```
> WARNING: /path/MODULE.bazel:26:22: The module extension tools defined in @rules_foreign_cc//foreign_cc:extensions.bzl reported incorrect imports of repositories via use_repo():
> 
> Not imported, but reported as direct dependencies by the extension (may cause the build to fail):
>     cmake-3.31.8-linux-aarch64, cmake-3.31.8-linux-x86_64, cmake-3.31.8-macos-universal, cmake-3.31.8-windows-i386, cmake-3.31.8-windows-x86_64, ninja_1.13.0_linux, ninja_1.13.0_linux-aarch64, ninja_1.13.0_mac, ninja_1.13.0_mac_aarch64, ninja_1.13.0_win
> 
> Fix the use_repo calls by running 'bazel mod tidy'.
> ```
> 
> If so, I put up a solution over here to help keep the list in sync: #1420 (hopefully I understood what you meant?)

Not quite; the error occurs if you try to actually configure the version. Yes you need to fix the `if not mod.is_root` line too but when you do that then you can be in the situation where e.g. cmake-3.31.8 repositories are never created, but the rules_foreign_cc MODULE.bazel file trys to register a toolchain from it and so you end up with errors.

PR #1158 is the more complete fix for these issues which moves the toolchains into a single hub which removes the need to know the versions in the module. I need to update the docs on that PR but other than that I think its ready to go so any testing / another pair of eyes on those changes are gratefully received if you have the bandwidth to do so.

novas0x2a:
> Not quite; the error occurs if you try to actually configure the version. Yes you need to fix the `if not mod.is_root` line too but when you do that then you can be in the situation where e.g. cmake-3.31.8 repositories are never created, but the rules_foreign_cc MODULE.bazel file trys to register a toolchain from it and so you end up with errors.

I actually thought about this a bit more, and my gut says now that the problem is you shouldn't actually register the toolchains in rfcc's MODULE.bazel, except as dev_dependencies- registering a toolchain as a non-dev dependency means it can't be disabled by the root module without patching the dep. (This feels like a wart in the bzlmod api that the root module has no way to prevent the registration of toolchains in deps without patches).

I left a bit of a comment about it on the main MR, too: https://github.com/bazel-contrib/rules_foreign_cc/pull/1158#discussion_r2178693151

novas0x2a:
rules_rust appears to have come to a similar conclusion [here](https://github.com/bazelbuild/rules_rust/issues/2818) though the mechanism is different.

PR #1429, opened 2025-07-10
Add ninja 1.13.1

jsharpe:


UebelAndre:
As of https://github.com/bazel-contrib/rules_foreign_cc/pull/1582 you can choose your ninja version by bumping your own BCR pin

PR #1443, opened 2025-10-20
feat: add MSBuild rule to build MSBuild projects.

jsun-splunk:
This is an attempt to add a rule to build MSBuild project from bazel. This rule only supports `MSBuild.exe` from MSVC. `dotnet msbuild` is not supported.

[MSBuild, aka Microsoft Build Engine](https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild?view=vs-2022) is a platform for building applications. This engine provides an XML schema for a project file that controls how the build platform processes and builds software. 

There are no source code to build from for MSBuild. There are also no prebuilt binaries that is available. Therefore, this rule is only availabe for pre-installed toolchain. This usually involves installing MSBuild along side MSVC.

MSBuild generates compile and link flags from project configuration and solution files. I have not been able to find a definitve way to override that so we can applied only the bazel generated flags. As a compromised, the generated bazel flags from cc_toolchain are added to a `msbuild.props` files and passed to MSBuild via the `-p:ForceImportAfterCppTargets` property. This at the minimum confirms bazel flags are applied.

### DIFF
--- a/examples/msbuild_simple/BUILD.bazel
+++ b/examples/msbuild_simple/BUILD.bazel
@@ -6,6 +6,10 @@ msbuild(
     lib_source = "//msbuild_simple/code:srcs",
     out_lib_dir = "",
     out_static_libs = ["mylib.lib"],
+    # Configuration defaults to Debug/Release based on compilation mode.
+    # Platform must match a pair declared in the solution; the solution
+    # declares x64 and Win32, so pin x64 here.
+    platform = "x64",
     sln_file_path = "mysolution.sln",
     target_compatible_with = ["@platforms//os:windows"],
     targets = ["Build"],
--- a/examples/msbuild_simple/BUILD.bazel
+++ b/examples/msbuild_simple/BUILD.bazel
@@ -1,0 +1,19 @@
+load("@rules_cc//cc:defs.bzl", "cc_test")
+load("@rules_foreign_cc//foreign_cc:defs.bzl", "msbuild")
+
+msbuild(
+    name = "mylib",
+    lib_source = "//msbuild_simple/code:srcs",
+    out_lib_dir = "",
+    out_static_libs = ["mylib.lib"],
+    sln_file_path = "mysolution.sln",
+    target_compatible_with = ["@platforms//os:windows"],
+    targets = ["Build"],
+)
+
+cc_test(
+    name = "mylib_test",
+    srcs = ["test_mylib.cpp"],
+    target_compatible_with = ["@platforms//os:windows"],
+    deps = [":mylib"],
+)
--- a/foreign_cc/msbuild.bzl
+++ b/foreign_cc/msbuild.bzl
@@ -1,0 +1,283 @@
+"""# [MSBuild](#msbuild)
+
+This rule is tailored for MSBuild.exe from the Visual Studio installation. It does not support
+the dotnet msbuild execution path. MSBuild determines it own compile and linker flags based
+on the project/solution file configuration. This cannot be fully controlled from bazel like
+other rules (e.g. cmake). We do our best by generating a `msbuild.props` file that is used by
+specifying `-p:ForceImportAfterCppTargets=msbuild.props` to append bazel flags and override
+existing flags where possible.
+
+Since MSBuild is closed source project from Microsoft, there is no prebuilt toolchain available
+and we cannot build it from source. The default is a pre-installed toolchain which assumes
+MSBuild.exe is installed as a part of Visual Studio and locateable by bazel. If you want to
+implement your own MSBuild toolchain you will need to define a toolchain that implements the
+toolchain type `@rules_foreign_cc//toolchains:msbuild_toolchain`. E.g.
+
+```bzl
+toolchain(
+    name = "msbuild_toolchain",
+    exec_compatible_with = [
+        "@platforms//os:windows",
+        "@platforms//cpu:x86_64",
+    ],
+    target_compatible_with = [
+        "@platforms//os:windows",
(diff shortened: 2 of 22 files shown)


.bazelci/config.yaml
```diff
@@ -267,6 +267,8 @@ tasks:
       - "//..."
       # The min supported version of rules_rust is `>=7`
       - "-//rust/..."
+      # 6.5.0 seem to have a bug that does not honor target_compatible_with, so the msbuild target is getting triggered.
```
matt-sm: can you provide a link?

jsun-splunk: I wasn't able to pinpoint an issue or PR. However, using a later version of bazel solves this problem.

The error was that the target under //msbuild_simple/... what had a target_compatible_with for windows was still being executed on linux.

foreign_cc/msbuild.bzl
```diff
@@ -0,0 +1,199 @@
+"""# [MSBuild](#msbuild)
```
matt-sm: do you know if this will auto-magically get a dedicated doc page?

examples/msbuild_simple/BUILD.bazel
```diff
@@ -0,0 +1,19 @@
+load("@rules_cc//cc:defs.bzl", "cc_test")
+load("@rules_foreign_cc//foreign_cc:defs.bzl", "msbuild")
+
+msbuild(
+    name = "mylib",
+    lib_source = "//msbuild_simple/code:srcs",
+    out_lib_dir = "",
+    out_static_libs = ["mylib.lib"],
+    sln_file = "mysolution.sln",
+    target_compatible_with = ["@platforms//os:windows"],
```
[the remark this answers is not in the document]
jsun-splunk: I think this is true for MSVC msbuild, which is the one we adding. However, rule() doesn't allow you to set default `target_compatible_with`. It does support `exec_compatible_with` which also applies in this case. I've added that instead.

foreign_cc/msbuild.bzl
```diff
@@ -0,0 +1,199 @@
...
+```
+# Use env from bazel toolchain, don't override it.
+"-p:UseEnv=true",
+"-p:SetEnvOverride=false",
+
+# Toolchain config
+"-p:PlatformToolset=", # e.g "v143" for Visual Studio 2022
+"-p:VCToolsRedistVersion=",
+"-p:VCInstallDir=<path to Visual C++ install dir>",
+"-p:VCInstallDir_170=<path to Visual C++ install dir for Visual Studios 2022",
```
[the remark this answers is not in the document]
jsun-splunk: fixed

foreign_cc/msbuild.bzl
jsun-splunk: That is what i'm hoping, but I haven't worked out a way to verify yet.

examples/msbuild_simple/BUILD.bazel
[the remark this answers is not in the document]
matt-sm: It seems crazy to have to set this every time you use the rule.  Could we use a macro wrapper?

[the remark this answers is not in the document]
jsun-splunk: I've done this locally via a wrapper, I may upstream it in a future MR.

jsun-splunk:
> > > I'm trying to picture how deps would work eg. msbuild with dep for rfcc/cc_library
> > 
> > 
> > wouldn't it be the same as other rules? i.e what you declare in `out_shared_lib` or `out_static_lib` become part of the CcInfo to be use downstream?
> 
> I am thinking the other way - if msbuild had a dep on another ccinfo

(I don't understand why I can't start a thread on a comment in github. 😞 )

Anyway for your question, the consumption of deps by the msbuild rule would mostly dependant on the upstream build configuration. What the current msbuild rule does is make the dependencies available to msbuild through the properties file. Specifically the `_create_props_file_text` function.

The `include_dirs` are added to `AdditionalIncludeDirectories` for ClCompile action and `lib_dirs` are added to `AdditionalLibraryDirectories` for Link action.

It will be up to the upstream msbuild configurations to use these.

However, the upstream msbuild configuration that I've interacted with (namely python) does even need this.

For example, python exposes `opensslOutDir` and `opensslIncludeDir` to tell it where to locate python. In this case, python does not even need `AdditionalIncludeDirectories` or `AdditionalLibraryDirectories` and openssl can be set as `build_data` to the msbuild rule.

foreign_cc/msbuild.bzl
```diff
@@ -0,0 +1,297 @@
...
+
+    all_args = _properties_to_args(all_properties)
+    all_args.append("-v:{}".format(verbosity))
+    if args:
+        all_args.extend(args)
+    if targets:
+        all_args.append("-t:{}".format(",".join(targets)))
+
+    expanded_args = expand_locations_and_make_variables(ctx, all_args, "args", data)
+    dep_paths = _msbuild_dep_paths(attrs.deps, inputs)
```
jsun-splunk, line 146: the dep paths that gets fed into the `msbuild.props` file now includes exact foreign cc type dep and bazel native type deps.

jsun-splunk:
List of updates in the latest round of updates:

- `msbuild.props` now has calculated _include_/_lib_ directories for bazel native deps and foreign cc deps.
- properly quote msbuild command args. It was working previously, but a simple space would have broken it. E.g. `C:\Program Files\...`
- Fix to thread stardoc properly.
- Changed attr `sln_file` to `sln_file_path` to be more indicative it is path and not a label.
- Added a unittest suite for msbuild.

PR #1444, opened 2025-10-20
do not set deprecated target args

matt-sm:
I was unable to set a setup arg via `target_args`, because the value would be overridden by the empty value in `setup_args`.  And if I tried to set `setup_args` I got a fail saying it was deprecated.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -14,6 +14,16 @@ rules.
 
 ## Setup
 
+### Using bzlmod
+
+Add the following content to your MODULE.bazel:
+
+```python
+bazel_dep(name = "rules_foreign_cc", version = "{version}")
+```
+
+### Using WORKSPACE
+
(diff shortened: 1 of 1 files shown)


jjmaestro:
You are right that my code was wrong, sorry!

Having said that, I don't think this PR addresses or fixes the actual problem. IMHO it's just side-stepping it by deleting a line of code.

I think the proper fix is:

```starlark
for target_name, args_ in deprecated:
    if args_:
        if target_name in target_args:
            fail("Please migrate '{t}_args' to 'target_args[\"{t}\"]'".format(t = target_name))
        target_args[target_name] = args_
```

This will be a no-op if `args_` is the default `[]` for each of the args it checks (all of them are `string_list` without any `default`, so they default to `[]`). If `args_` is set, it will only set the arguments in `target_args` after checking that there's no conflict for that argument. And if there was, it `fails` informing the user to migrate to the new `target_args`.

PR #1445, opened 2025-10-24
Use bzl mirror over pkgconfig.freedesktop.org

travisdowns:
The pkgconfig.freedesktop.org mirror is listed above the bazel mirror of same for pkg-config, but pkgconfig.freedesktop.org is not very reliable. A few months ago it was down for a week and, lately, it is timing out frequently.

Even it does work it is 20-50 _times_ slower than the bazel mirror (measured locally, 25s vs < 1s for mirror).

Flip the order which is already the case for many other deps.


rockwotj:
@cgrindel can you help merge this? Thanks!

PR #1447, opened 2025-11-16
meson: add missing tool paths

matt-sm:
Adds `ar` and `strip` tool paths to meson

Previously, these were taken from the host.

Added some logic for platforms where `strip` does not exist, and also had to add a check to prevent `ar=libtool` on Darwin.


cerisier (approved):
LGTM.

TIL meson hardcodes ARFLAGS.

cerisier:
Worth add @mering as co-author of this btw https://github.com/bazel-contrib/rules_foreign_cc/pull/1302

cerisier:
Closing in favor of #1302 
Attribution kept there as well. 
Thanks for contributing.

PR #1450, opened 2025-11-26
Bump buildifier's version

JSGette:


.bazelci/config.yaml
```diff
@@ -271,6 +271,6 @@ tasks:
     test_targets: *min_supported_targets
 
 buildifier:
-  version: "7.3.1"
+  version: "8.2.1"
   # keep this argument in sync with .pre-commit-config.yaml
```
jsharpe: Can you also upgrade the version in .pre-commit-config.yaml?

PR #1451, opened 2025-12-08
allow out_data_dirs to be referenced through output groups

chouquette:
This change allows one to easily reference the content of folders listed in `out_data_dirs` through an output group, instead of relying on `gen_dir` & `genrule`


rdesgroppes:
@cerisier Would you consider merging it?

PR #1465, opened 2026-02-20
feat: integrate bazel resource_sets into the build actions

novas0x2a:
### summary
This is an attempt to make parallelization of builds safer by telling bazel what resources an action will consume. By default, Bazel allocates 1 CPU and 250M of RAM, and this makes the current methods of parallelization (i.e. `--action_env=CMAKE_BUILD_PARALLEL_LEVEL=16`) pretty unpleasant, since bazel won't know and will happily schedule 160 cores worth of actions on a 10-core machine. Although it doesn't solve the whole parallelization problem (see https://github.com/bazel-contrib/rules_foreign_cc/issues/329) it helps a lot; it reduces the length of our internal build by 30% (by allowing us to fine-tune threads on a per-action basis).

### core design
- All of the build system rules now take a `resource_size` attribute, which can be set to a logical size similar to bazel's other sizing parameters (`default`, `tiny`, `small`, `medium`, `large`, `enormous`).
    -  Each size is associated with a default cpu and memory allocation.
    - The defaults can be overridden as settings; `bazel run @rules_foreign_cc//foreign_cc/settings` will list all the settings in `.bazelrc` format, along with their default values.
- Setting a non-default size will do two things:
    - A resource_set will be passed to the action, which will give bazel some idea about the load the action will create
    - The various parallelization options will be passed to the underlying build systems: `CMAKE_BUILD_PARALLEL_LEVEL`, `GNUMAKEFLAGS`, `MESON_NUM_PROCESSES`, `NINJA_JOBS`. All vars are passed to all build systems, since they can delegate to each other (for example, cmake calling ninja or make).
    - Note that the size is part of the cacheable interface, and changing it will pop the cache.
- `resource_size="default"` (the default) does the same thing as what rfcc/bazel does today (no resource_set, no env var overrides)
    - If a user is currently using `action_env` (e.g. `--action_env=CMAKE_BUILD_PARALLEL_LEVEL=4`) or similar:
        - resource_size="default" will do the same thing as before the change
        - setting an explicit resource_size via any means (on the action itself, or via `@rules_foreign_cc//foreign_cc/settings:size_default`) will override the action_env vars.

### other notes
- I have also added resource_size declarations to several of the base requirement tools `(make|cmake|ninja|pkgconfig)_tool`, which drastically speeds up the CI environment and build iteration speed. (This knocks at least 10 minutes off the `cmake_tool` build time).
- This also adds a wrapper binary for ninja, because [ninja doesn't support environment variables](https://github.com/ninja-build/ninja/issues/1482); this is how `NINJA_JOBS` is implemented. It's c++ instead of bash or python because it needs to be invoked after changing the working directory, and the shim scripts that are used to launch bash or python on windows rely on runfiles, which don't work if you change the wd the way rfcc must.
- My understanding of the [bazel scheduler algorithm](https://github.com/bazelbuild/bazel/blob/4fdff7841202ac6937ea92f82a4d1756e1c3c959/src/main/java/com/google/devtools/build/lib/actions/ResourceManager.java) is:
    - resource_sets tell the scheduler approximately how big an action is
    - resource_sets are hints, they are not enforced by anything
    - the scheduler will try to avoid scheduling too badly over the limit (it does allow some overcommit)
    - the scheduler uses `--local_resources` to determine the limit; see that for defaults
    - if no actions are running and all remaining actions in the queue are over the limit, then the scheduler will schedule actions one at a time.
- AI disclosure: I have used some AI on this PR, but I stress that I (and not the bot) am ultimately responsible for the code. If you're feeling paranoid, the area with the greatest AI involvement is the ninja_wrapper, where I wrote the unix version and then begged codex to save me from having to write windows code.
- Not implemented in this PR:
    - Any method to declare overcommit; for example, to set the resource_set to 10 cpus but then pass `CMAKE_BUILD_PARALLEL_LEVEL=12`. This is probably implementable as a setting if someone feels so inclined.
    - None of the underlying build systems accept memory-limiting parameters, which means the resource_set memory size is even more of a suggestion than the cpu size is!

### DIFF
--- a/examples/make_simple/BUILD.bazel
+++ b/examples/make_simple/BUILD.bazel
@@ -1,34 +1,46 @@
 load("@rules_cc//cc:defs.bzl", "cc_test")
 load("@rules_foreign_cc//foreign_cc:defs.bzl", "make")
 
-make(
-    name = "make_lib",
-    build_data = [
-        "//make_simple/code:cxx_wrapper.sh",
-        "README.md",
-    ],
-    copts = [
-        "-DREQUIRED_DEFINE",
-    ],
-    env = {
-        "CXX_WRAPPER": "$(execpath //make_simple/code:cxx_wrapper.sh)",
-        "README_DIR": "$$(dirname $(execpath README.md))",
-    },
-    lib_source = "//make_simple/code:srcs",
-    out_data_dirs = ["share"],
-    out_static_libs = ["liba.a"],
-)
+SIZES = [
+    "default",
+    "tiny",
+]
 
-cc_test(
-    name = "test_lib",
-    srcs = [
-        "test_libb.cpp",
-    ],
-    copts = [
-        "-std=c++11",
-    ],
-    tags = ["windows"],
-    deps = [
-        ":make_lib",
-    ],
-)
+[
+    make(
+        name = "make_lib_" + size,
+        build_data = [
+            "//make_simple/code:cxx_wrapper.sh",
+            "README.md",
+        ],
+        copts = [
+            "-DREQUIRED_DEFINE",
+        ],
+        env = {
+            "CXX_WRAPPER": "$(execpath //make_simple/code:cxx_wrapper.sh)",
+            "README_DIR": "$$(dirname $(execpath README.md))",
+        },
+        lib_source = "//make_simple/code:srcs",
+        out_data_dirs = ["share"],
+        out_static_libs = ["liba.a"],
+        resource_size = size,
+    )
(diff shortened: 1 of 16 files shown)


novas0x2a:
Note: depends on https://github.com/bazel-contrib/rules_foreign_cc/pull/1464, which should be merged first.

jsun-splunk:
This LGTM. The fact it needed a ninja wrapper, in cpp no less, is a bit crazy.

PR #1541, opened 2026-05-21
framework: add support armv7 and x86_32

### LINKED ISSUE #1540: android: unknown target error when compiling for 32bit after upgrading to v0.15.1
I'm trying to migrate a project from rules_foreign_cc `v0.14.0` to `v0.15.1`. But I'm getting an error when compiling an android target that depends on a `cmake` rule.

Our Android project supports the following cpu architectures: armeabi-v7a, arm64-v8a, x86, x86_64. But it only seems to fail when targeting 32 bit architectures: armeabi-v7a and x86.

In both cases the build command seems to be using the `-arch arm64` flag, which (I believe) is incorrect. Here's the error message:

```
DEBUG: /private/var/tmp/_bazel_jagoba/c980e16e0f0b2c4ec6b42c0d4e784505/external/rules_foreign_cc+/foreign_cc/private/cmake_script.bzl:125:14: target_arch is unknown, please update foreign_cc/private/framework/platform.bzl and foreign_cc/private/cmake_script.bzl; triggered by ...
...
    clang: error: unsupported option '-arch' for target 'arm-linux-androideabi31'
```

jagobagascon:
Adds CMAKE support for armv7 and x86_32 architectures.

fixes #1540


foreign_cc/private/framework/platform.bzl
```diff
@@ -2,10 +2,12 @@
 
 SUPPORTED_CPU = [
     "aarch64",
+    "armv7",
```
novas0x2a, line 5: don't forget to update triplet_name, too; one way to get the correct value is to run `config.guess` on the machine in question, or you can use config.sub to normalize from a different platform.

novas0x2a:
@jagobagascon can you retest with https://github.com/bazel-contrib/rules_foreign_cc/pull/1555 and make sure I didn't break it? I did a more comprehensive fix (android and linux use different names for the same cpu family...)

jagobagascon:
> @jagobagascon can you retest with #1555 and make sure I didn't break it? I did a more comprehensive fix (android and linux use different names for the same cpu family...)

Yes it works. Thank you so much!

novas0x2a:
should be fixed now!

jagobagascon:
Thank you so much!

PR #1542, opened 2026-05-27
feat(make): add module linker flag variables

jsun-splunk:
Add dynamic_module_ldflags_vars to make and configure_make.

These variables are mainly for Darwin builds that need linker flags for loadable modules, such as plugins. They are separate from shared library linker flags because Darwin module links use -bundle, which conflicts with -shared and -dynamiclib.

CMake already handled module linker flags separately; this change makes that behavior consistent with make and configure_make. Also add focused Starlark tests for the new make variables and Darwin flag filtering.

Needs https://github.com/bazel-contrib/rules_foreign_cc/pull/1443 to be merged first due to conflicts.


jsun-splunk:
looks like there is an issue with macos remote cache.

PR #1552, opened 2026-06-03
chore(deps): upgrade rules_cc to 0.2.18

jsun-splunk:
Upgrade rules_cc so rules_foreign_cc can load CcSharedLibraryInfo from rules_cc directly instead of using
bazel_features.globals.CcSharedLibraryInfo.

Also refresh the nested workspace lockfiles and align bazel_features direct dependencies with the resolved rules_cc graph.


jsun-splunk:
checks are failing due to flaky downloads :(

PR #1564, opened 2026-06-13
Only ignore macOS AR if it is libtool

keith:
This way if you've configured `ar` in your toolchain, you get a hermetic version
where if you wipe this out you get the system default


keith:
https://github.com/bazel-contrib/rules_foreign_cc/pull/1069#pullrequestreview-2013545284

keith:
ok needs https://github.com/bazelbuild/apple_support/pull/601

PR #1592, opened 2026-09-15
prebuilt ninja: map the Windows arm64 archives

vadikmironov:
Following #1582 by @UebelAndre which moved away from pkg-config and ancient Windows glib dependency, windows arm64 builds became possible. This is the first follow-up change out of two to make underlying build tools compatible with windows arm64 builds.

`NINJA_TARGETS` maps `win` to x86_64 only. The hub emits one constrained row per mapped platform plus the unconstrained `preinstalled_ninja` row:

```
20_ninja_000_000 … 000_004   five rows, each constrained to a platform (linux aarch64/x86_64, macos aarch64/x86_64, windows x86_64)
20_ninja_001_000             preinstalled_ninja, unconstrained
```

On Windows ARM64 no prebuilt row matches and resolution proceeds to the system entry. If the host happens to have `ninja.exe` on PATH then happy days otherwise build fails. The registry-backed source mode is not a fallback here, since the default tag set registers `binary` and `system` only.

ninja-build has published `ninja-winarm64.zip` since 1.12.0, alongside the archives already mapped. Adding these exactly the same way as for `linux-aarch64` and `mac_aarch64`, but with exclusions for 1.10 and 1.11 series.

`bin` resolves to `ninja.exe` through the existing `"win" in target` check, and the archive is flat like the other ninja zips.


toolchains/private/ninja_versions.bzl
```diff
@@ -168,6 +168,19 @@ NINJA_BIN_SRCS = {
             ],
             bin = "ninja",
         ),
+        ("windows", "aarch64"): struct(
+            urls = [
+                "https://github.com/ninja-build/ninja/releases/download/v1.12.1/ninja-winarm64.zip",
+            ],
+            strip_prefix = "",
+            sha256 = "79c96a50e0deafec212cfa85aa57c6b74003f52d9d1673ddcd1eab1c958c5900",
+            integrity = "",
```
UebelAndre, line 177: We should probably move to integrity.

toolchains/private/ninja_versions.bzl
```diff
@@ -235,6 +248,19 @@ NINJA_BIN_SRCS = {
             ],
             bin = "ninja",
         ),
+        ("windows", "aarch64"): struct(
+            urls = [
+                "https://github.com/ninja-build/ninja/releases/download/v1.13.2/ninja-winarm64.zip",
+            ],
+            strip_prefix = "",
+            sha256 = "e52f0bdef9dfb1003229dbd6508a508c4073fd017247002adc66e5e806cb0391",
+            integrity = "",
```
UebelAndre, line 257: Switch to integrity?

UebelAndre (changes requested):

UebelAndre:
Also, if you're using bzlmod, wouldn't this be fixable within your own repo by bumping the BCR pin of `ninja`?

toolchains/private/ninja_versions.bzl
vadikmironov: Sure, happy to, but if I do it via generator, it will do them all including cmake and meson too. Do you want me to update it partially for new entries or do via generator and if so perhaps via standalone PR once this one is merged? I can do it tomorrow before cmake arm64 second patch no problems.

vadikmironov:
As I understand, `ninja` is already in the graph at 1.13.2 via rules_foreign_cc's MODULE.bazel, but is registered [`binary`](https://github.com/bazel-contrib/rules_foreign_cc/blob/931cb33cf882380394a9c32dcd548259a2a0a357/MODULE.bazel#L78) and [`system`](https://github.com/bazel-contrib/rules_foreign_cc/blob/main/MODULE.bazel#L82) only. 

I guess I can do `tools.ninja(mode = "source", version = "1.13.2")` in the root module, but then it's going to be global and then I'll be building it from source for every platform. WORKSPACE users actually are fine via `rules_foreign_cc_dependencies()` registering source-mode toolchains.

Another option to fix it downstream obviously would be to do what I've been doing by registering an arm64-constrained `native_tool_toolchain`, but again it's a bit awkward to carry that around and hence the PR to push prebuilt binaries for windows arm64.

UebelAndre:
Oh, given the recent advancements of BCR modules I'd say we should promote `source` over `binary` for ninja. If you're a WORKSPACE user I'm happy to do what works for you (as I think WORKSPACE should be deleted) but for bzlmod we should rely on source builds for everything except `cmake` which is a massive project yet to be wrangled.

toolchains/private/ninja_versions.bzl
vadikmironov: You meant something along these lines: https://github.com/bazel-contrib/rules_foreign_cc/pull/1593 ?

vadikmironov:
well, few bumps along this path: https://github.com/bazel-contrib/rules_foreign_cc/pull/1594 . It works afterwards, but I did not give it a proper testing yet. I am on bzlmod and don't have any projects left on WORKSPACE, but #1594 left me wondering how it actually worked for WORKSPACE users if at all...

UebelAndre (approved):
Thank you!

PR #1593, opened 2026-09-16
prebuilt_toolchains: transform archive digests from sha256 to integrity

vadikmironov:
Move prebuilt toolchain sha256 hashes to `integrity` as discussed in review on #1592.

The change includes helper change that re-encodes a hex digest plus the four places that put one into a dict. Everything else is the output of:

    python3 toolchains/prebuilt_toolchains.py

`sha256` is left empty for a generated entry to have a single source of truth.


UebelAndre (approved):
Oh wow, thank you for doing this!

UebelAndre:
I think you got hit by network flakes, can you amend and force push?

vadikmironov:
ok, seems like second time lucky, but now code-format-checks sitting at action_required after the force-push. Is this something you can approve and re-run when you get a chance?

UebelAndre (approved):

vadikmironov:
I am terribly sorry, @UebelAndre , but missed that I had to sign the commit, which then led to another force-push again and now code-format-checks stuck again on this one and ninja from source fix at #1594 . When you get a chance would you be so kind to re-approve it again?

PR #1597, opened 2026-09-21
prebuilt cmake: map the Windows arm64 archives (+ the two ninja rows from #1592)

vadikmironov:
This is a second PR for arm64 support on Windows similar to #1592, but for CMake. `CMAKE_TARGETS` maps `windows-x86_64` and `windows-i386` only and similar to ninja for arm64 resolution falls to system cmake.exe or fails when there is no cmake.exe on PATH.

CMake run also rewrote two ninja arm64 entries from #1592 which got merged and now this is also aligned to integrity field.

Note for #1596: this PR rewrites `prebuilt_toolchains.py` and if lands second needs a small rebase which i am happy to do.


UebelAndre (approved):
Thank you!

